PR #128, opened 2021-06-15
Chris’s review

chris-morgan:
As mentioned, there are currently conflicts but you suggested that I leave the rebasing to you rather than doing it myself (but ask me and I can do it readily enough).

Notes are mostly on the commits. Here are the other notes I made:

src/utils/environment.rs:18, better to avoid String, use FromStr instead of TryFrom<String>.

src/utils/configuration.rs:15, "local".into() is a strictly unnecessary allocation of a String, though this being a one-off thing and the ergonomics of Cow being what they are I generally wouldn’t worry, but the teaching opportunity is good—Cow is worth knowing about. In this instance it ends up a surprisingly painless alteration, but it’s still a technical distraction from the logic, and thus potentially unwarranted.

Good stuff on using .expect(message) rather than .unwrap().

mod.rs files are commonly using `pub mod foo; pub use self::foo::*;`. This exposes the things at two paths. I recommend going one way or the other in each case, either using that module for local structure (`mod foo; pub use self::foo::*;`) or for namespacing (`pub mod foo;`). You can definitely mix the two; tooling should make that pretty painless. But having multiple paths to the one object means you have to keep deciding which to use. (In other news, I’ve become progressively less and less convinced of the wisdom of the last line of the Zen of Python.)

The code base has mod.rs files that are almost exclusively just `mod` and `use` statements. Some of the inner modules, then, are almost empty (e.g. src/server/services/graphql/schema/subscriptions.rs) or even empty (e.g. src/server/services/graphql/schema/subscriptions.rs which is just an impl). I suggest merging at least some of those into the mod.rs file. In the case of src/server/services/graphql/schema/mod.rs, I’d merge context.rs into mod.rs. I can see that types, queries and mutations have value in being separated for code organisation purposes, and I suppose it then makes sense for subscriptions to stay separate in like manner.

Look for places where you take String, and see if you can take &str. Gratuitous cloning is the sort of thing that *can* end up with Rust code performing worse than other languages, though you tend to have to go quite some way before it gets that bad.

src/server/services/graphql/schema/types.rs: lots of effective duplication here, *might* be worth looking into using macros, though probably not at this exploratory/experimental stage.

### DIFF
--- a/src/server/services/graphql/schema/types.rs
+++ b/src/server/services/graphql/schema/types.rs
@@ -7,7 +7,6 @@ use crate::database::DatabaseConnection;
 use juniper::{graphql_object, GraphQLEnum, GraphQLInputObject};
 
 #[derive(Clone)]
-// A name.
 pub struct Name {
     pub name_row: NameRow,
 }
@@ -24,7 +23,6 @@ impl Name {
 }
 
 #[derive(Clone)]
-// A store.
 pub struct Store {
     pub store_row: StoreRow,
 }
@@ -46,7 +44,6 @@ impl Store {
 }
 
 #[derive(Clone)]
-// An item.
 pub struct Item {
     pub item_row: ItemRow,
 }
@@ -63,7 +60,6 @@ impl Item {
 }
 
 #[derive(Clone)]
-// An item line.
 pub struct ItemLine {
     pub item_line_row: ItemLineRow,
 }
@@ -125,7 +121,6 @@ pub enum RequisitionType {
 }
 
 #[derive(Clone)]
-// A requisition.
 pub struct Requisition {
     pub requisition_row: RequisitionRow,
 }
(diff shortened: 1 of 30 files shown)


wlthomson:
Awesome! Thanks @chris-morgan. Will resolve the conflicts and leave this hear for a bit so everyone can ask any followup questions they have.

src/server/services/graphql/schema/types.rs
```diff
@@ -126,21 +120,46 @@ pub enum RequisitionType {
...
+        match requisition_type {
+            RequisitionType::Imprest => RequisitionRowType::Imprest,
+            RequisitionType::StockHistory => RequisitionRowType::StockHistory,
+            RequisitionType::Request => RequisitionRowType::Request,
+            RequisitionType::Response => RequisitionRowType::Response,
+            RequisitionType::Supply => RequisitionRowType::Supply,
+            RequisitionType::Report => RequisitionRowType::Report,
+        }
+    }
+}
```
wlthomson, line 147: @joshxg See https://doc.rust-lang.org/rust-by-example/conversion/from_into.html for explanation as to how this works.

Issue #148, opened 2021-06-23
Enable GraphiQL


joshxg:
GraphiQL (Like, graphical- so cooool) is super useful with autocompletion, syntax highlighting and showing the schema in the UI. It would be great to have it enabled for testing.

The standard is to have this enabled on `/graphiql`.

If OK, I think this can be a part of the graphql service, adding an additional route and handler. Then, I think added into the graphql service config

joshxg:
@wlthomson what you reckon 😁 ?

wlthomson:
Sounds good to me! `actix_juniper` supports this out-of-the-box with `graphiql_handler` (most examples include it, I initially removed it to keep things simple).

wlthomson:
Fixed by #152.

PR #152, opened 2021-06-23
#148 enable graphiql

joshxg:
CHOOO CHOOO


wlthomson:
@joshxg Just took what you had and simplified it a bit! Also, decided to go with `graphiql` as a separate service... maybe a bit OTT, but more consistent with the existing design patterns.

Issue #210, opened 2021-07-12
Githook for `rustfmt`  does not handle deleted files


wlthomson:
Githook for `rustfmt` githook is returning an error status for deleted `.rs` files.

Temporary workaround for anyone encountering this problem is to pass the `--no-verify` flag to `git commit`.

PR #218, opened 2021-07-14
Improve the Git hooks

chris-morgan:
I tried actually using the Git hook (which requires manual installation) today. That got me started on these. Then I deleted files and hit #210, which fortunately I knew how to fix.


wlthomson (approved):
LGTM.

PR #219, opened 2021-07-15
Migrate to async-graphql, and a bit of code flattening/decomplication

chris-morgan:
~~And all the other stuff I’ve done from #216 and #218, because I’m building on top of it because those changes are so broad that it makes life easiest if you expect those to be merged.~~ *(Rebased now that that stuff has landed.)*


src/server/service/graphql/mod.rs
```diff
@@ -1,16 +1,60 @@
...
+// Sugar that helps make things neater and avoid errors that would only crop up at runtime.
+trait ContextExt {
+    fn get_repository<T: anymap::any::CloneAny + Send + Sync>(&self) -> &T;
+}
+
+impl<'a> ContextExt for Context<'a> {
+    fn get_repository<T: anymap::any::CloneAny + Send + Sync>(&self) -> &T {
+        self.data_unchecked::<Data<RepositoryRegistry>>().get::<T>()
+    }
+}
```
chris-morgan, line 20: This here is a Rust pattern called *extension traits*, a way of adding new methods to arbitrary types. You can’t just `impl<'a> Context<'a>` because Rust restricts implementations where the type and trait if applicable all come from another crate, because otherwise you could clash with that crate, so instead you create a new trait, conventionally suffixed `Ext`, and implement *that* for the type(s); then by importing the extension trait, the method becomes available.

Not the nicest syntax with how you have to write the trait name and signature twice, but it’s not bad.

src/server/service/graphql/mod.rs
```diff
@@ -1,16 +1,60 @@
...
+            .route("", actix_web::web::post().to(graphql))
+            // It’s nicest to have the playground on the same URL, but if it’s a GET request and
+            // there’s a `query` parameter, we want it to be treated as a GraphQL query. The
+            // simplest way of doing this is to just require no query string for playground access.
+            .route(
+                "",
+                actix_web::web::get()
+                    .guard(fn_guard(|head| head.uri.query().is_none()))
+                    .to(playground),
+            )
```
chris-morgan, line 41: Deliberate change here from GraphiQL to GraphQL Playground (juniper has helpers for both, async-graphql only supports the playground which seems to be the better of the two), and to putting it on the same URL for convenience. For the record, if you wanted to change the guard to be “is there a query string parameter named `query`”, you could either parse the entire query string (something silly like `url::Url::parse(uri.to_string()).query_pairs().any(|(k, _)| k == "query")`, not really sure why `http::Uri` doesn’t have a method equivalent to `url::Url::query_pairs`), or do the more efficient `.starts_with("query=") || .contains("&query=")` that I’m not sure is robust against percent encoding (`qu%65ry=`) because I don’t know if `http::Uri` normalises the URI in any way.

src/server/service/graphql/mod.rs
```diff
@@ -1,16 +1,59 @@
...
+
+impl<'a> ContextExt for Context<'a> {
+    fn get_repository<T: anymap::any::CloneAny + Send + Sync>(&self) -> &T {
+        self.data_unchecked::<Data<RepositoryRegistry>>().get::<T>()
+    }
+}
+
+pub fn config(
+    registry: Data<RepositoryRegistry>,
+) -> impl FnOnce(&mut actix_web::web::ServiceConfig) {
```
wlthomson, line 24: TIL about `FnOnce` 😃 .

src/server/service/graphql/mod.rs
```diff
@@ -1,16 +1,59 @@
...
+        )
+        .data(registry)
+        .finish();
+        cfg.service(
+            actix_web::web::scope("/graphql")
+                .data(schema)
+                .route("", actix_web::web::post().to(graphql))
+                // It’s nicest to have the playground on the same URL, but if it’s a GET request and
+                // there’s a `query` parameter, we want it to be treated as a GraphQL query. The
+                // simplest way of doing this is to just require no query string for playground access.
```
wlthomson, line 39: Looks good!

src/server/service/graphql/mod.rs
```diff
@@ -1,16 +1,59 @@
...
+                        .guard(fn_guard(|head| head.uri.query().is_none()))
+                        .to(playground),
+                )
+                .route("", actix_web::web::get().to(graphql)),
+        );
+    }
+}
+
+async fn graphql(schema: Data<Schema>, req: Request) -> Response {
+    schema.execute(req.into_inner()).await.into()
```
wlthomson, line 52: Nice and concise!

wlthomson (commented):
A lot of new stuff here for me, so going to review each commit separately (c8f12a3, 4a1e997, f06a166).

c8f12a3 done!

src/main.rs
```diff
@@ -69,13 +69,13 @@ async fn main() -> std::io::Result<()> {
     let listener = std::net::TcpListener::bind(configuration.server.address())
         .expect("Failed to bind server to address");
 
+    let registry = actix_web::web::Data::new(registry);
     let http_server = actix_web::HttpServer::new(move || {
         actix_web::App::new()
-            .data(registry.clone())
+            .app_data(registry.clone())
```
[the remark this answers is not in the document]
chris-morgan: `actix_web::web::Data<T>` is largely just an `Arc` wrapper.

Consequently any clones are just of an `Arc`, which is near enough to free. In theory you could get borrowing working (presuming the underlying type was `Send + Sync`), but in practice actix-web’s design thwarts it because of the tyranny of the `Any`-based polymorphism within which it stores data and the likes, which requires `'static` (that is, it must own the data). I think this is true of all the major web server things, actually, which makes me a bit sad, but as an early implementer in that exact space I understand the pragmatic reasons of exactly why it is so. As I mentioned one time earlier, you *could* box and leak the RegistryRepository and deal with `&'static RepositoryRegistry`, but I’d frown upon such code, better to end up with what’s effectively an `Arc<RepositoryRegistry>`.

In actix-web 3, `.data(foo)` is roughly equivalent to `.app_data(actix_web::web::Data::new(foo))`, though it’s implemented in a different way. In actix-web 4, it first became precisely equivalent, and then `.data` was removed altogether. See https://github.com/actix/actix-web/issues/1454 for that discussion. The reason for the shift here, then, is so that we have a handle on the `Data<RepositoryRegistry>` straight away, so we can pass it to graphql::config, so we keep using just one registry.

I have a vague feeling that we might actually be able to use an `Arc<RepositoryRegistry>` instead of a `Data<RepositoryRegistry>`, but I’m not certain.

So in the end what’s happening is that we’re using the same registry across the entire *system*, where before we were using a registry for each *thread*. The GraphQL resolver gets that same registry.

src/server/service/graphql/mod.rs
chris-morgan: Yay! Showing that you can do stuff like this was truly most of the reason for this last part of the patch; the actual impact of the change itself is negligible.

src/main.rs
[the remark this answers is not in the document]
andreievg: sweet, so `#[derive(Clone)]` on RepositoryRegistry is not needed anymore ?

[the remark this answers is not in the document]
chris-morgan: Correct! And indeed it’s a good idea to remove it to help ensure that there’s only one registry, not multiple by accident.

src/server/service/graphql/mod.rs
wlthomson: Nice! Is the lifetime annotation required due to [the definition of `Context`](https://docs.rs/async-graphql/2.9.8/async_graphql/context/type.Context.html)?

src/server/service/graphql/schema/types.rs
```diff
@@ -90,7 +90,7 @@ impl Store {
     }
 }
 
-#[derive(juniper::GraphQLEnum)]
+#[derive(Enum, Copy, Clone, PartialEq, Eq)]
```
wlthomson, line 93: Maybe personal preference, but I often have to cross-reference with the top of the file to parse some of the `derive(...)` macros. Possibly a reflection of my unfamiliarity with the core libraries, but I think it might be helpful to include the `async_graphql` namespace in situations like this? E.g.

```
#[derive(async_graphql::Enum, Copy, Clone, PartialEq, Eq)]
```

Not as concise, but a bit clearer (to me 😆 ).

wlthomson (commented):
282642278749fcbd4d60803f6b584fef428fff42.

wlthomson (commented):
f06a166.

wlthomson (approved):
A few questions, but no blockers!

src/server/service/graphql/schema/types.rs
chris-morgan: Quite up to you.

src/server/service/graphql/mod.rs
chris-morgan: You can elide lifetimes much of the time in function signatures, but impl block signatures must define all the lifetimes. So yes.

PR #224, opened 2021-07-21
Configuration improvements (improvements)

wlthomson:
Some more improvements to build on #222.

E.g. of what the resulting output looks like:

```
thread 'main' panicked at 'Failed to parse configuration settings: configuration file "/mnt/c/Users/will-desktop/Desktop/dev/repos/configuration-improvements-improvements/configuratio/base" not found'
```


src/util/settings.rs
```diff
@@ -40,3 +45,41 @@ impl DatabaseSettings {
         )
     }
 }
+
+pub enum SettingsError {
```
andreievg, line 49: I think I like this, as long as we don't try make generic generic error type, but by logic area like this (i.e. return of get_configuration), i think it's good, even if it adds 100 lines of code for little (maybe my future more experience rustacean self would differ)

I made a contradictory comment in my PR:

```
Would be nice to have conversion from error returned by current_dir to ConfigError, but they are from external crates (could make a common type, but thought map_err is enough)
```

I think I was afraid that we would make this new type too generic

src/util/configuration.rs
```diff
@@ -1,20 +1,54 @@
...
+    let current_directory_path =
+        env::current_dir().map_err(|err| SettingsError::Path(err.to_string()))?;
+
+    let configuration_directory = current_directory_path.join(CONFIGURATION_DIRECTORY_PATH);
+
+    configuration.merge(
+        File::from(configuration_directory.join(BASE_CONFIGURATION_FILE_PATH)).required(true),
+    )?;
+
+    let environment: Environment = env::var(APP_ENVIRONMENT_VAR)
```
andreievg, line 32: I tend to think that if we are implementing a helper for Environment, it should be:
* Easy to use 
* Allow for more then one environmental variable

I think interface for it should be something like this:

```
use crate util::{Environment, VariableName}
// below would return 'production' or 'local' as String
let environment  = Environment::get(VariableName::AppEnvironment);
```

Sounds like a fun coding exercise (and some generic practice, for multi type ::get return

andreievg (changes requested):
Looks pretty good, some change requests (since you've gone down this path, I think it can made a bit more dynamic).

Also it's a good idea to describe all changes in the PR comment, what do you think ?

src/util/settings.rs
```diff
@@ -40,3 +45,41 @@ impl DatabaseSettings {
...
+    Config(ConfigError),
+    Environment(EnvironmentError),
+    Path(String),
+}
+
+impl Debug for SettingsError {
```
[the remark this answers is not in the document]
wlthomson: Slightly different! If you use the macro, you get something more like:

```
thread 'main' panicked at 'Failed to parse configuration settings: Config(configuration file "/mnt/c/Users/will-desktop/Desktop/dev/repos/configuration-improvements-improvements/configuratio/base" not found)'
```

Notice the `Config(...)` bit. I thought it was nicer to abstract away the underlying error, and I just found it easier to read!

src/util/configuration.rs
```diff
@@ -1,20 +1,54 @@
...
+    let app_environment_file_path = match environment {
+        Environment::Local => LOCAL_CONFIGURATION_FILE_PATH,
+        Environment::Production => PRODUCTION_CONFIGURATION_FILE_PATH,
+    };
+
+    configuration.merge(
+        File::from(configuration_directory.join(app_environment_file_path)).required(true),
+    )?;
+
+    configuration.merge(
```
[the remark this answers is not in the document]
wlthomson: https://github.com/openmsupply/remote-server/pull/222#discussion_r673638039 😆

src/util/configuration.rs
```diff
@@ -1,20 +1,54 @@
...
+static CONFIGURATION_ENVIRONMENT_PREFIX: &str = "app";
+static CONFIGURATION_ENVIRONMENT_SEPARATOR: &str = "__";
 
-    settings.merge(File::from(configuration_directory.join(extra_file)).required(true))?;
+pub fn get_configuration() -> Result<Settings, SettingsError> {
+    let mut configuration = Config::default();
 
-    settings.try_into()
+    let current_directory_path =
+        env::current_dir().map_err(|err| SettingsError::Path(err.to_string()))?;
```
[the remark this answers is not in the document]
wlthomson: I tried, but couldn't figure out what actual `Error` was being returned here! The [docs](https://doc.rust-lang.org/std/env/fn.current_dir.html) just say `Result<PathBuf>`. 

If you can figure out how to do it in the same way as the other errors, would definitely be interested to learn how!

src/util/configuration.rs
wlthomson: Yeah, I agree that `Environment` has a very restricted application ATM. I think there's also the option to use generics and traits to have something like:

```
Environment<Environment::VarApp>
```

Where `Environment::VarApp` implemented a trait, e.g. `Environment::Var` (that way the behaviour would always be consistent, and could include things like built-in defaults etc.). 

Just another option, I like your way too!

src/util/settings.rs
wlthomson: Yeah, from what I understand it's quite a common pattern in Rust for handling multiple errors types in a single function. I think if functions are kept well-defined and we respect the SRP rule, then we should be well-placed to avoid over-genericisation!

src/util/configuration.rs
wlthomson: I think this is a relatively bigger design discussion, so maybe best left for a dedicated issue?

src/util/settings.rs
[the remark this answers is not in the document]
andreievg: Ahh ok, in that case can you change Display fmt to just return `write!(f, "{}", self)`. I wonder if there is a way to say use Display trait for my Debug yow !

src/util/configuration.rs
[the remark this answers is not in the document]
andreievg: It's std::io::Error

[std-io-error.gif.zip](https://github.com/openmsupply/remote-server/files/6852848/std-io-error.gif.zip)

src/util/settings.rs
andreievg: Hmm on second thought though, are we expecting to do any logic based on different config errors ? I would image it unrecoverable state with some panic and logging. I wonder if it's possible to just tell rust that get_configuration -> `Result< Settings, Display + Debug>` (I played around with it with no success)

src/util/settings.rs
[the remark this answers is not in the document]
chris-morgan: `{}` uses `Display`. (`{:?}` would use `Debug`.) https://doc.rust-lang.org/std/fmt/index.html#formatting-traits

src/util/settings.rs
chris-morgan: For *one* trait, you could use a trait object: `Result<Settings, Box<dyn Display>>`. Eventually you’ll come across things like `dyn Foo + Send + Sync`, but you can only do that for specially-designated marker traits, not all traits. (It comes down to some moderately complex vtable reasons; Rust might eventually support doing it, but for now you can only work around it on a case-by-case basis.)

For easier error type definitions, it’s common to use helper crates; I think the most popularly recommended ones at present are thiserror for libraries and anyhow for applications (read their docs and you’ll see the conceptual differences).

src/util/configuration.rs
andreievg: > Where Environment::VarApp implemented a trait, e.g. Environment::Var (that way the behaviour would always be consistent, and could include things like built-in defaults etc.).

It seems pretty trivial, not sure if we need to add specific trait for it, it's either a string with default value or error, or a number with default value or error, we just need to describe what is the mapping between names, values types and defaults. I guess you are thinking about different data types for the future (arrays via env vars ?)

> I think this is a relatively bigger design discussion, so maybe best left for a dedicated issue?

Ok, sure, although I wonder what the difference is between this design decision and the one in this PR (for generic settings error), btw my pr had a comment saying i considered it, but refrained, which kind of left it open for discussion.

Anyways I am happy if we make an issue to discuss and reach a consensus on implementation of generic envs, before any further code.

However: since this is one env variable Environment struct, I think it should contain the full logic and be easy to use, without 3 method chain.

i.e. to be used this way:
```
const environment = Environment::as_string()?;
configuration.merge(
        File::from(configuration_directory.join(environment).required(true),
    )?;
```

src/util/settings.rs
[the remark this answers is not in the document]
andreievg: Thanks Chris, my comment was mainly about the difference of derived Debug trait vs implemented one here, and there was a difference as pointed out by Will. And the second comment suggested replacing Debug impl with just `write!(f, "{}", self)`, since it's a complete duplicate of Display impl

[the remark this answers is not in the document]
chris-morgan: Gotcha; I’m just skimming and missed that.

One other thing you can do is to call it explicitly: `fmt::Display::fmt(self, f)` (which can also be written in Universal Function Call Syntax as `<Self as fmt::Display>::fmt(self, f)`, but not as `self.fmt(f)` because ambiguity).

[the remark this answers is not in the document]
wlthomson: Hm, maybe I'm misunderstanding something here, but the `Debug` and `Display` implementations are deliberately different here (note that display formats its variants with `"{}"` and debug with `"{:?}"`).

The idea was to make sure `"{:?}"` and `"{}"` behaves similarly for `SettingsError` as it does for it's underlying variants. I think making the `Debug` implementation `write!(f, "{}", self)` would mean this property waould be lost, no?

src/util/settings.rs
wlthomson: Again, maybe missing something, but with the current definition of `SettingsError`, you should be able to match variants recursively (i.e. match on `ConfigError` and also match on the variants of `ConfigError`)?

src/util/configuration.rs
wlthomson: You're right that theres no fundamental difference, I just think it's a tidy point at which we can draw the line on the scope of this PR. I think there's a lot of ways to do it and it would be cool to throw around some ideas without bloating this PR up even more from what I have already done 😆 .

wlthomson:
> Also it's a good idea to describe all changes in the PR comment, what do you think ?

@andreievg Oops, missed this. I agree! I tried to make my commits as self-documenting as possible, but did slack on the PR description. I was thinking it might be a good pattern going forward to adopt @chris-morgan use of commit descriptions (unfortunately ship has sailed on this PR, at least without rewriting history). It seems like you have a pretty good idea of the changes now, but I can add more detail in the description if you'd like?

src/util/configuration.rs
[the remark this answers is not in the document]
wlthomson: Nice spot! About to sign off for the night, so will make this change when I'm back at my machine tomorrow!

src/util/settings.rs
andreievg: Sure, in my comment:
`are we expecting to do any logic based on different config errors ? I would image it unrecoverable state with some panic and logging`

The above got me thinking that perhaps there is no need to fine grain SettingsError implementation, and was playing around with alternatives (which proved difficult for me).

Anyways thanks Chris for the explanation, I am going to play around trying to make a type that's both Debug and Display and can also `From` any Debug + Display, famous last words.

src/util/settings.rs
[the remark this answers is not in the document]
andreievg: True, forget about that, sorry comment miss-leading. (i thought both implementations are fully equivalent)

src/util/settings.rs
wlthomson: Ah, I misunderstood! I thought you were suggesting we couldn't do that with the current implementation, but I understand now you are asking whether it is redundant? Doesn't that take us back to the original implementation? It seems to me that logging and panicing is the use-case that `.expect` is designed for?

andreievg:
@wlthomson
> @andreievg Oops, missed this. I agree! I tried to make my commits as self-documenting as possible, but did slack on the PR description. I was thinking it might be a good pattern going forward to adopt @chris-morgan use of commit descriptions (unfortunately ship has sailed on this PR, at least without rewriting history). It seems like you have a pretty good idea of the changes now, but I can add more detail in the description if you'd like?

No need.

> I was thinking it might be a good pattern going forward to adopt @chris-morgan use of commit descriptions

Sure, if we can be that pragmatic, great (only issue would be with commits upstream that changes things in previous commits, the commit descriptions would be confusing for me since I prefer looking at full final diff).

src/util/configuration.rs
andreievg: Agree, would be happy to approve without dynamic env vars, but since this is an improvement PR that introduces back Environment, would like to see how this would look
```
const environment = Environment::as_string()?;
configuration.merge(
        File::from(configuration_directory.join(environment).required(true),
    )?;
```
I can almost guarantee it would reduce code and complexity (not suggesting it's complex)

src/util/configuration.rs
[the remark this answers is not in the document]
andreievg: https://github.com/openmsupply/remote-server/pull/222#discussion_r673640270

src/util/settings.rs
andreievg: > It seems to me that logging and panicing is the use-case that .expect is designed for?

Originally it was `expect` or '?' I wasn't sure why return ConfigError `?` or Panic with `.expect` in the same funct (thought it would be more consistent to have one or the other), and `?` seemed like a nicer syntax. Perhaps it's ok to use both all along.

Moral of the story, SettingsError still cool, can implement std::io::Error, and good to go on that front, the Debug + Display thing is just for me to play around, i'll share the result sometimes.

chris-morgan: FYI on the Debug + Display thing: the approach to take is roughly: `trait DebugPlusDisplay: Debug + Display`, `impl<T: Debug + Display> DebugPlusDisplay for T` and `impl Debug for dyn DebugPlusDisplay` and `impl Display for dyn DebugPlusDisplay`. Normally the trait could have no extra methods, but I think you might need to add methods to get around the naming collision.

andreievg: Ok, i tried a few way, only successful one for me was this: 

https://github.com/openmsupply/remote-server/compare/configuration-improvements-improvements...display-debug1

this one failed since DebugPlusDisplay doesn't auto convert to T + Display + Debug
https://github.com/openmsupply/remote-server/compare/configuration-improvements-improvements...display-debug1

With the first one i still had a few hickups, in the comments, and it store debug and display in two seperate strings : /, i had something like this (boxed), but had borrow issues:

```RUST
pub struct DisplayDebug {
    display: Box<&'static dyn Debug>,
    debug: Box<&'static dyn Display>,
}
...

   fn from(err: T) -> Self {
        DisplayDebug {
            display: Box::new(&err),
            debug: Box::new(&err), // already borrowed
        }
    }
```

Anyways, closing that chapter for now, thanks for help.

chris-morgan: I remembered the `impl Debug`/`impl Display` wasn’t actually necessary. Here’s what you need:

```rust
trait DebugAndDisplay: fmt::Debug + fmt::Display {}

impl<T: fmt::Debug + fmt::Display> DebugAndDisplay for T {}

fn inspect(x: &dyn DebugAndDisplay) {
    println!("Debug says {:?}, Display says {}", x, x);
}

fn main() {
    inspect(&"Hello, world!");
}
```

For storage, you’d use `Box<dyn DebugAndDisplay>`, and `&x` (`&Box<dyn DebugAndDisplay>`) will coerce to `&dyn DebugAndDisplay` so any methods that take that would work. (One last note: the syntactic grouping is different from `&mut T`: `&dyn Trait` is equivalent to `&(dyn Trait)`, an immutable reference to a dynamically sized type, but `&mut` is a type of reference and `&(mut T)` is a syntax error.)

src/util/configuration.rs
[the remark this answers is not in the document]
Chris-Petty: #222 (comment)

andreievg:
@wlthomson, just FYI, the only two things remaining is adding std::io::Error to SettingError, and making Environment just return 'local' or 'production' (I think that would simplify thing a little).

Also do you have any idea why ConfigError doesn't auto convert to SettingError (i've also encountered that when trying the dynamic thingy)

src/util/configuration.rs
wlthomson: I'm not sure if you are suggesting removing the chained `configuration.merge` calls? If so, I think I disagree, mainly because I think it's good to differentiate between environment variables, files and configurations. They are all working together here (i.e. env variable to find the file which specifies the configuration, with optional env variables to override specific configuration properties). TLDR; I agree that the environment variable should be easy to use to get the environment variable value, but I think its responsibility should be limited to that (i.e. it doesn't have any opinion or responsibility re. how that value is interpreted or used, that's up to `get_configuration`).

However! If you're suggesting getting rid of the parsing back-and-forwards for the `APP_ENVIRONMENT` value, I can get on board with that 😃. I've simplified a few things in 1fc934b, and I think it makes it a bit tider (I actually quite like the elegance of the chained `.merge(...)` expressions!) maybe that's a decent compromise (for now)?

wlthomson:
> adding std::io::Error to SettingError

2aa4653

> making Environment just return 'local' or 'production' 

1fc934b (see also [my reply above](https://github.com/openmsupply/remote-server/pull/224#discussion_r674597999)).

> Also do you have any idea why ConfigError doesn't auto convert to SettingError (i've also encountered that when trying the dynamic thingy)

Not sure what you mean by auto convert, do you have an example?

src/util/settings.rs
```diff
@@ -40,3 +47,47 @@ impl DatabaseSettings {
...
+        }
+    }
+}
+
+impl Display for SettingsError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        match self {
+            SettingsError::Config(err) => write!(f, "{}", err),
+            SettingsError::Environment(err) => write!(f, "{}", err),
+            SettingsError::File(err) => write!(f, "{}", err),
```
wlthomson, line 72: @chris-morgan No rush on this, but whenever you have time... is there a way to simplify these repeated expressions into a one-liner? Psuedo-code of what would be ideal, would be something like:

```
match self { 
  _(err) => write!(f, "{}", err)
}
```

chris-morgan: No.

If you had enough cases using this pattern you could consider a macro:

```rust
macro_rules! map_all_variants {
    ($self:ident, _($err:ident) => $expr:expr) => {
        match $self {
            SettingsError::Config($err) => $expr,
            SettingsError::Environment($err) => $expr,
            SettingsError::File($err) => $expr,
        }
    }
}

impl fmt::Display for SettingsError {
    fn fmt(&self, f: fmt::Formatter) -> fmt::Result {
        map_all_variants(self, _(err) => write!(f, "{}", err))
    }
}
```

But I wouldn’t recommend this until you had a lot more code, and I suspect it’s not a good sign about the code anyway if you go like that.

In this case, I’d expect to be mentioning something about the enum variant, e.g. “config error: {}”, “environment error: {}” or “file error: {}” (not good text, but you see the idea). Debug especially should mention the variant, with the wrapped error more debuggy-outputty rather than Englishy. This is all ideals, of course. In practice, do whatever’s most useful!

src/util/configuration.rs
andreievg: >  I'm not sure if you are suggesting removing the chained configuration.merge calls

@wlthomson, nope, .merge is good to me. Btw my suggestions was (which has merge)
```
const environment = Environment::as_string()?;
configuration.merge(
        File::from(configuration_directory.join(environment).required(true),
    )?;
```

> I've simplified a few things in 1fc934b

yeah looking good, except you should probably use `get` not `try_get` in `configuration.rs`, unless the intention is to make sure that environmental variable is always set before running the script.

Also i would have though you would keep the check that limits to `local` or `production` (error if not match), but if it's not defined use default.

i.e.

```RUST
    fn as_string() -> Result<String, String> {
        match env::var(APP_ENVIRONMENT_KEY) {
            Err(err) => Ok(String::from(APP_ENVIRONMENT_LOCAL)),
            Ok(v) if v.eq_ignore_ascii_case(APP_ENVIRONMENT_PRODUCTION) => Ok(v),
            Ok(v) if v.eq_ignore_ascii_case(APP_ENVIRONMENT_LOCAL) => Ok(v),
            _ => Err("Only 'local' or 'production' is allowed in APP_ENVIRONMENT environmental variable"),
        }
    }
```

andreievg:
> Not sure what you mean by auto convert, do you have an example?

I am confused about:

```
  configuration
        .try_into()
        .map_err(|err| SettingsError::from(err))
```

And why `configuration.try_into()` by itself doesn't work, i thought that 'from' would happen automatically (from ConfigError to SettingsError), i.e. this works:
```
    Ok(configuration.try_into()?)
```

but not
```
configuration.try_into()
```

which seems equivalent to me.

Anyways, thanks for latest changes, i've made another comment

https://github.com/openmsupply/remote-server/pull/224#discussion_r676218547

And also forgot to mention this one: https://github.com/openmsupply/remote-server/pull/224#discussion_r673618145

Sorry for prolonging approval of this pr, the above two points should be very quick to fix though

src/util/configuration.rs
wlthomson: > Also i would have though you would keep the check that limits to local or production (error if not match), but if it's not defined use default.

Yup, that's the idea of `get` (if not `production`, then it defaults to `local`), whereas `try_get` was not meant to apply a default, and was meant to be more like your suggestion (I must have thought I had updated it before committing).

wlthomson: Actually, I remember the intention now. The idea with `get` is that it always applies a default (including if an error occurs). In comparison, `try_get` will also always apply a default, but won't smother the `VarError`. Does that make sense?

wlthomson: > And why `configuration.try_into()` by itself doesn't work, i thought that 'from' would happen automatically (from ConfigError to SettingsError), i.e. this works:
> 
> ```
>     Ok(configuration.try_into()?)
> ```
> 
> but not
> 
> ```
> configuration.try_into()
> ```
> 
> which seems equivalent to me.

As I understand it, it's because `try_into` is attempting to deserialize into `Result<Settings, SettingsError>`. As `SettingsError` doesn't implement the `Deserialize` trait, this doesn't compile. The first example does work, as you are only attempting to deserialize into `Settings` (the `Result<Settings, SettingsError>` part is handled by the `Ok(...)`.

I have updated to:

```
let mut configuration: Config = Config::default();
configuration
    .merge(get_configuration_base_file()?)?
    .merge(get_configuration_app_file()?)?
    .merge(get_configuration_environment())?;
let settings: Settings = configuration.try_into()?;
Ok(settings)
```

which I think strikes a decent balance of clarity and conciseness.

wlthomson:
@andreievg 

> And why `configuration.try_into()` by itself doesn't work, i thought that 'from' would happen automatically (from ConfigError to SettingsError), i.e. this works:
> 
> ```
>     Ok(configuration.try_into()?)
> ```
> 
> but not
> 
> ```
> configuration.try_into()
> ```
> 
> which seems equivalent to me.

As I understand it, it's because `try_into` is attempting to deserialize into `Result<Settings, SettingsError>`. As `SettingsError` doesn't implement the `Deserialize` trait, this doesn't compile. The first example does work, as you are only attempting to deserialize into `Settings` (the `Result<Settings, SettingsError>` part is handled by the `Ok(...)`).

I have updated to:

```
let mut configuration: Config = Config::default();
configuration
    .merge(get_configuration_base_file()?)?
    .merge(get_configuration_app_file()?)?
    .merge(get_configuration_environment())?;
let settings: Settings = configuration.try_into()?;
Ok(settings)
```

which I think strikes a decent balance of clarity and conciseness.

andreievg (approved):
Thanks for all of the changes ! 

Looking good. I am sure we can speed the review process up with future PRs.

Thinks @chris-morgan for your input

wlthomson:
> 
> 
> Thanks for all of the changes !
> 
> Looking good. I am sure we can speed the review process up with future PRs.
> 
> Thinks @chris-morgan for your input

Thi PR was as much a code audit as a feature change (and this may be a trend for the foreseeable future while we get to grips with the Rust language and the crates we are using), so the speed of the review can be interpreted in that context. It might be OK to accept that reviews will go slow-ish at these early stages (in the hope that they will speed up once we get down to more granular features and bug fixes).

Disclaimer: I really like the kanban idea of recording metrics on wait times etc. but I don't think we are there yet!

Issue #236, opened 2021-08-04
[SYNC] - Refine Structure of Sync Processor


andreievg:
MW job: `[OMSBES1] omSupply Back End - Refine Structure of Sync Processor`

### Description

Chris provided us with a pretty good starting point, want to expand this to a more refined skeleton

### Scope

Sync structure in place:
<ins>Includes</ins>
* Relocate sync from main 
* Provide sync with handles to:
   * DB
   * Channel to send msgs
* Placeholders for sequence of actions sync needs to perform
   * I think we just want a mechanism to process things sequentially (i.e. try to pull central data, try to pull remote data, try to push sync q)  with choice of a given action happening on specific error type, )i.e. can't reach server, error integrating record)

<ins>Excludes</ins>
* Implementation of actual sync actions
* Specifying all of the actions that need to happen and their sequence
* Implementation of cursor handling or any DB actions

### Tasks
- [x] MW job added
- [x] https://github.com/openmsupply/remote-server/issues/234
- [x] https://github.com/openmsupply/remote-server/issues/215
- [x] https://github.com/openmsupply/remote-server/issues/240

wlthomson:
@andreievg I'm still adding work under this to the rust prototype job, can we get this job made soonish?

wlthomson:
Closing as all tasks completed!

PR #239, opened 2021-08-15
#215 Add sync consumer struct

### LINKED ISSUE #215: Add sync consumer struct
From epic: https://github.com/openmsupply/remote-server/issues/236

Add struct (and traits?) for consuming the `sync_queue` from the central server.

Thinking to keep it relatively simple for now. We need to fetch data from `v5/sync_queue`, and integrate it into the db. 

As the current tables are pretty minimal, the mapping logic can probably be done explicitly for now, just as a proof of concept.

I was thinking it would be quite cool to have an interface compatible with #214, e.g. to push data would be something like:

```
central_server = CentralServer::fromUri( // uri of central server );

let records_to_integrate = central_server.sync_queue.get_records( // ...params, e.g. number of records );

// Do something with `records_to_integrate`

let records_to_push = sync_queue.get_records( /// ...params, same as above ); // note this is the local sync queue, not the central server one

central_server.push(records_to_push);
```

Just a rough idea, but I like the idea of abstracting the concept of the queues from the actual messiness of the endpoints and `sync_out` records!

wlthomson:
Closes #215.

Lots of stuff here, but should build a decent foundation for other issues in #236 (although goes a bit out-of-scope as also includes some implementation logic for pulling remote records).


src/util/sync/connection.rs
```diff
@@ -0,0 +1,119 @@
...
+impl SyncConnection {
+    pub fn new(settings: &SyncSettings) -> SyncConnection {
+        let host = settings.host.clone();
+        let port = settings.port.clone();
+        let username = settings.username.clone();
+        let password = settings.password.clone();
+
+        let client = Client::new();
+        let server = SyncServer::new(host, port);
+        let credentials = SyncCredentials::new(username, password);
```
andreievg, line 28: don't need to clone any of the strings, just pass as ref

src/util/sync/connection.rs
```diff
@@ -0,0 +1,119 @@
...
+    // Should only be called on initial sync or when re-initializing an existing data file.
+    //
+    // TODO: add custom error to return type.
+    pub async fn initialize(&self) -> Result<SyncQueueBatch, reqwest::Error> {
+        // TODO: add error handling.
+        let url = self.server.initialize_url();
+
+        // Server rejects initialization request if no `content-length` header included.
+        let mut headers = HeaderMap::new();
+        headers.insert(CONTENT_LENGTH, "0".parse().unwrap());
```
andreievg, line 48: surely reqwest handles this automatically ?

src/util/sync/connection.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        let request = self
+            .client
+            .post(url)
+            .basic_auth(&self.credentials.username, Some(&self.credentials.password))
+            .headers(headers);
```
andreievg, line 54: i see where you are heading with sync `SyncCredentials` and `SyncServer`, i think a little bit more abstraction would be good (combine the two and the client, so that methods in SyncConnection can all use that abastraction)

src/util/sync/connection.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        Ok(sync_batch)
+    }
+
+    // Pull batch of records for sync queue.
+    //
+    // TODO: add custom error to return type.
+    pub async fn queued_records(&self) -> Result<SyncQueueBatch, reqwest::Error> {
+        // Arbitrary batch size.
+        const BATCH_SIZE: u32 = 500;
```
andreievg, line 68: this is also a setting

src/util/sync/connection.rs
andreievg: Also it can include some basic error handling

src/util/sync/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+            info!("Sending queued records request...");
+            sync_batch = self
+                .connection
+                .queued_records()
+                .await
+                .expect("Failed to pull sync queue records");
+            info!("Received queued records response");
+
+            // TODO: acknowledge after integration.
+            if let Some(data) = sync_batch.data {
```
andreievg, line 96: FYI, there are two different ways to sync records, central log and sync queue.

With central log we increment counter.

With sync queue we acknowledge, for acknowledge we would store acknowledgement if it fails to sync (at which stage we stop sync and on start sync we try to send acknowledgement first).

Don't need to worry about that in this PR but something to be aware of

andreievg (changes requested):
Look very good! A few change requests thought.

Haven't reviewed the test, looks like they are trying to mock real server (which is ok for now but, actual integration tests need to be done against real services).

src/util/sync/connection.rs
wlthomson: Required a bit of refactoring (and still need to do some cloning, as moving `username` into a struct), but maybe this is more what you were thinking: 1c1991d.

src/util/sync/connection.rs
wlthomson: Doesn't seem so. If you remove this, server will reject the request! My guess is that `reqwest` doesn't generate `content-length` headers for `POST` requests without bodies.

src/util/sync/connection.rs
wlthomson: Yeah I agree with all this... I stopped because the PR was already getting pretty big. I want to get this merged so I can use it as a base for some gql investigation, do you have a problem with this abstraction being done in another PR?

src/util/sync/connection.rs
wlthomson: I wasn't sure if it would cause some angry reactions if I put this into the `.yml` configs! As above, ok for this to be done in a separate PR (I'm also aware my PRs are quite big lately, so I'm trying to break tasks down into easier to review chunks)?

src/util/sync/mod.rs
wlthomson: Yup, I just focused on sync queue here as the central log required a bit more fiddling with cursors etc!

wlthomson:
Merging this so that branches don't fall too far behind now that @clemens-msupply is starting to work on this.

PR #247, opened 2021-08-18
#246 Add default APP_ENVIRONMENT

### LINKED ISSUE #246: Error on running app without APP_ENVIRONMENT
If `APP_ENVIRONMENT` is not set, the following error is thrown:

```
thread 'main' panicked at 'Failed to parse configuration settings: NotPresent', src/main.rs:156:44
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

wlthomson:
Closes #246.

### DIFF
--- a/src/util/configuration.rs
+++ b/src/util/configuration.rs
@@ -35,8 +35,7 @@ pub fn get_configuration_base_file() -> Result<File<FileSourceFile>, SettingsErr
 /// environments are `local` and `production`.
 pub fn get_configuration_app_file() -> Result<File<FileSourceFile>, SettingsError> {
     let configuration_directory = get_configuration_directory()?;
-    let app_file =
-        File::from(configuration_directory.join(AppEnvironment::try_get()?)).required(true);
+    let app_file = File::from(configuration_directory.join(AppEnvironment::get())).required(true);
     Ok(app_file)
 }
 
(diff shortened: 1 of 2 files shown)


andreievg:
readme too ?

wlthomson:
> 
> 
> readme too ?

9076a10.

src/util/configuration.rs
```diff
@@ -35,8 +35,7 @@ pub fn get_configuration_base_file() -> Result<File<FileSourceFile>, SettingsErr
 /// environments are `local` and `production`.
 pub fn get_configuration_app_file() -> Result<File<FileSourceFile>, SettingsError> {
     let configuration_directory = get_configuration_directory()?;
-    let app_file =
-        File::from(configuration_directory.join(AppEnvironment::try_get()?)).required(true);
+    let app_file = File::from(configuration_directory.join(AppEnvironment::get())).required(true);
```
clemens-msupply, line 38: Unrelated, why is there a try_get() method? seems to be not needed?

clemens-msupply (approved):

src/util/configuration.rs
wlthomson: Tbh, it's not really needed anymore. I left it in as an example of a "pattern" that I like and think would be good to follow throughout the codebase (I've noticed it's used by a lot of rust crates). It goes completely against the laws of YAGNI, but I think it's fine to leave in there and exposes a nice interface (I could see the configuration stuff all being modularised as a crate at some point).

wlthomson: Going to go ahead and merge (if anyone strongly opposes the `try_get()` being included, I won't protest if they remove it).

PR #250, opened 2021-08-19
#249 Add items query

### LINKED ISSUE #249: Add items query
Add gql query to return list of items.

wlthomson:
Closes #249.

### DIFF
--- a/src/database/repository/mock/item.rs
+++ b/src/database/repository/mock/item.rs
@@ -23,15 +23,21 @@ impl ItemRepository {
     }
 
     pub async fn find_all(&self) -> Result<Vec<ItemRow>, RepositoryError> {
-        let mut items = vec![];
-        self.mock_data.lock().unwrap().clone().into_iter().for_each(
-            |(_id, row): (String, DatabaseRow)| {
-                if let DatabaseRow::Item(item) = row {
-                    items.push(item);
-                }
-            },
-        );
-        Ok(items)
+        let filter_item = |row: &DatabaseRow| -> Option<ItemRow> {
+            if let DatabaseRow::Item(item) = row {
+                Some(item.clone())
+            } else {
+                None
+            }
+        };
+
+        Ok(self
+            .mock_data
+            .lock()
+            .unwrap()
+            .values()
+            .filter_map(filter_item)
+            .collect())
     }
 
     pub async fn find_one_by_id(&self, id: &str) -> Result<ItemRow, RepositoryError> {
--- a/src/database/repository/mock/item.rs
+++ b/src/database/repository/mock/item.rs
@@ -22,6 +22,18 @@ impl ItemRepository {
         Ok(())
     }
 
+    pub async fn find_all(&self) -> Result<Vec<ItemRow>, RepositoryError> {
+        let mut items = vec![];
+        self.mock_data.lock().unwrap().clone().into_iter().for_each(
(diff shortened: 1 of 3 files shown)


src/database/repository/mock/item.rs
```diff
@@ -22,6 +22,18 @@ impl ItemRepository {
         Ok(())
     }
 
+    pub async fn find_all(&self) -> Result<Vec<ItemRow>, RepositoryError> {
+        let mut items = vec![];
+        self.mock_data.lock().unwrap().clone().into_iter().for_each(
```
clemens-msupply, line 27: Why do you use clone() here?

clemens-msupply (commented):
Should the mock also get a info log?

src/database/repository/mock/item.rs
wlthomson: To put it simply, iterating over the hashmap values requires "consuming" the hashmap. Since we are passing `&self` as a ref, we promised not to mutate any local state. By cloning the data, we can iterate over it without any problems.

It's not the most efficient way to do things, and you could probably get around it, but it keeps things consistent and we're not too worried about performance for mocks anyway!

wlthomson:
> 
> 
> Should the mock also get a info log?

Yup, I've updated all the logging in the data loader PR (you'll have to take my word for it... it isn't too far away now 😄 ).

src/database/repository/mock/item.rs
chris-morgan: (I ignore most notifications on this repository, but glance at a few, and this one caught my attention. All code and suggestions here are offered without warranty, and I haven’t tried running anything.)

I’d expect to defer the cloning as late as possible, so that you’re only cloning the items you need to clone. Something like:

```rust
for row in self.mock_data.lock().unwrap().values() {
    if let DatabaseRow::Item(item) = row {
        items.push(item.clone());
    }
}
```

Note also the switch away from `.for_each`; idiomatic Rust would normally prefer `for` to `.for_each`; in a few cases `.for_each` will optimise better, but `for` is more expressive due to `break`/`continue`/`return`/`?` (here’s some [good related reading](https://medium.com/@veedrac/rust-is-slow-and-i-am-the-cure-32facc0fdcb)). Also `.values()` rather than `.iter()`-and-throwing-away-the-key (`.into_iter()` also has its counterpart `.into_values()`).

But better than both in Rust (where convenient) is to manipulate an iterator and collect straight into the collection, in this case like this:

```rust
Ok(self.mock_data.lock().unwrap().values().filter_map(|item| {
    if let DatabaseRow::Item(item) = row {
        Some(item.clone())
    } else {
        None
    }
}).collect())
```

(Sometimes you have to give collect a hint, like `.collect::<Vec<_>>()`, since you can collect to many different types of collections, including things like `impl Iterator<Item = (K, V)>` → `HashMap<K, V>`, but here the return type lets it infer it easily.)

clemens-msupply (approved):
Probably good to use the iterator that Chris mentioned to collect the items.

src/database/repository/mock/item.rs
wlthomson: @chris-morgan Awesome, thanks!

wlthomson:
> 
> 
> Probably good to use the iterator that Chris mentioned to collect the items.

Done!

PR #255, opened 2021-08-23
#254 graphql filtering documentation

### LINKED ISSUE #254: Document GraphQL Filtering
- Add a directory for the graphql API
- Add a page for filtering/pagination/sorting
- Fill in the filtering section

joshxg:
Fixes #254 

- Went with just filtering on this page
- I just copy pasted stuff, I don't really know how this works, if there's stuff that shouldn't be there, feel free to remove!
- We should choose a language to write code examples in the documentation?
- If there are updates it is easier to just commit to the branch rather than go back and forward, please!

### DIFF
--- a/docs/content/docs/api/_index.md
+++ b/docs/content/docs/api/_index.md
@@ -1,0 +1,15 @@
++++
+title = "API"
+description = "Frontend Graphql API"
+date = 2025-05-01T18:00:00+00:00
+updated = 2021-05-01T18:00:00+00:00
+template = "docs/section.html"
+sort_by = "weight"
+weight = 1
+draft = false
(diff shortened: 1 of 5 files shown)


wlthomson (approved):
Looks about right to me. If we can get this signed off, I think we can move forward with a proof-of-concept implementation for a given query.

I imagine a lot of doc changes are going to happen incrementally. If possible, it would be good to group doc changes with the PRs where the functionality is changed (obviously this won't be possible until the docs move from being speculative to descriptive).

joshxg:
Thanks @wlthomson 

Will merge this, we can iterate if needed.

Issue #123, opened 2021-06-11
Update mock data


wlthomson:
Update mock data based on existing mSupply datafiles (anonymising/sanitising any identifying or otherwise sensitive data).

wlthomson:
With hindsight, I think a better process might be to update the mock data as tests are added for particular workflows. Closing.

PR #256, opened 2021-08-23
[No issue] Add linux support to the init script + update readme

clemens-msupply:
Not optimal, but spinning up Postgres in rootless mode doesn't seem to work...

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,14 +10,20 @@ mSupply remote server is a component of the Open mSupply system:
 
 ### Windows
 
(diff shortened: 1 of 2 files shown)


wlthomson (approved):
LGTM!

Obligatory pendantry: 

Ideally is nice to have a matching issue for PRs. Usual pattern we use is to append `#123` to the PR title and the branch name (where `123` is the issue number). Theres a few reasons for this: main one being that others have visibility of what you're working on, but even for small fixes it makes the git tree a lot nicer to parse and makes branch tidyup a lot easier.

Issue #268, opened 2021-08-26
Refactor repository return types


wlthomson:
Repository methods should abstract away details of the underlying data layer. Currently, repository methods return data types which map directly to the underlying database representation. As an example, the `item` repository should be able to return properties for a given item id without the caller knowing anything about the existence or details of the underlying `item`, `item_property` and `item_property` join tables.

wlthomson:
Closing as recent discussions have indicated repository refactoring is a no go!

PR #269, opened 2021-08-26
#263 Add item properties tables

### LINKED ISSUE #263: Add item properties tables
Add item property tables:

- `item_property_join`
- `property`

wlthomson:
Closes #263.

Have purposely avoided adding repository methods for `item_property_join` (see #268).


src/database/schema/item_property_join.rs
```diff
@@ -0,0 +1,9 @@
+use serde_json::Value as JsonValue;
+
+#[derive(Clone)]
+pub struct ItemPropertyJoinRow {
+    pub id: String,
+    pub item_id: String,
+    pub property_id: String,
+    pub value: JsonValue,
```
clemens-msupply, line 8: For both ItemPropertyRow and ItemPropertyJoinRow it would be good to have some comments for the structs members. For example, what it the item property key? What is the relationship between ItemPropertyRow and ItemPropertyJoinRow.

clemens-msupply, line 8: How is the JsonValue used later on? with sqlite in mind would it be easier to store it as a string and then parse it when needed?

clemens-msupply (approved):

migrations/20210825070620_create_item_property_join.sql
```diff
@@ -0,0 +1,8 @@
+-- Create item_property_join table
+
+CREATE TABLE item_property_join (
+    id varchar(255) NOT NULL PRIMARY KEY,
+    item_id varchar(255) NOT NULL,
+    property_id varchar(255) NOT NULL,
+    value jsonb NOT NULL
```
andreievg, line 7: the value is on the property

andreievg (changes requested):

migrations/20210825070620_create_item_property_join.sql
andreievg: I've dismissed the review, because I am actually not sure what the full intention of property is. Should be ok on property join for now

clemens-msupply: had problems with that as well though, `value` would make more sense to be in `item_property`?

migrations/20210825070620_create_item_property_join.sql
```diff
@@ -0,0 +1,8 @@
+-- Create item_property_join table
```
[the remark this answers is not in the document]
wlthomson: Yeah, that's definitely true (also duplicated in the filename!). It's more just trying to establish a convention that all migrations should have a top-level comment summarising what the migration does. Agree that for the initial migrations this is completely redundant, but it'll probably be a nice convention to stick with (also all migrations have these comments already, so there's not much point removing this one unless we remove them all)!

andreievg:
I would put this on hold until
A - We agree on schema
B - We set a milestone for it

wlthomson:
Converted to WIP as is on hold.

andreievg:
doing PR clean up, so closing, will know it's here for reference

PR #272, opened 2021-08-27
simple translate example

andreievg:
draft PR, to go along with https://github.com/openmsupply/org-issues/discussions/34, and to be open for comments


simple_translate/src/main.rs
```diff
@@ -0,0 +1,183 @@
...
+
+impl Mutatable for SimpleTranslatedRecord2 {
+    fn make_sync_mutation(&self, sync_type: &SyncType) -> Query {
+        match sync_type {
+            SyncType::Delete => Query {
+                query: "this will be a DELETE query, not a string",
+                record: Box::new(self),
+            },
+            _ => Query {
+                query: "this will be a UPDATE query, not a string",
```
clemens-msupply, line 72: Some general questions:

1) How do we actually access the central server? do we have direct access to the DB?

2) Are we always updating/deleting full rows or do we sometimes only sync some values in a row?

3) Should these queries be put into a "LegacyRepositories"? e.g. these repos can contain the code to retrieve and update data at the central server.

andreievg: > How do we actually access the central server? do we have direct access to the DB?

http request, can help you set up (mSupply 4D has rest API for this), as per postman collection I've sent on telegram yesterday.

> Are we always updating/deleting full rows or do we sometimes only sync some values in a row?

This would be ideal, eventually. There is no current plan or design for this (from remote back end or mSupply end). I guess we could future proof by first doing a query on the record, then merge incoming record into it and only after making a mutation.

> Should these queries be put into a "LegacyRepositories"? e.g. these repos can contain the code to retrieve and update data at the central server.

Not sure what you mean. Definitely some layer of abstraction would be good, so that when we swap to using new central server, where schema is aligned, we only need to worry about update/insert and delete and not the translations (which is something that is missing from this example)

andreievg: about that http request, data at this layer (integration layer), would come from a table (the actual data pulling layer would store it in a table)

andreievg:
already implemented

PR #277, opened 2021-08-30
#276 Add sorting docs

### LINKED ISSUE #276: Document GraphQL sorting
Need to document graphql sorting spec as defined in 

https://docs.google.com/document/d/1dBBPSCRLzJ7mJaU1RnNhg9YT_hJ3VbXo0HgABuNJUHg/edit

joshxg:
Fixes #276 

- Simple sorting!
- If there are any errors/mistakes/grammar issues please feel free to commit to the branch

### DIFF
--- a/docs/content/docs/api/sorting.md
+++ b/docs/content/docs/api/sorting.md
@@ -1,0 +1,52 @@
++++
+title = "Sorting"
+description = "Description of all API sorting"
+date = 2021-05-01T19:30:00+00:00
+updated = 2021-05-01T19:30:00+00:00
+draft = false
+weight = 51
+sort_by = "weight"
+template = "docs/page.html"
+
+[extra]
+toc = true
++++
+
+This API exposes a parameter on each query accepting an array of sort objects.
+
+### The sort object
+
+The sort object is two Key:Value pairs defining the field the result set is sorted by and an indicator of ascending or descending. Each entity is sortable by any of the fields of that entity.
+
+```
+{
+   key: [EntityField],
+   desc: Boolean,
+}
+```
+
+- key: Any field on the entity.
+- desc: Boolean indicator whether the sort should be descending.
+
+### Limitations
+
+The API currently supports an array of a single sort object.
+
+### Error handling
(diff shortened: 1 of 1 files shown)


docs/content/docs/api/sorting.md
```diff
@@ -0,0 +1,52 @@
...
+However, when a field does not exist, or the key value is null, an error will be returned.
+
+TODO: Shape of error
+
+### Examples
+
+```
+{
+    query transaction([{ key: "confirmDate", desc: true }]){
+        id,
```
andreievg, line 48: I think technical graphQL syntax is
```gql
{
    query { 
          transaction([{ key: "confirmDate", desc: true }]) {
            id,
             status
       }
    }
}
```
And zola syntax highlight seem to like the overall brackets to be omitted (like it's in the upstream doc PR)

Btw graphQL does it this way so that variables can be defined in`query(here)`, to then be use in entity arguments

andreievg (approved):

docs/content/docs/api/sorting.md
```diff
@@ -0,0 +1,52 @@
...
+This API exposes a parameter on each query accepting an array of sort objects.
+
+### The sort object
+
+The sort object is two Key:Value pairs defining the field the result set is sorted by and an indicator of ascending or descending. Each entity is sortable by any of the fields of that entity.
+
+```
+{
+   key: [EntityField],
+   desc: Boolean,
```
mark-prins, line 24: There was a question whether these are required or not. Definitely for `[EntityField]!` and the question was whether the client has to specify `desc` explicitly or not.

joshxg: Ah right. I made a comment in this PR about a better representation for this https://github.com/openmsupply/remote-server/pull/287#discussion_r699726307

joshxg:
Fixed (I think) gql syntax in #287 will merge this one and also do nicer object representations at some point in the future. Want to make sure doing them in typescript (v rust struct, for example) is OK

PR #289, opened 2021-08-31
#284 Make RepositoryError an enum

### LINKED ISSUE #284: Make RepositoryError an enum
We might want to have dedicated error types that we want to handle specially, e.g. "row already exists"; an enum makes this possible. 

For example:

```
pub enum RepositoryError {
    DatabaseError { msg: String },
    ...
}
```

clemens-msupply:
Closes #284.

Thanks to Chris M. for suggesting thiserror.

### DIFF
--- a/src/database/repository/diesel/mod.rs
+++ b/src/database/repository/diesel/mod.rs
@@ -52,7 +52,7 @@ impl From<DieselError> for RepositoryError {
                 };
                 format!("DIESEL_DATABASE_ERROR_{}", err_str)
             }
-            DieselError::NotFound => "DIESEL_NOT_FOUND".to_string(),
+            DieselError::NotFound => return RepositoryError::NotFound,
             DieselError::QueryBuilderError(_) => "DIESEL_QUERY_BUILDER_ERROR".to_string(),
             DieselError::DeserializationError(_) => "DIESEL_DESERIALIZATION_ERROR".to_string(),
             DieselError::SerializationError(_) => "DIESEL_SERIALIZATION_ERROR".to_string(),
--- a/src/database/repository/diesel/mod.rs
+++ b/src/database/repository/diesel/mod.rs
@@ -40,8 +40,12 @@ impl From<DieselError> for RepositoryError {
             DieselError::InvalidCString(_) => "DIESEL_INVALID_C_STRING".to_string(),
             DieselError::DatabaseError(err, _) => {
                 let err_str = match err {
-                    DieselDatabaseErrorKind::UniqueViolation => "UNIQUE_VIOLATION",
-                    DieselDatabaseErrorKind::ForeignKeyViolation => "FOREIGN_KEY_VIOLATION",
+                    DieselDatabaseErrorKind::UniqueViolation => {
+                        return RepositoryError::UniqueViolation
+                    }
+                    DieselDatabaseErrorKind::ForeignKeyViolation => {
+                        return RepositoryError::ForeignKeyViolation
+                    }
                     DieselDatabaseErrorKind::UnableToSendCommand => "UNABLE_TO_SEND_COMMAND",
                     DieselDatabaseErrorKind::SerializationFailure => "SERIALIZATION_FAILURE",
                     _ => "UNKNOWN",
@@ -57,14 +61,14 @@ impl From<DieselError> for RepositoryError {
             _ => "DIESEL_UNKNOWN".to_string(),
         };
 
-        RepositoryError { msg }
+        RepositoryError::DBError { msg }
     }
 }
 
 fn get_connection(
     pool: &Pool<ConnectionManager<DBBackendConnection>>,
 ) -> Result<PooledConnection<ConnectionManager<DBBackendConnection>>, RepositoryError> {
-    pool.get().map_err(|_| RepositoryError {
+    pool.get().map_err(|_| RepositoryError::DBError {
         msg: "Failed to open Connection".to_string(),
     })
 }
--- a/src/database/repository/pgsqlx/mod.rs
+++ b/src/database/repository/pgsqlx/mod.rs
@@ -34,7 +34,7 @@ impl From<sqlx::Error> for RepositoryError {
             sqlx::Error::Io(_) => "SQLX_ERROR_IO",
             sqlx::Error::Tls(_) => "SQLX_ERROR_TLS",
             sqlx::Error::Protocol(_) => "SQLX_ERROR_PROTOCOL",
-            sqlx::Error::RowNotFound => "SQLX_ERROR_ROW_NOT_FOUND",
+            sqlx::Error::RowNotFound => return RepositoryError::UniqueViolation,
             sqlx::Error::ColumnIndexOutOfBounds { index: _, len: _ } => {
                 "SQLX_ERROR_COLUMN_INDEX_OUT_OF_BOUNDS"
             }
@@ -51,7 +51,7 @@ impl From<sqlx::Error> for RepositoryError {
             _ => "SQLX_ERROR_UNKNOWN",
         });
 
-        RepositoryError { msg }
+        RepositoryError::DBError { msg }
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2850,6 +2850,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "sqlx",
+ "thiserror",
  "tokio 0.2.25",
  "uuid",
 ]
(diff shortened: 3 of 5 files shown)


src/database/repository/diesel/mod.rs
```diff
@@ -57,14 +61,14 @@ impl From<DieselError> for RepositoryError {
             _ => "DIESEL_UNKNOWN".to_string(),
         };
 
-        RepositoryError { msg }
+        RepositoryError::DBError { msg }
```
andreievg, line 64: Shouldn't we just store the errors ? And implement our own display or debug on RepositoryError (in case we are passing it on to front end) ?

See SettingError
```rust
pub enum RepositoryError {
    Diesel(DieselError),
    Sqlite(...)
}
...
fn from(err: DieselError) -> Self {
       Diesel(err)
}
..
// or debug or own trait for error response to front end
impl Display for RepositoryError {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result {
        match { 
              Sqlite(err) => ..,
              Diesel(err)=> ...

```

clemens-msupply: Not sure if we would ever need the original error? the error probably not even leaves the backend and gets mapped to an error message that is returned in GraphQL...

Since the Repository is the layer between DB/ORM and the rest of the app I wouldn't actually "leak" the DB specific errors out of it. (Event though the DBError msg is DB/ORM specific but I think that is a reasonable compromise)

My main motivation was to be able to define some RepositoryErrors that are DB/ORM independent and can be handled in our code, e.g. RepositoryError::NotFound. The RepositoryError::DBError is just a bucket for all other errors.

andreievg: > Not sure if we would ever need the original error?

I think we would, I would assume we will have logging (of full error). And we might decide to send more information to the front end. 

Our operations outside of API may also consume a RepositoryErrors differently to API.

Although I've made that comment without fully looking at context, I'll have a deeper look (just thought that remapping is best done at the error consumption vs  propagation).

chris-morgan: Rule of thumb: keep the original error objects as long as possible, and avoid converting to strings unnecessarily. Such conversions are generally lossy (for example, the error may have another error as context, which applied recursively is similar to a stack trace) and take extra effort, while typically not making anything easier.

clemens-msupply: The point of the the Repository is to separate the DB from the rest of the code. Thus, we shouldn't use the original error for our business logic, e.g. if we decide to change ORM or DB our error handling breaks.

However, if we are just talking about debugging, I would say yes we can keep the original value in a our enum, e.g. in an source or origin field.

andreievg: ok, if we going for full abstraction (on top of ORM), then I guess repository error can remap everything from diesel error into it's own form. Including all other data from diesel error (all messages that would be needed upstream, i.e. DatabaseErrorInformation content). <- I guess this can be done on per case scenario.

or we could just store diesel error an implement 'from' for various consumers. (from traits would need to be re-implemented if we swap DB or ORM)

up to you.

I agree with Chris that string constants are pretty much redundant in rust, we can map from enums to string representation at the layer that need to serialize them (API error return etc..) <- that would be good to do in this PR

wlthomson: I agree with @clemens-msupply re. repositories providing a database agnostic interface to the consumer, but I also agree that errors are a bit of an exception in that we want to expose as much implementation detail as we can. I think the idea of a `source` field would do the job nicely (I assume this would be stored as a trait object or an enum of its own?).

src/database/repository/pgsqlx/mod.rs
```diff
@@ -34,7 +34,7 @@ impl From<sqlx::Error> for RepositoryError {
             sqlx::Error::Io(_) => "SQLX_ERROR_IO",
             sqlx::Error::Tls(_) => "SQLX_ERROR_TLS",
             sqlx::Error::Protocol(_) => "SQLX_ERROR_PROTOCOL",
-            sqlx::Error::RowNotFound => "SQLX_ERROR_ROW_NOT_FOUND",
+            sqlx::Error::RowNotFound => return RepositoryError::UniqueViolation,
```
wlthomson, line 37: Hmm, this is a little unintuitive to me it would be nice if we could handle these cases in a separate expression... maybe with an `if let`?

wlthomson (approved):

src/database/repository/pgsqlx/mod.rs
clemens-msupply: Didn't like it much either... What would you do? Only thing I could think of is transforming everything directly into a `RepositoryError`, e.g. `RepositoryError::DbError { msg: "SQLX_ERROR_ROW_NOT_FOUND".to_string() }`

src/database/repository/diesel/mod.rs
clemens-msupply: Ok lets keep this discussion in mind when we do some error handling. I will just merge these changes for now since I think its at least a slight improvement to the previous error struct.

src/database/repository/pgsqlx/mod.rs
wlthomson: I was just thinking of something uncomplicated like:

```
impl From<sqlx::Error> for RepositoryError {
    fn from(err: sqlx::Error) -> Self {
        if let err = sqlx::Error::RowNotFound {
            return RepositoryError::UniqueViolation;
        }

        let msg = String::from(match err {
            sqlx::Error::Configuration(_) => "SQLX_ERROR_CONFIGURATION",
            sqlx::Error::Database(_) => "SQLX_ERROR_DATABASE",
            sqlx::Error::Io(_) => "SQLX_ERROR_IO",
            sqlx::Error::Tls(_) => "SQLX_ERROR_TLS",
            sqlx::Error::Protocol(_) => "SQLX_ERROR_PROTOCOL",
            sqlx::Error::ColumnIndexOutOfBounds { index: _, len: _ } => {
                "SQLX_ERROR_COLUMN_INDEX_OUT_OF_BOUNDS"
            }
            sqlx::Error::ColumnNotFound(_) => "SQLX_ERROR_COLUMN_NOT_FOUND",
            sqlx::Error::ColumnDecode {
                index: _,
                source: _,
            } => "SQLX_ERROR_COLUMN_DECODE",
            sqlx::Error::Decode(_) => "SQLX_ERROR_DECODE",
            sqlx::Error::PoolTimedOut => "SQLX_ERROR_POOL_TIMED_OUT",
            sqlx::Error::PoolClosed => "SQLX_ERROR_POOL_CLOSED",
            sqlx::Error::WorkerCrashed => "SQLX_ERROR_WORKER_CRASHED",
            sqlx::Error::Migrate(_) => "SQLX_ERROR_MIGRATE",
            _ => "SQLX_ERROR_UNKNOWN",
        });

        RepositoryError::DBError { msg }
    }
}
```

wlthomson: For now, I was just thinking of something uncomplicated like:

```
impl From<sqlx::Error> for RepositoryError {
    fn from(err: sqlx::Error) -> Self {
        if let err = sqlx::Error::RowNotFound {
            RepositoryError::UniqueViolation
        } else {
	        let msg = String::from(match err {
	              sqlx::Error::Configuration(_) => "SQLX_ERROR_CONFIGURATION",
	              sqlx::Error::Database(_) => "SQLX_ERROR_DATABASE",
	              sqlx::Error::Io(_) => "SQLX_ERROR_IO",
	              sqlx::Error::Tls(_) => "SQLX_ERROR_TLS",
	              sqlx::Error::Protocol(_) => "SQLX_ERROR_PROTOCOL",
	              sqlx::Error::ColumnIndexOutOfBounds { index: _, len: _ } => {
	                  "SQLX_ERROR_COLUMN_INDEX_OUT_OF_BOUNDS"
	              }
	              sqlx::Error::ColumnNotFound(_) => "SQLX_ERROR_COLUMN_NOT_FOUND",
	              sqlx::Error::ColumnDecode {
	                  index: _,
	                  source: _,
	              } => "SQLX_ERROR_COLUMN_DECODE",
	              sqlx::Error::Decode(_) => "SQLX_ERROR_DECODE",
	              sqlx::Error::PoolTimedOut => "SQLX_ERROR_POOL_TIMED_OUT",
	              sqlx::Error::PoolClosed => "SQLX_ERROR_POOL_CLOSED",
	              sqlx::Error::WorkerCrashed => "SQLX_ERROR_WORKER_CRASHED",
	              sqlx::Error::Migrate(_) => "SQLX_ERROR_MIGRATE",
	              _ => "SQLX_ERROR_UNKNOWN",
	          });
        	RepositoryError::DBError { msg }
		}
    }
}
```

PR #291, opened 2021-09-01
#285 remove sqlx repositories

### LINKED ISSUE #285: Remove sqlx repositories
Finalising migration away from sqlx now that all sqlx repositories have been reimplemented using deisel (see https://github.com/openmsupply/remote-server/pull/266). 

Seemed to be agreement that there is no benefit in duplicating future workloads by writing duplicate SQL and query builder statements, and can't see much to gain from retaining a bunch of dependencies and bloating our binary for a subset of functionality we have no plan to use (and can always cherry-pick from the commit history if we decide some of the retired code has some currently unforesee utility!).

wlthomson:
Closes #285.

<strike>Note that this breaks existing database tests which are using sqlx for migrations! If this PR isn't too big already, can migrate in this PR to use diesel migration functionality in place of the old sqlx migration, or can do that in a follow-up PR branched off these changes.<strike/>


wlthomson:
> 
> 
> Take in mind that the tests are using sqlx for migrations... Maybe it was just me but I had some problems with the Diesel migration macro, e.g. it required this global diesel macro and crashed rust-analyser...

Yup, that's what I was referring to with:

> Note that this breaks existing database tests which are using sqlx for migrations! 

Would you prefer me to address that in this PR or a separate follow-up?

clemens-msupply:
Think it would be good to try out the diesel migration macro first because if you don't like it either we would have to reshuffle the migration files again to work with sqlx...

wlthomson:
@clemens-msupply Cool. Have read up a bit and experimented and it seems to me that the difficulty may be due to attempting to embed the migrations in the binary. This seems like pretty cool functionality, but probably more than necessary atm.

Ideally don't want to keep sqlx in the repo just for migrations, but if we decide to go that way hopefully we can avoid having to bloat the binary with sqlx functionality we don't use or want!

clemens-msupply:
if I remember correctly Diesel requires to include the migration files in the main file (?)

sqlx can do it in the test, is the Rust compiler smart enough to keep sqlx completely out of the release binary?

wlthomson:
> if I remember correctly Diesel requires to include the migration files in the main file (?)

I'm f hoping that this is only the case if the migrations are embedded in the binary. Currently researching if it's possible to avoid that and just read directly from file (like with sqlx).

wlthomson:
> 
> 
> if I remember correctly Diesel requires to include the migration files in the main file (?)
> 
> sqlx can do it in the test, is the Rust compiler smart enough to keep sqlx completely out of the release binary?

Took a bit of experimentation (and reading some of the diesel source... the diesel migration docs are  a bit lacking), but have managed to replicate the existing test database setup using (unembedded) diesel migrations!

migrations/postgres/00000102_create_user_account/up.sql
```diff
@@ -0,0 +1,11 @@
+-- Create user_account table.
+
+CREATE EXTENSION pgcrypto;
```
clemens-msupply, line 3: Needed?

migrations/postgres/00000101_diesel_initial_setup/up.sql
```diff
@@ -0,0 +1,36 @@
...
+BEGIN
+    IF (
+        NEW IS DISTINCT FROM OLD AND
+        NEW.updated_at IS NOT DISTINCT FROM OLD.updated_at
+    ) THEN
+        NEW.updated_at := current_timestamp;
+    END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
```
clemens-msupply, line 36: What do we use updated_at for? what is the equivalent in sqlite?

tests/repository/test_db.rs
```diff
@@ -1,48 +1,73 @@
...
+    let mut migrations_dir =
+        find_migrations_directory().expect("Failed to locate migrations directory");
+    migrations_dir.push(MIGRATION_PATH);
+
+    let mut migrations = mark_migrations_in_directory(&connection, &migrations_dir).unwrap();
+    migrations.sort_by(|(m, ..), (n, ..)| m.version().cmp(&n.version()));
+
+    for (migration, ..) in migrations.iter() {
+        migration.run(&connection).unwrap();
+    }
```
clemens-msupply, line 72: The migration part could be put into its own function...

clemens-msupply (approved):
Looks good to me.

There is only the initial_setup file I am unsure about...

clemens-msupply:
> Took a bit of experimentation (and reading some of the diesel source... the diesel migration docs are a bit lacking), but have managed to replicate the existing test database setup using (unembedded) diesel migrations!

nice!

migrations/postgres/00000101_diesel_initial_setup/up.sql
wlthomson: This is automatically generated by diesel for pg but not for sqlite. Pretty sure it's unneeded but kept it in for now just to be safe 😆.

migrations/postgres/00000102_create_user_account/up.sql
wlthomson: May as well, can't hurt to have a bit of additional security!

clemens-msupply: Yes we should definitely hash passwords. Since we have sqlite as an alternative DB we have to do it outside of the DB...

https://github.com/openmsupply/remote-server/issues/270

tests/repository/test_db.rs
wlthomson: Have spent a bit of time trying to, and struggling with the type definitions. Don't think it's a prority atm, but definitely go ahead and do a PR for it if you can figure it out (I'll have another go when I've got some higher priority things done)!

clemens-msupply: Yes its tricky, I had a chat with @chris-morgan about a similar issue, i.e. having a generic repo that takes a diesel Connection. The conclusion was: there is some compiler limitation/bug (?) that make type interference harder than necessary but it should be possible somehow (I gave up to follow up on it though...)

chris-morgan: For the record, what we did there was on the [playing-with-types branch](https://github.com/openmsupply/remote-server/commits/playing-with-types), ending up with a nice `MyConnection` trait that you can parameterise over without excessive verbosity. https://github.com/openmsupply/remote-server/commit/cbed4bc333ecfe1e143eb62139641c6311133d62 is the specific commit working around the unexpected rustc limitation (which turns out to be a [long-standing issue](https://github.com/rust-lang/rust/issues/20671) that people keep on tripping over (just *count* the duplicates and references!), but which can fortunately be worked around without much difficulty for associated types).

Issue #281, opened 2021-08-31
[SYNC] - Simple Translation Full - M1


andreievg:
# Description

Add enough table and simple translations for them to satisfy M1 spec

### Scope

<ins>Includes </ins>
* item - (code)
* name - (code, is_customer, is_supplier)
* store - (code)
* ~~item_store_join~~
* ~~name_store_join~~ (would be part of remote sync, one way from central only)
* master_list_join
* master_list
* master_list_line

`list to be revised based on front end specs`

<ins>Excludes</ins>
* Api
* Complex translation in name_store_join

### Tasks

- [ ] Break down tasks into digestible (by reviewer) units.
- [ ] MW job added

clemens-msupply:
Is it guaranteed that the sync records come in in order from the central server? e.g. if a store refers to a name is it guaranteed that this name comes before the store in the received sync list?

andreievg:
No guarantee:

https://github.com/openmsupply/org-issues/discussions/33

Under 'Atomic Batched Sync'. Two ways to solve it:

* Transaction with defferrred fk check (which only work in postgres, there is an alternative in sqlite, but it's not good: https://github.com/openmsupply/org-issues/discussions/33#discussioncomment-1232353
* Order records in transaction in order of reference
     * Need to have a list of reference priority (which i what i was suggesting for what we call 'migrations' (database schma initilisation scripts)
     * This can either be hard coded, or we can calculate it (in the snapshot branch i've sent there is this: `calculate_table_order` method that does it, and something similar in mFlow project, from this [sql](https://github.com/openmsupply/application-manager-server/blob/develop/database/buildSchema/47_database_info.sql), generates type object in this [shape](https://github.com/openmsupply/application-manager-server/blob/a147d9bc656cfce02b64b411771d0cfdca16718f/src/components/exportAndImport/types.ts#L1-L24), and then [order is by reference](https://github.com/openmsupply/application-manager-server/blob/a147d9bc656cfce02b64b411771d0cfdca16718f/src/components/exportAndImport/getDatabaseInfo.ts#L112-L127)

Keen to discuss this today in sprint meeting.

Btw i didnt' quite spec this out properly because I wasn't aware that sqlite doesn't have the same deferred fk check functionality in transacitons.

I think we would need to query buffer queue in batches of table_name (from highest to lowest)

andreievg:
@clemens-msupply i think this one was closed by accident ? It should be the mechanism one ? (we can look over it at sprint meeting)

[...]

PR #292, opened 2021-09-02
#281  Api pattern pagination

andreievg:
Partial fix to #281

As per discussion, just using first and offset 

### Changes

* Use first and offset
* As page object rather then individual argument

### DIFF
--- a/docs/content/docs/api/patterns.md
+++ b/docs/content/docs/api/patterns.md
@@ -12,22 +12,53 @@ template = "docs/page.html"
 toc = true
 +++
 
+
+# Lists
+
+Lists query and result have different shapes of request and response and lists share three generic arguments,
+
+<ins>Query</ins>
+
+For example query node, `invoices`
+
+```graphqls
+type query {
+  invoices(sort: InvoiceSort, filter: InvoiceFilter, page: Page): {
+      totalCount: Int,
+      nodes: Invoice
+  }
+}
+```
+
+[Sort](/docs/api/patterns/#sorting) [Filter](/docs/api/patterns/#filtering) [Page](/docs/api/patterns/#pagination)
+
+
 # Sorting
 
-This API exposes a parameter on each query accepting an array of sort objects.
+This API exposes a parameter on each [list](/docs/api/patterns/#lists) query accepting an array of sort objects.
 
 **The sort object**
 
 The sort object is two Key:Value pairs defining the field the result set is sorted by and an indicator of ascending or descending. Each entity is sortable by any of the fields of that entity.
 
-```
-{
-   key: [EntityField],
-   desc: Boolean,
+For example, `InvoiceSort`
+
+```graphqls
+type InvoiceSort: [InvoiceSortOption]
+
+type InvoiceSortOption {
+   key: [InvoiceSortField]
+   desc: Boolean
+}
+
+enum InvoiceSortOptions {
+  CODE
+  OTHER_PARTY_NAME
+  ...
 }
 ```
 
-- key: Any field on the entity.
+- key: Any scalar field on the entity in standard graphQL enum value case (capital snake).
 - desc: Boolean indicator whether the sort should be descending.
 
 **Limitations**
@@ -36,7 +67,7 @@ The API currently supports an array of a single sort object. (will apply only on
 
 **Error handling**
 
-NULL and empty sort fields will result in no filtering and will not return an error.
+Empty/undefined sort fields will result in no filtering and will not return an error.
 
 However, when a field does not exist, or the key value is null, an error will be returned.
 
@@ -44,9 +75,9 @@ TODO: Shape of error
 
 **Examples**
 
-```
+```graphql
 query { 
-    transaction([{ key: "confirmDate", desc: true }]) {
+    transaction(filter: [{ key: CONFIRM_DATE, desc: true }]) {
         id
         status
     }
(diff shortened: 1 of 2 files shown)


andreievg:
@joshxg @mark-prins small one for you to review

docs/content/docs/api/patterns.md
```diff
@@ -212,19 +212,33 @@ query {
...
+
+`first > 0 && first <= 1000` (*exception**)
+
+`offset >= 0`
 
 *exception** limit is NOT set for inner lists (like lines in a transaction) {TODO or should it be ? or should each list just have a pagination limit ?}
 
+Offset above row range will not result in an error or rows.
+
+`totalCount` is added as another field at the same level as `nodes` in the [list](/docs/api/patterns/#lists-and-pagination)
```
joshxg, line 240: I was a little confused - this is more of a 'list' concept rather than pagination?

docs/content/docs/api/patterns.md
```diff
@@ -212,19 +212,33 @@ query {
...
+   first: Int,
+   offset: Int
+}
+```
+
+Both fields are optional, with default values: `first` = 100, `offset` = 0.
+
+Values are restricted by these rules:
+
+`first > 0 && first <= 1000` (*exception**)
```
joshxg, line 232: Will flag that I think 1000 seems limiting but can be adjusted later.

For example I would be interested in determining if it would be viable for selecting a supplier to perform a query such as (I don't remember syntax :)):

```gql
query name(filter: {isSupplier: true}){
    id
    code
    name
}
```

Receive say, ~3000 names, and filter this list within JS. Rather than performing debounced network requests like:

```gql
query name(filter: {isSupplier: {equalTo: true}, name: {contains: "a"}} }){
    id
    code
    name
}
```

but this would need some research

docs/content/docs/api/patterns.md
```diff
@@ -212,19 +212,33 @@ query {
 }
 ```
 
-**Pagination**
+# Pagination
```
joshxg, line 215: The headers are now `Lists and Pagination` and then below is `Pagination`. Would be nice to clean this up and to have List be a top level nav unit, with all list specific related information and have each of the list related concepts listed below

docs/content/docs/api/patterns.md
joshxg: Also would be good to use ts (or rust structs) to document shapes

-----

The API exposes a pagination object parameter on each list query with the following shape:

```ts
interface PaginationObject {
  first?: number; // The number of records to return.
  offset?: number; // The offset number of records to to return from.
}
```

Note: `first > 0 && first <= 1000` & `offset >= 0`

and 

List responses have the following shape:

```ts
interface ListResponse<Entity> {
  nodes: Entity[];
  totalCount: number; // The total number of records matching the applied filter
}
```

Could use gql but I think this falls over with generics

```gql
type PaginationObject {
    first: Int
    offset: Int
}
```

```gql
type ListResponse<Entity> {
  nodes: Entity[];
  totalCount: number; // The total number of records matching the applied filter
}
```

joshxg (approved):
Great, thanks.

Sorry for the delay. Are you able to use the request review feature in the future?

docs/content/docs/api/patterns.md
andreievg: I guess it depend on the number (and record size, and how many things you want to search by). Above example would seem better done in JS.  

If we had updatedDatetime on name record, you can just always get all names > updatedDatetime (current datetime), and make sure you keep existing ones and merge new ones in there. 

And maybe only do DB search if pageLimit is exceeded (i would assume both would be implemented anyways, and choosing between the two would be trivial ?).

> Will flag that I think 1000 seems limiting but can be adjusted later

We best move this to configuration in implementation

docs/content/docs/api/patterns.md
andreievg: good suggestions thanks, relocated and added schema definitions (for all patterns):
3befbe7fb68d2af49593bba7840e0afb9cf945cd

docs/content/docs/api/patterns.md
andreievg: Also done here, cheers: 3befbe7fb68d2af49593bba7840e0afb9cf945cd

docs/content/docs/api/patterns.md
andreievg: Was hoping we could use graphiql for more descriptive schema docs (fully typed), but I guess it's good to be concrete in shapes  here too.

andreievg:
@joshxg with this commit: https://github.com/openmsupply/remote-server/commit/3befbe7fb68d2af49593bba7840e0afb9cf945cd, sort query needs to use enums (if we are going with standards). It might be better to use this suggestion, seems easier on front end: https://github.com/openmsupply/remote-server/pull/287#discussion_r699863061

joshxg:
Committed some minors, like:
- `transaction` -> `invoice` everywhere
- `graphqls` -> `graphql` spelling
- Some sentences to make more sense to me (i'm shelfish)
- Some consistency in the structure
- Some url fixes
- Added a explanation to examples, I thought it was helpful

joshxg:
> @joshxg with this commit: 3befbe7, sort query needs to use enums (if we are going with standards). It might be better to use this suggestion, seems easier on front end: #287 (comment)

This would make multicolumn sorting harder (ordering) and I'm not seeing that it's easier. Also my preference is the other one, but I don't really mind.

> And you'll note I've changed example for datetime filtering,

Yes I'm not sure why - can revisit

andreievg:
@joshxg 

> Committed some minors, like:
> transaction -> invoice everywhere
> Some sentences to make more sense to me (i'm shelfish)
> Added a explanation to examples, I thought it was helpful

Great, thanks

> graphqls -> graphql spelling

graphqls was listed as one of valid languages in zola docs, I thought there would be differentiation between schema highlight and query highlight (there isn't lol)

> Some consistency in the structure

There are some example headers now in bold and some underline, i'll change it in upcoming PR.

> Some url fixes

Couldn't see them, was there something left to commit ?

> This would make multicolumn sorting harder (ordering) and I'm not seeing that it's easier.

I didn't think about multi sort, that's true. I was referring to not needing to map enums to column names.

>  Also my preference is the other one, but I don't really mind.

>> And you'll note I've changed example for datetime filtering,

> Yes I'm not sure why - can revisit

I think we are choosing convention over preferences: https://github.com/openmsupply/remote-server/pull/287#discussion_r700584279

Last note (back pedal on chat in telegram), I think I would rather a draft PR (based of a PR) or suggested changes, when multiple devs commit changes spanning beyond typo/syntax in the same PR/branch, just makes it easier for PRer to know what they are merging at the end.

PR #293, opened 2021-09-02
#281 Specify arguments in api examples

andreievg:
Partial fix to #281

I couldn't find any GraphQL api specs that just takes an object as argument. Typical GraphQL clients would take `variables` object as an object and pass each key in variables as parameter to corresponding argument

### Changes

* Add arguments to example queries
* Change filter and sorting to apply to lists only

@joshxg you had a question about both of those.


joshxg (approved):

andreievg:
closing this since downstream PR now has this changes (since there was a request to document schema, decided to add this too)

PR #298, opened 2021-09-06
281- Implement first translation version

clemens-msupply:
Here is my work in progress for the sync translations. There are two issues that are somewhat unrelated to the ticket but I would like to get some feedback on:

1) I would like to reuse existing Repositories for the sync transactions, e.g. use multiple repository calls in the same transaction. I did a proof of concept for the NameRepository. There are currently two versions of a generic NameRepositoryImp, one that works with a connection pool (as before) and one that works with a reference to an ongoing transaction. This works fine but I find the code a bit cumbersome/verbose. The blocker for me to make it a bit more concise is in:

```
pub type RepositoryMap = Map<AnyRepository>;
pub type AnyRepository = dyn CloneAny + Send + Sync;
```
I.e. the Repository has to be Clone + Send + Sync which is not possible for a reference to a transaction. My Question is, do we need to have it Send + Sync? If yes I could wrap the transaction reference into a Mutex and make the code look a bit nicer... other ideas?

2) Question for @chris-morgan : In TxManager I had to introduce a create_tx() method to work around a lifetime problem related to async functions. Any idea how to avoid it? (see todo comment)

### DIFF
--- a/src/util/sync/translation.rs
+++ b/src/util/sync/translation.rs
@@ -1,0 +1,241 @@
+use crate::{
+    database::{
+        repository::{new_tx_name_repository, DBTransaction, RepositoryError, TxManager},
+        schema::NameRow,
+    },
+    server::data::RepositoryRegistry,
+};
+
+use async_trait::async_trait;
+use serde::Deserialize;
+
+#[derive(Debug)]
+pub enum SyncType {
+    Delete,
+    Update,
+    Insert,
+}
+
+#[derive(Debug)]
+pub struct SyncRecord {
+    sync_type: SyncType,
+    record_type: String,
+    data: String,
+}
+
+#[async_trait]
+trait Mutatable {
+    async fn make_sync_mutation(&self, sync_type: &SyncType, tx: &DBTransaction);
+}
+
+#[derive(Deserialize)]
+struct LegacyNameTable {
+    #[serde(rename = "ID")]
+    id: String,
+    name: String,
+    fax: String,
+    phone: String,
+    customer: bool,
+    bill_address1: String,
+    bill_address2: String,
+    supplier: bool,
+    #[serde(rename = "charge code")]
+    charge_code: String,
+    margin: i64,
+    comment: String,
+    currency_ID: String,
+    country: String,
+    freightfac: i64,
+    email: String,
+    custom1: String,
+    code: String,
+    last: String,
+    first: String,
+    title: String,
+    female: bool,
+    date_of_birth: String,
+    overpayment: i64,
+    group_ID: String,
+    hold: bool,
+    ship_address1: String,
+    ship_address2: String,
+    url: String,
+    barcode: String,
+    postal_address1: String,
+    postal_address2: String,
+    category1_ID: String,
+    region_ID: String,
+    #[serde(rename = "type")]
+    table_type: String,
+    price_category: String,
+    flag: String,
+    manufacturer: bool,
+    print_invoice_alphabetical: bool,
+    custom2: String,
+    custom3: String,
+    default_order_days: i64,
+    connection_type: i64,
+    //PATIENT_PHOTO": "[object Picture]",
+    NEXT_OF_KIN_ID: String,
+    POBOX: String,
+    ZIP: i64,
+    middle: String,
+    preferred: bool,
+    Blood_Group: String,
+    marital_status: String,
+    Benchmark: bool,
+    next_of_kin_relative: String,
+    mother_id: String,
+    postal_address3: String,
+    postal_address4: String,
+    bill_address3: String,
+    bill_address4: String,
+    ship_address3: String,
+    ship_address4: String,
+    ethnicity_ID: String,
+    occupation_ID: String,
+    religion_ID: String,
+    national_health_number: String,
+    Master_RTM_Supplier_Code: i64,
+    ordering_method: String,
+    donor: bool,
+    latitude: i64,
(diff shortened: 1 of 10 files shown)


andreievg:
Thanks, I'll check it out today sometime. Just had a quick scroll, saw a lot of fields listed, the plan was to do just M1 translations thanks (https://docs.google.com/spreadsheets/d/1TLxjqZ96o67aScQ0abfQ5juzRPr7dZ1Y1vnwdElv3XY/edit#gid=719069601)

wlthomson:
> I would like to reuse existing Repositories for the sync transactions, e.g. use multiple repository calls in the same transaction.

wlthomson:
> I would like to reuse existing Repositories for the sync transactions, e.g. use multiple repository calls in the same transaction.

wlthomson:
> I would like to reuse existing Repositories for the sync transactions, e.g. use multiple repository calls in the same transaction. 

Hm, I see what you`re trying to do here... but I think there may be some misunderstandings of the repo pattern :(. Probably my fault for not clarifying the pattern. 

The idea I had with the repo pattern was to separate the data-layer from the persistence-layer by abstracting away concerns specific to the underlying storage (usually a database). This is a tricky one, as it maps more to a domain rather than an entity, i.e. the sync domain. I think that's understandably why you have taken the approach of reusing the repos, as they currently all map to db tables.

The idea I would suggest here is to design a specific repository to handle this use-case (it could also handle sync cursor and sync buffer). The problem with the suggested approach is that it tightly binds repositories together. Ideally, repositories shouldn't need to depend on each other, and no state should persist between them. 

I think instead of trying to combine repositories to get functionality, you want to be working at the data layer rather than the persistence layer. The giveaway here is that you are trying to use a repository that maps to a db concept rather than a domain or entity (the concept of a transaction is data-layer specific, it shouldn't be something the calling code knows about).

Is there a reason you want to reuse existing repository methods rather than having a `SyncRepository` that handles these use-cases and keeping the transaction logic there? I think it would be a lot simpler and concistent with the repository design pattern.

Happy to give more details or work on an example if helpful... I think it's important to get this pattern right at an early stage.

@clemens-msupply @andreievg Thoughts?

@chris-morgan I recall you doing some research on the repository pattern when you we're in the office! Do you have any thoughts on this?

src/util/sync/translation.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+pub async fn import_sync_records(
+    registry: &RepositoryRegistry,
+    records: &Vec<SyncRecord>,
+) -> Result<(), String> {
+    let tx_manager = registry.get::<TxManager>();
+    let tx = tx_manager.create_tx().unwrap();
+    tx_manager
+        .transaction(&tx, || async {
```
andreievg, line 168: why does this have to be async ? (this method `import_sync_records` I would assume that sync operations would already happen in async space (whatever runtime decided that is), so these operations would be non blocking for the rest of the app.

I might be missing something, so please do correct me if i am wrong

andreievg:
@clemens-msupply I might be missing something, but would this work just as well ? 
https://github.com/openmsupply/remote-server/compare/281-simple-translation...281-example

We would need just one extra repository to do the translations ?

src/util/sync/translation.rs
clemens-msupply: It needs to be async because in the transaction I call Repos that have an async interface.

I tried to have `import_sync_records ` sync and then spin up another Runtime in the transaction but tokio doesn't like to have nested Runtimes...

Other option is to make repos sync since Diesel is sync. But this would make it very hard if we ever want to go back.

clemens-msupply:
> @clemens-msupply I might be missing something, but would this work just as well ?
> [281-simple-translation...281-example](https://github.com/openmsupply/remote-server/compare/281-simple-translation...281-example)
> 
> And then the repository is just to run transactions (also keeps translation logic out of repositories).
> [281-simple-translation...281-example-2](https://github.com/openmsupply/remote-server/compare/281-simple-translation...281-example-2)

Yes we can do that. The reason I didn't went for it is that you still have to duplicate a method to make them able work in a transaction. Another concern is that `integrate_records` might be to restrictive to handle more complex translations?

clemens-msupply:
@wlthomson I am simply using existing repos to do the translations, i.e. so that I don't have to think in DB terms. Basically I am just making use of the Repos (thats what they are for). Translation/sync is one layer up.

Regarding transactions, I think it would be very useful to have a concept of transactions outside of the DB (discussion https://github.com/openmsupply/remote-server/discussions/300). If you ignore transactions the repository abstraction becomes a bit leaky, i.e. you need methods like `get_items_and_item_lines_sync()` or `do_all_translations()` to ensure data consistency.

Think we should shift the discussion to a zoom meeting, lets talk about it at 10.

andreievg:
> Yes we can do that. The reason I didn't went for it is that you still have to duplicate a method to make them able work in a transaction. 

No methods are duplicated, there is a sync version that's called inside async one. Seems way less code

> Another concern is that integrate_records might be to restrictive to handle more complex translations?

I would have thought they will already be translated by then. Btw our business logic (like translations), shouldn't be inside repos afaik

wlthomson:
@clemens-msupply random nit: if you include "Closes #298" in the PR description it will automatically close the associated issue when the PR is merged.

clemens-msupply:
Since the code derived too much from the initial version I will create a new PR...

PR #303, opened 2021-09-06
#302 Fix mock repository error handling

### LINKED ISSUE #302: Fix mock repository error handling
Running app with feature `mock` enabled results in compilation errors related to recent changes to the `RepositoryError` struct (https://github.com/openmsupply/remote-server/pull/289).

### LINKED ISSUE #309: Regenerate diesel schema
Diesel schema needs to be regenerated in order to add new tables. Code needs to be updated as running `diesel setup` currently causes compilation errors.

wlthomson:
Closes #302 (apologies for typo in branch name!).


andreievg:
I wonder if we need mock repository anymore. I think there is a lot of overhead in maintaining three repos, and we should always check against db, it's not like we can skip that step (so not sure what use mock repo would be).

wlthomson:
> 
> 
> I wonder if we need mock repository anymore. I think there is a lot of overhead in maintaining three repos, and we should always check against db, it's not like we can skip that step (so not sure what use mock repo would be).

Hm, I've found the mock db really useful for prototyping and debugging resolvers. If we only want to run integration tests against live dbs that's all good, but I would still see a place for the mock repos in assisting graphql and sync debugging. Having said that I am likely biased as I put a lot of work into them, so if you and @clemens-msupply agree on getting rid of them, I won't protest!

clemens-msupply (approved):
I think you can do everything using sqlite (either on file or :memory:) that you can do with the mocks, right? If yes, I would remove it as well...

I currently don't delete the sqlite test db files after the test (probably need some cleanup in the future) which makes it easy to debug as well...

wlthomson:
> 
> 
> I think you can do everything using sqlite (either on file or :memory:) that you can do with the mocks, right? If yes, I would remove it as well...
> 
> I currently don't delete the sqlite test db files after the test (probably need some cleanup in the future) which makes it easy to debug as well...

Yeah, that was my thinking. The advantage of mocks is that there is no dependencies, but with sqlite being quite portable I think it can do the job the mocks were doing.

This can be closed, I'll make an issue for removing the mocks.

PR #310, opened 2021-09-07
[No issue] Fix sqlite tests

clemens-msupply:
- Use bundled sqlite lib
- Fix feature name

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ async-trait = "0.1.16"
 diesel = { version = "1.4.7", default-features = false, features = ["sqlite", "postgres", "r2d2", "numeric"] }
 diesel-derive-enum = { version = "1.1.1", features = ["sqlite", "postgres"] }
 diesel_migrations = "1.4.0"
+libsqlite3-sys = { version = "0.22.2", features = ["bundled"] }
 r2d2 = "0.8.9"
 config = "0.11.0"
 env_logger = "0.8.3"
--- a/src/util/settings.rs
+++ b/src/util/settings.rs
@@ -62,7 +62,7 @@ impl DatabaseSettings {
 #[cfg(feature = "sqlite")]
 impl DatabaseSettings {
     pub fn connection_string(&self) -> String {
-        format!("{}.db", self.database_name.to_string())
+        format!("{}.sqlite", self.database_name.to_string())
     }
 
     pub fn connection_string_without_db(&self) -> String {
--- a/src/util/settings.rs
+++ b/src/util/settings.rs
@@ -42,7 +42,7 @@ pub struct DatabaseSettings {
     pub database_name: String,
 }
 
-#[cfg(not(feature = "dieselsqlite"))]
+#[cfg(not(feature = "sqlite"))]
 impl DatabaseSettings {
     pub fn connection_string(&self) -> String {
         format!(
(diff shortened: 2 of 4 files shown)


wlthomson:
Issue?

Cargo.toml
```diff
@@ -22,6 +22,7 @@ async-trait = "0.1.16"
 diesel = { version = "1.4.7", default-features = false, features = ["sqlite", "postgres", "r2d2", "numeric"] }
 diesel-derive-enum = { version = "1.1.1", features = ["sqlite", "postgres"] }
 diesel_migrations = "1.4.0"
+libsqlite3-sys = { version = "0.22.2", features = ["bundled"] }
```
wlthomson, line 25: Ohhh, fancy. What does this do?

src/util/settings.rs
```diff
@@ -59,10 +59,10 @@ impl DatabaseSettings {
     }
 }
 
-#[cfg(feature = "dieselsqlite")]
+#[cfg(feature = "sqlite")]
 impl DatabaseSettings {
     pub fn connection_string(&self) -> String {
-        format!("{}.sqlite", self.database_name.to_string())
+        format!("{}.db", self.database_name.to_string())
```
wlthomson, line 65: Oops, the `.db` extension was just an arbitrary choice (I didn't see this connection string when updating the tests). I guess `.sqlite` is a bit clearer, but we can stick with `.db` now if it's easier 😄 .

wlthomson (approved):
LGTM! 

Don't worry this time, but would be helpful to have an issue for each PR (even small ones). I recently read "The Pragmatic Programmer", which has an interesting discussion of the "broken window theory" and evidence that making exceptions to practices in the smaller instances is correllated with the practice not being applied in the bigger instances. Even if you don't agree on that, it's super helpful for repo admin to be able to identify the issue associated with each brach!

src/util/settings.rs
clemens-msupply: don't mind, will switch back to sqlite the think its clearer as well...

Cargo.toml
clemens-msupply: I don't have a sqlite lib installed locally and the linker fails when building the crate. This bundles sqlite into the binary.

This should be required for CI as well...

src/util/settings.rs
chris-morgan: The `.to_string()` here is unnecessary.

clemens-msupply: sorry missed it will fix it later

Cargo.toml
chris-morgan: By default, libsqlite3-sys requires that you have libsqlite3 installed locally, and builds against that, which allows the system to customise SQLite configuration, which is often considered a good thing. But the crate also ships the SQLite source code, and the `bundled` feature says to compile and use that, rather than whatever may be present on the system. This is good for reproducible builds and makes life easier because now you don’t need to have it installed locally.

wlthomson: Ah right, I'm assuming this will have an effect on the size of the final binary?

chris-morgan: Correct; it builds to libsqlite3.a, which I think will be statically linked in (depending on your Cargo/Rust linking configuration—you could make it so you got a libsqlite3.so adjacent to it instead), and now you don’t depend on /usr/lib/libsqlite3.so or equivalent.

PR #311, opened 2021-09-07
#309 Regenerate diesel schema

### LINKED ISSUE #309: Regenerate diesel schema

wlthomson:
Closes #309.

### DIFF
--- a/migrations/postgres/00000102_create_user_account/up.sql
+++ b/migrations/postgres/00000102_create_user_account/up.sql
@@ -3,9 +3,9 @@
 CREATE EXTENSION pgcrypto;
 
 CREATE TABLE user_account (
-    id VARCHAR(255) NOT NULL PRIMARY KEY,
-    username VARCHAR(255) NOT NULL,
+    id TEXT NOT NULL PRIMARY KEY,
+    username TEXT NOT NULL,
     -- Password is stored as salted MD5 hash.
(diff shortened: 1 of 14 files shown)


migrations/postgres/00000102_create_user_account/up.sql
```diff
@@ -3,9 +3,9 @@
 CREATE EXTENSION pgcrypto;
 
 CREATE TABLE user_account (
-    id VARCHAR(255) NOT NULL PRIMARY KEY,
-    username VARCHAR(255) NOT NULL,
+    id TEXT NOT NULL PRIMARY KEY,
```
clemens-msupply, line 6: Interesting do you know more why it doesn't like VARCHAR?

clemens-msupply (approved):

migrations/postgres/00000102_create_user_account/up.sql
wlthomson: I think `VARCHAR` is probably fine, but `String` can store arbitrary length data (at least AFAIK), so it seemed wrong to map that to a fixed-length `VARCHAR`. It also brings the sqlite and pg schemas a bit closer, at least syntactically.

Issue #279, opened 2021-08-31
[SYNC] - Pull Central


andreievg:
MW parent job: `[OMSBES] omSupply Back End - SYNC`

### Description

I think this is mainly implemented, just need to store in intermediary central sync queue

### Table

Need to store central cursor

### Scope

<ins>Includes</ins>
* Able to request and pull X number of records of central data
* Able to store those records
* Able to increment cursor queue
* Able to result in "all done" (all records were pulled)

<ins>Excludes</ins>
* Validation of incoming sync record shape/type

### Tasks

- [ ] Break down tasks into digestible (by reviewer) units.
- [ ] MW job added

wlthomson:
I'm gonna make a start on this by implementing the HTTP logic for pulling from the central queue (currently only pulling from remote is implemented).

wlthomson:
Next step: add logic for saving successfully pulled records into central sync buffer (#331).

wlthomson:
Next step: add table and logic for persisting sync cursor (#336).

wlthomson:
Closing as all tasks completed!

Issue #278, opened 2021-08-31
[SYNC] - Central queue


andreievg:
MW parent job: `[OMSBES] omSupply Back End - SYNC`

### Description

Intermediary sync queue, to store incoming sync record before they are integrated.

### Table

Common columns to be fields (i.e. record_id, record_type), payload to be a text (not JSON), since it will be de-serialised with serde for simple translations

### Scope

<ins>Includes</ins>
* Add table
* Able to insert many record
* Able to get one records
* Able to clear queue

<ins>Excludes</ins>
* GraphQL mutation for getting/setting settings

### Tasks

- [ ] Break down tasks into digestible (by reviewer) units.
- [ ] MW job added

wlthomson:
I think I may have duplicated this with #296. Will delete that and rename the branch.

PR #314, opened 2021-09-08
281 simple translation 2

### LINKED ISSUE #281: [SYNC] - Simple Translation Full - M1

clemens-msupply:
This aims to implement the translaions as we discusses last time.

There are two things that I am not entirely happy with but couldn't fix to this point (because of Rust type and async issues).
1) After another discussion with Will we realized that Andrei and me probably planted to use something more closer to a DAO pattern and that DAO and repository pattern are complimentary (e.g. https://www.baeldung.com/java-dao-vs-repository). For this reason I did some experimenting using a DAO layer between ORM and Repository where the Repository layer is mostly independent from Diesel and only uses the DAO layer. However, I run into some further async issues and gave up on it. However, I think it might be an options to move the auxilary methods like `pub fn insert_one_tx(connection: &DBConnection, name_row: &NameRow)` out of the Repositories into something like `NameQueries` to keep the Repositories more tidy.

2) Upsert for sqlite is only supported in the upcoming Diesel 2.0 version. I wasn't manages to implement the upsert in a generic way, i.e. upsert for sqlite must be implemented multiple times and doing the raw queries is a bit error prone (easy to forget to add new table columns).

Closes #281

### DIFF
--- a/migrations/postgres/00000105_create_item_table/up.sql
+++ b/migrations/postgres/00000105_create_item_table/up.sql
@@ -1,6 +1,6 @@
 -- Create item table.
 
-CREATE TYPE item_type AS ENUM ('general', 'service', 'cross_reference');
+CREATE TYPE item_type AS ENUM ('general', 'service', 'cross_reference', 'none_stock');
 
 CREATE TABLE item (
     id VARCHAR(255) NOT NULL PRIMARY KEY,
--- a/migrations/postgres/00000105_create_item_table/up.sql
+++ b/migrations/postgres/00000105_create_item_table/up.sql
@@ -1,9 +1,6 @@
 -- Create item table.
 
-CREATE TYPE item_type AS ENUM ('general', 'service', 'cross_reference', 'none_stock');
-
 CREATE TABLE item (
     id VARCHAR(255) NOT NULL PRIMARY KEY,
-    item_name VARCHAR(255) NOT NULL,
-    type_of item_type NOT NULL
+    name VARCHAR(255) NOT NULL
 )
--- a/src/database/repository/diesel/item.rs
+++ b/src/database/repository/diesel/item.rs
@@ -1,4 +1,4 @@
-use super::DBBackendConnection;
+use super::{DBBackendConnection, DBConnection};
 
 use crate::database::{
     repository::{repository::get_connection, RepositoryError},
@@ -20,12 +20,56 @@ impl ItemRepository {
         ItemRepository { pool }
     }
 
-    pub async fn insert_one(&self, item_row: &ItemRow) -> Result<(), RepositoryError> {
+    #[cfg(feature = "postgres")]
+    pub fn upsert_one_tx(
+        connection: &DBConnection,
+        item_row: &ItemRow,
+    ) -> Result<(), RepositoryError> {
         use crate::database::schema::diesel_schema::item::dsl::*;
-        let connection = get_connection(&self.pool)?;
+
         diesel::insert_into(item)
             .values(item_row)
-            .execute(&connection)?;
+            .on_conflict(id)
+            .do_update()
+            .set(item_row)
+            .execute(connection)?;
+        Ok(())
+    }
+
+    #[cfg(feature = "sqlite")]
+    pub fn upsert_one_tx(
+        connection: &DBConnection,
+        item_row: &ItemRow,
+    ) -> Result<(), RepositoryError> {
+        use diesel::sql_types::Text;
+
+        let query = r#"
+        INSERT INTO item(id,item_name,type_of) VALUES($1, $2, $3)
+        ON CONFLICT(id) DO UPDATE SET
+            item_name=excluded.item_name,
+            type_of=excluded.type_of;"#;
+        let q = diesel::sql_query(query)
+            .bind::<Text, _>(&item_row.id)
+            .bind::<Text, _>(&item_row.item_name)
+            .bind::<crate::database::schema::item::ItemRowTypeMapping, _>(&item_row.type_of);
+        q.execute(connection)?;
+        Ok(())
+    }
+
+    pub fn insert_one_tx(
+        connection: &DBConnection,
+        item_row: &ItemRow,
+    ) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::item::dsl::*;
+        diesel::insert_into(item)
+            .values(item_row)
+            .execute(connection)?;
+        Ok(())
+    }
+
+    pub async fn insert_one(&self, item_row: &ItemRow) -> Result<(), RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+        ItemRepository::insert_one_tx(&connection, item_row)?;
         Ok(())
     }
 
--- a/src/database/repository/diesel/item.rs
+++ b/src/database/repository/diesel/item.rs
@@ -41,18 +41,10 @@ impl ItemRepository {
         connection: &DBConnection,
         item_row: &ItemRow,
     ) -> Result<(), RepositoryError> {
-        use diesel::sql_types::Text;
-
-        let query = r#"
-        INSERT INTO item(id,item_name,type_of) VALUES($1, $2, $3)
-        ON CONFLICT(id) DO UPDATE SET
-            item_name=excluded.item_name,
-            type_of=excluded.type_of;"#;
-        let q = diesel::sql_query(query)
-            .bind::<Text, _>(&item_row.id)
-            .bind::<Text, _>(&item_row.item_name)
-            .bind::<crate::database::schema::item::ItemRowTypeMapping, _>(&item_row.type_of);
-        q.execute(connection)?;
+        use crate::database::schema::diesel_schema::item::dsl::*;
+        diesel::replace_into(item)
+            .values(item_row)
+            .execute(connection)?;
         Ok(())
     }
 
@@ -69,8 +61,7 @@ impl ItemRepository {
 
     pub async fn insert_one(&self, item_row: &ItemRow) -> Result<(), RepositoryError> {
         let connection = get_connection(&self.pool)?;
-        ItemRepository::insert_one_tx(&connection, item_row)?;
-        Ok(())
+        ItemRepository::insert_one_tx(&connection, item_row)
     }
 
     pub async fn find_all(&self) -> Result<Vec<ItemRow>, RepositoryError> {
--- a/src/database/repository/diesel/mod.rs
+++ b/src/database/repository/diesel/mod.rs
@@ -8,6 +8,7 @@ mod name;
 mod requisition;
 mod requisition_line;
 mod store;
+mod sync;
 mod transact;
 mod transact_line;
 mod user_account;
@@ -18,6 +19,7 @@ pub use name::NameRepository;
 pub use requisition::RequisitionRepository;
 pub use requisition_line::RequisitionLineRepository;
 pub use store::StoreRepository;
+pub use sync::{IntegrationRecord, IntegrationUpsertRecord, SyncRepository};
 pub use transact::{CustomerInvoiceRepository, TransactRepository};
 pub use transact_line::TransactLineRepository;
 pub use user_account::UserAccountRepository;
@@ -31,8 +33,10 @@ use diesel::{
 #[cfg(feature = "sqlite")]
 type DBBackendConnection = SqliteConnection;
 
-#[cfg(feature = "postgres")]
+#[cfg(not(feature = "sqlite"))]
 type DBBackendConnection = PgConnection;
+
+pub type DBConnection = PooledConnection<ConnectionManager<DBBackendConnection>>;
 
 impl From<DieselError> for RepositoryError {
     fn from(err: DieselError) -> Self {
@@ -91,6 +95,7 @@ pub async fn get_repositories(settings: &Settings) -> RepositoryMap {
     repositories.insert(TransactRepository::new(pool.clone()));
     repositories.insert(TransactLineRepository::new(pool.clone()));
     repositories.insert(UserAccountRepository::new(pool.clone()));
+    repositories.insert(SyncRepository::new(pool.clone()));
 
     repositories
 }
--- a/src/database/repository/diesel/name.rs
+++ b/src/database/repository/diesel/name.rs
@@ -1,4 +1,4 @@
-use super::DBBackendConnection;
+use super::{DBBackendConnection, DBConnection};
 
 use crate::database::{
     repository::{repository::get_connection, RepositoryError},
@@ -20,26 +20,82 @@ impl NameRepository {
         NameRepository { pool }
     }
 
-    pub async fn insert_one(&self, name_row: &NameRow) -> Result<(), RepositoryError> {
+    pub fn insert_one_tx(
+        connection: &DBConnection,
+        name_row: &NameRow,
+    ) -> Result<(), RepositoryError> {
         use crate::database::schema::diesel_schema::name_table::dsl::*;
-        let connection = get_connection(&self.pool)?;
         diesel::insert_into(name_table)
             .values(name_row)
-            .execute(&connection)?;
+            .execute(connection)?;
+        Ok(())
+    }
+
+    #[cfg(feature = "postgres")]
+    pub fn upsert_one_tx(
+        connection: &DBConnection,
+        name_row: &NameRow,
+    ) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::name_table::dsl::*;
+
+        diesel::insert_into(name_table)
+            .values(name_row)
+            .on_conflict(id)
+            .do_update()
+            .set(name_row)
+            .execute(connection)?;
+        Ok(())
+    }
+
+    #[cfg(feature = "sqlite")]
+    pub fn upsert_one_tx(
+        connection: &DBConnection,
+        name_row: &NameRow,
+    ) -> Result<(), RepositoryError> {
+        use diesel::sql_types::Text;
+
+        let query = r#"
+        INSERT INTO name(id,name) VALUES($1, $2)
+        ON CONFLICT(id) DO UPDATE SET
+            name=excluded.name;"#;
+        let q = diesel::sql_query(query)
+            .bind::<Text, _>(&name_row.id)
+            .bind::<Text, _>(&name_row.name);
+        q.execute(connection)?;
+        Ok(())
+    }
+
+    pub fn find_one_by_id_tx(
+        connection: &DBConnection,
+        name_id: &str,
+    ) -> Result<NameRow, RepositoryError> {
+        use crate::database::schema::diesel_schema::name_table::dsl::*;
+        let result = name_table.filter(id.eq(name_id)).first(connection)?;
+        Ok(result)
+    }
+
+    pub fn find_many_by_id_tx<'a>(
+        connection: &'a DBConnection,
+        ids: &[String],
+    ) -> Result<Vec<NameRow>, RepositoryError> {
+        use crate::database::schema::diesel_schema::name_table::dsl::*;
+        let result = name_table.filter(id.eq_any(ids)).load(connection)?;
+        Ok(result)
+    }
+
+    pub async fn insert_one(&self, name_row: &NameRow) -> Result<(), RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+        NameRepository::insert_one_tx(&connection, name_row)?;
         Ok(())
     }
 
     pub async fn find_one_by_id(&self, name_id: &str) -> Result<NameRow, RepositoryError> {
-        use crate::database::schema::diesel_schema::name_table::dsl::*;
         let connection = get_connection(&self.pool)?;
-        let result = name_table.filter(id.eq(name_id)).first(&connection)?;
-        Ok(result)
+        NameRepository::find_one_by_id_tx(&connection, name_id)
     }
 
     pub async fn find_many_by_id(&self, ids: &[String]) -> Result<Vec<NameRow>, RepositoryError> {
-        use crate::database::schema::diesel_schema::name_table::dsl::*;
         let connection = get_connection(&self.pool)?;
-        let result = name_table.filter(id.eq_any(ids)).load(&connection)?;
-        Ok(result)
+        NameRepository::find_many_by_id_tx(&connection, ids)
     }
 }
--- a/src/database/repository/diesel/name.rs
+++ b/src/database/repository/diesel/name.rs
@@ -52,16 +52,10 @@ impl NameRepository {
         connection: &DBConnection,
         name_row: &NameRow,
     ) -> Result<(), RepositoryError> {
-        use diesel::sql_types::Text;
-
-        let query = r#"
-        INSERT INTO name(id,name) VALUES($1, $2)
-        ON CONFLICT(id) DO UPDATE SET
-            name=excluded.name;"#;
-        let q = diesel::sql_query(query)
-            .bind::<Text, _>(&name_row.id)
-            .bind::<Text, _>(&name_row.name);
-        q.execute(connection)?;
+        use crate::database::schema::diesel_schema::name_table::dsl::*;
+        diesel::replace_into(name_table)
+            .values(name_row)
+            .execute(connection)?;
         Ok(())
     }
 
--- a/src/database/repository/diesel/sync.rs
+++ b/src/database/repository/diesel/sync.rs
@@ -1,3 +1,5 @@
+use super::{get_connection, DBBackendConnection, ItemRepository, NameRepository};
+
 use crate::database::{
     repository::RepositoryError,
     schema::{ItemRow, NameRow},
@@ -7,8 +9,6 @@ use diesel::{
     prelude::*,
     r2d2::{ConnectionManager, Pool},
 };
-
-use super::{get_connection, DBBackendConnection, ItemRepository, NameRepository};
 
 pub enum IntegrationUpsertRecord {
     Name(NameRow),
--- a/src/database/repository/diesel/sync.rs
+++ b/src/database/repository/diesel/sync.rs
@@ -1,0 +1,51 @@
+use crate::database::{
+    repository::RepositoryError,
+    schema::{ItemRow, NameRow},
+};
+
+use diesel::{
+    prelude::*,
+    r2d2::{ConnectionManager, Pool},
+};
+
+use super::{get_connection, DBBackendConnection, ItemRepository, NameRepository};
+
+pub enum IntegrationUpsertRecord {
+    Name(NameRow),
+    Item(ItemRow),
+}
+
+pub struct IntegrationRecord {
+    pub upserts: Vec<IntegrationUpsertRecord>,
+}
+
+#[derive(Clone)]
+pub struct SyncRepository {
+    pool: Pool<ConnectionManager<DBBackendConnection>>,
+}
+
+impl SyncRepository {
+    pub fn new(pool: Pool<ConnectionManager<DBBackendConnection>>) -> SyncRepository {
+        SyncRepository { pool }
+    }
+
+    pub async fn integrate_records(
+        &self,
+        integration_records: &IntegrationRecord,
+    ) -> Result<(), RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+        connection.transaction(|| {
+            for record in &integration_records.upserts {
+                match &record {
+                    IntegrationUpsertRecord::Name(record) => {
+                        NameRepository::upsert_one_tx(&connection, record)?
+                    }
+                    IntegrationUpsertRecord::Item(record) => {
+                        ItemRepository::upsert_one_tx(&connection, record)?
+                    }
+                }
+            }
+            Ok(())
+        })
+    }
+}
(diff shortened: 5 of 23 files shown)


wlthomson:
@clemens-msupply thanks again for that article, it articulates what I was trying to say about repos != DAOs better than I was able to. The key summary I think I was going for was:

> In other words, a repository also deals with data and hides queries similar to DAO. However, it sits at a higher level, closer to the business logic of an app.

**Disclaimer: what comes below is just to help clarify why I suggested the approach I did, if others still disagree, I'm happy to move forward with a more hybrid "DAO-repo" interpretation where the repo is kept segregated from any domain layer business logic.**

This is the justification behind my preference for the sync repo methods to take domain model structs rather than already mapped ones. I think it would be really nice if the sync actors only have to exist at the domain model level and just worry about sending and receiveing sync records as they are structured in the sync API (ahd leave all the mapping between the domain objects and database objects to the repos). Hopefully it also helps explain what I meant by saying that ideally transactions would be kept at the repo level? I think one way to conceptualise transactions is as part of the mapping between the domain model and the database model, in that they are mapping a single atomic domain level operation, e.g. integration a list of records, into multiple database level operations on multiple database entities.

Anyways, will get on with a proper review now. Just wanted to acknowledge my appreciation for the searching for that article and having a go at incorporating the approach. I feel a lot better moving forward now that we've considered it even if we end up taking a different approach.

src/database/repository/diesel/item.rs
```diff
@@ -20,12 +20,56 @@ impl ItemRepository {
...
+        item_row: &ItemRow,
+    ) -> Result<(), RepositoryError> {
+        use diesel::sql_types::Text;
+
+        let query = r#"
+        INSERT INTO item(id,item_name,type_of) VALUES($1, $2, $3)
+        ON CONFLICT(id) DO UPDATE SET
+            item_name=excluded.item_name,
+            type_of=excluded.type_of;"#;
+        let q = diesel::sql_query(query)
```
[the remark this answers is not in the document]
andreievg: I kind of thought the point of diesel was to maximise the static typing.. 
At this stage I would rather query for existence of ID and do insert or update (generic code for both), vs using raw sql.

src/database/repository/diesel/item.rs
```diff
@@ -20,12 +20,56 @@ impl ItemRepository {
...
         diesel::insert_into(item)
             .values(item_row)
-            .execute(&connection)?;
+            .execute(connection)?;
+        Ok(())
+    }
+
+    pub async fn insert_one(&self, item_row: &ItemRow) -> Result<(), RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+        ItemRepository::insert_one_tx(&connection, item_row)?;
```
andreievg, line 72: insert_one_tx has the same return type and insert_one

src/database/repository/diesel/mod.rs
```diff
@@ -31,9 +33,11 @@ use diesel::{
 #[cfg(feature = "sqlite")]
 type DBBackendConnection = SqliteConnection;
 
-#[cfg(feature = "postgres")]
+#[cfg(not(feature = "sqlite"))]
 type DBBackendConnection = PgConnection;
 
+pub type DBConnection = PooledConnection<ConnectionManager<DBBackendConnection>>;
```
andreievg, line 39: nice

src/database/repository/diesel/item.rs
```diff
@@ -20,12 +20,56 @@ impl ItemRepository {
         ItemRepository { pool }
     }
 
-    pub async fn insert_one(&self, item_row: &ItemRow) -> Result<(), RepositoryError> {
+    #[cfg(feature = "postgres")]
+    pub fn upsert_one_tx(
```
andreievg, line 24: i feel like `_tx` implies transaction too much. Wonder what the practice in in rust of utility methods, from my understanding it's underscore, i.e. `_upsert_one` (at least form the code i've been reading)

src/database/repository/diesel/name.rs
```diff
@@ -20,26 +20,82 @@ impl NameRepository {
...
+        ON CONFLICT(id) DO UPDATE SET
+            name=excluded.name;"#;
+        let q = diesel::sql_query(query)
+            .bind::<Text, _>(&name_row.id)
+            .bind::<Text, _>(&name_row.name);
+        q.execute(connection)?;
+        Ok(())
+    }
+
+    pub fn find_one_by_id_tx(
```
andreievg, line 68: do we really need to be extracting all methods to associated methods like this ? Would be good to touch just the bits that are in scope for an issue

src/database/repository/diesel/sync.rs
```diff
@@ -0,0 +1,51 @@
...
+pub enum IntegrationUpsertRecord {
+    Name(NameRow),
+    Item(ItemRow),
+}
+
+pub struct IntegrationRecord {
+    pub upserts: Vec<IntegrationUpsertRecord>,
+}
+
+#[derive(Clone)]
```
andreievg, line 22: ohh, just realised, we actually clone every repository with `registry.get::`. I guess derive would call `pool.clone()`. I wonder if we need to optimise that at some stage (just noting)

migrations/postgres/00000105_create_item_table/up.sql
```diff
@@ -1,6 +1,6 @@
 -- Create item table.
 
-CREATE TYPE item_type AS ENUM ('general', 'service', 'cross_reference');
+CREATE TYPE item_type AS ENUM ('general', 'service', 'cross_reference', 'none_stock');
```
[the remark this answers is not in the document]
andreievg: I think we should remove it form schema, so we know that tables we touch are 'clean' fully translated up to the current spec

src/util/sync/translation/item.rs
```diff
@@ -0,0 +1,167 @@
...
+    */
+}
+
+fn match_item_type(type_of: String) -> ItemRowType {
+    match type_of.as_str() {
+        "general" => ItemRowType::General,
+        "service" => ItemRowType::Service,
+        "cross_reference" => ItemRowType::CrossReference,
+        "non_stock" => ItemRowType::NoneStock,
+        _ => panic!("unknown type"),
```
andreievg, line 93: relax : ), even though this will be refactored, please no panics

src/util/sync/translation/item.rs
```diff
@@ -0,0 +1,167 @@
...
+    universalcodes_name: String,
+    //kit_data": null,
+    //custom_data": null,
+    doses: i64,
+    is_vaccine: bool,
+    restricted_location_type_ID: String,
+    */
+}
+
+fn match_item_type(type_of: String) -> ItemRowType {
```
andreievg, line 87: although i think this needs to be removed, but fyi, you can give a 'default' value if enums is not matched:
```rust
#[derive(Deserialize, Serialize, Debug)]
enum Check {
    A,
    B,
    #[serde(other)]
    Unknown,
}
```
if value being deserialised is neither A or B, Unknown will be used

src/util/sync/translation/item.rs
```diff
@@ -0,0 +1,167 @@
...
+            schema::ItemRowType,
+        },
+        server::data::RepositoryRegistry,
+        util::{
+            sync::translation::{import_sync_records, SyncRecord, SyncType},
+            test_db,
+        },
+    };
+
+    #[tokio::test]
```
andreievg, line 126: is tokio the runtime used by actix_web ? wonder if we should be testing in same runtime or if it doesn't matter

src/util/sync/translation/item.rs
```diff
@@ -0,0 +1,167 @@
...
+    async fn test_item_translation() {
```
andreievg, line 127: I feel like we should be testing methods in scope, what do you think ? (otherwise it's a small integration test, and they live in `tests` folders apperantly)

src/database/repository/diesel/sync.rs
```diff
@@ -0,0 +1,51 @@
...
+pub struct SyncRepository {
+    pool: Pool<ConnectionManager<DBBackendConnection>>,
+}
+
+impl SyncRepository {
+    pub fn new(pool: Pool<ConnectionManager<DBBackendConnection>>) -> SyncRepository {
+        SyncRepository { pool }
+    }
+
+    pub async fn integrate_records(
```
andreievg, line 32: any reason to have this async ?

src/util/sync/translation/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+    data: String,
+}
+
+/// Translates sync records into the local DB schema.
+/// Translated records are added to integration_records.
+fn do_translation(
+    sync_record: &SyncRecord,
+    integration_records: &mut IntegrationRecord,
+) -> Result<(), String> {
+    if let Some(row) = LegacyNameTable::try_translate(sync_record)? {
```
andreievg, line 33: What do you think should try_translate return `Option<IntegrationRecord>` ?
Also (and we cross that road when we get there), we might need to return a Vec of IntegrationRecord (for complex translations, but that could be a seperate method)

andreievg (changes requested):
Thanks Clemens!

Very easy to read and follow.
I would typically recommend PR's under 200 lines, but I've been guilty of submitting large one. And code is super easy to follow so I guess it's ok.

A few changes and questions

can you please do cargo check --features sqlite (and postgres) and get rid of warning.

> Upsert for sqlite is only supported in the upcoming Diesel 2.0 version. I wasn't manages to implement the upsert in a generic way, i.e. upsert for sqlite must be implemented multiple times and doing the raw queries is a bit error prone (easy to forget to add new table columns).

I've left a note about this, i think we should keep static as much as possible (since keeping static was one of the main reasons for diesel)

well written but break it down

andreievg:
> After another discussion with Will we realized that Andrei and me probably planted to use something more closer to a DAO pattern and that DAO and repository pattern are complimentary (e.g. https://www.baeldung.com/java-dao-vs-repository). For this reason I did some experimenting using a DAO layer between ORM and Repository where the Repository layer is mostly independent from Diesel and only uses the DAO layer

I am glad you didn't, because I would have disagreed, two layers of abstraction ORM, and DOA/Repo type is plenty for what we are trying to build.

Also can you please include me in these sorts of discussions. 

@wlthomson I didn't read your post, since we already made a decision. Didn't want to be distracted (wanted to actually focus on getting some work done)

migrations/postgres/00000105_create_item_table/up.sql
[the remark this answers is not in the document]
clemens-msupply: This was more to make the test work with "real" data. Agree, having the enum makes the schema to restricted. For example, if a newer server version has a new enum member the old server can't sync the new data anymore even though it doesn't need the new enum.

src/database/repository/diesel/item.rs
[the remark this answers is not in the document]
clemens-msupply: Yes good point I was too focused on the upsert syntax. Will do it with INSERT OR REPLACE till diesel supports upsert.

migrations/postgres/00000105_create_item_table/up.sql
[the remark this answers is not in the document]
andreievg: > For example, if a newer server version has a new enum member the old server can't sync the new data anymore even though it doesn't need the new enum.

I think it's fine with enums, see: https://github.com/openmsupply/remote-server/pull/314#discussion_r704332225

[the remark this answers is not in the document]
andreievg: My point is that the schema that we have is just for prototype reasons, we should be using schema in translated document from now on (to be certain that only agreed on schema is in, and so we don't need to reconsolidate it)

src/database/repository/diesel/item.rs
clemens-msupply: No idea about utility methods, up to you...

Think tx is quite fitting though because that is exactly what the method is for.

andreievg: good point, not fussy, i just saw it used in async method without transaction.
backing down on that request. if we end up using it more frequently we can change

src/database/repository/diesel/sync.rs
clemens-msupply: It's only async because other Repositories are having an async interface which is to keep the rest of the system independent from the DB backend / ORM...

src/util/sync/translation/item.rs
clemens-msupply: I assumed so but I couldn't really confirm it and I am probably wrong. I will use #[actix_rt::test] instead.

src/util/sync/translation/mod.rs
clemens-msupply: The IntegrationRecord was indented to describe the whole job that needs to be done by the SyncRepository (only upserts are currently supported).

I used `try_translate` more like a convention rather than an interface. A complex translation could return a tuple which is then filled into the IntegrationRecord depending on the translation... However, open for changes and probably best to decide when we have an actual complex translation.

src/util/sync/translation/item.rs
clemens-msupply: It tricky, depends on how you define unit tests. Here I am testing the unit of "legacy record integration" which has two parts the data translations and the SyncRepository. 

Personally find it easier to have the test close to the main logic but I am fine to move it to integration tests.

For me an integration test would be more like an API call to the running server, but there are multiple schools about it...

clemens-msupply: btw. I plan to move the existing repository tests into smaller unit tests, think that would fit better...

clemens-msupply:
> can you please do cargo check --features sqlite (and postgres) and get rid of warning.

these warning should get fixed once the translation is actually used... Find to ignore these warnings till then?

> Also can you please include me in these sorts of discussions.

sorry about that, it was just a follow up on a chat we had earlier...

src/util/sync/translation/item.rs
andreievg: > Personally find it easier to have the test close to the main logic

Wouldn't that mean you would test `try_translate` here (without import_sync_record), and `import_sync_records` in `translation/mod.rs` ? 

Btw that's what i meant by testing methods in scope

clemens-msupply: You would basically have the same test case twice, wouldn't you? for this reason I find it easier to do it in one pass...

andreievg:
Thanks for comments so far, i've unresolved some comments that were marked as resolved (will resolve them with changes), btw I found this process the easier for reviewer:
* Reviewer resolves comments they've made
* Code changes hash is pasted in comment blocks it addresses

andreievg:
> these warning should get fixed once the translation is actually used... Fine to ignore these warnings till then?

ohh true, some provisionary things will be used upstream

> sorry about that, it was just a follow up on a chat we had earlier...

no problems, sorry i feel like my comment didn't sound like i intended. (it was a soft request of being included in the loop)

src/util/sync/translation/item.rs
andreievg: It would point to failing area better, ok to just have one case, but it shouldn't be here (if it's test like this it should probably be in `import_sync_record`. 

I was suggesting:

try_translates checks for deserialisation in this case, assert comparison with result object (don't do any db inserts)

import_sync_record checks that it can integrate multiple records, assert db query result at the end.  (and yes it would double up here, but you would have a much better idea of the origin of the issue ? and when you deving, making further translation for item, you can just run the deserialisation test by it's own, you would have seen the 'play' button rust analyser put on each test )

src/util/sync/translation/item.rs
```diff
@@ -0,0 +1,167 @@
...
+    VEN_category: String,
+    universalcodes_code: String,
+    universalcodes_name: String,
+    //kit_data": null,
+    //custom_data": null,
+    doses: i64,
+    is_vaccine: bool,
+    restricted_location_type_ID: String,
+    */
+}
```
wlthomson, line 85: I understand why this is all here, but we could just go yagni and add the fields when they are needed? We will have this commit in the history, so can always grab these fields from there if the idea is to save time later!

src/database/repository/diesel/sync.rs
```diff
@@ -0,0 +1,51 @@
...
+        &self,
+        integration_records: &IntegrationRecord,
+    ) -> Result<(), RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+        connection.transaction(|| {
+            for record in &integration_records.upserts {
+                match &record {
+                    IntegrationUpsertRecord::Name(record) => {
+                        NameRepository::upsert_one_tx(&connection, record)?
```
wlthomson, line 41: Is the idea that this would be replaced with buffer inserts? Or would integration records be taken from the buffer after a full sync had occurred?

wlthomson (changes requested):
Mostly LGTM!

Have realised that a lot of the buffer work I've done will have to be changed to accomodate this. Would be good to get some agreement on how we think that will integrate!

src/database/repository/diesel/sync.rs
wlthomson: Not sure if this was accidentally resolved but I think this is an important question to get an answer to as I'm quite confused by the pattern we are using and how we expect buffering to work with it.

src/util/sync/translation/item.rs
clemens-msupply: especially if you know recent commits it should be easy to find the origin (if the test ever breaks). 

I thought about it (probably too long now ;-) ) but I only see theoretical benefits for splitting the tests. I see some disadvantages though:
- more annoying to write (quite a strong point for me, e.g. why spend time on a "boring" repetitive task while it can be avoided)
- as said, to get equivalent sorrow tests you need to feed the same data to both, the translation and RepositorySync, this will be quite error prone

Your call...

src/util/sync/translation/item.rs
```diff
@@ -0,0 +1,167 @@
...
+    match type_of.as_str() {
+        "general" => ItemRowType::General,
+        "service" => ItemRowType::Service,
+        "cross_reference" => ItemRowType::CrossReference,
+        "non_stock" => ItemRowType::NoneStock,
+        _ => panic!("unknown type"),
+    }
+}
+
+impl LegacyItemRow {
```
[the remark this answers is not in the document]
clemens-msupply: Agree, I only did it because there was an enum in there before for which reason I couldn't use From. I already use from for Name.

See discussion above for test location...

src/database/repository/diesel/sync.rs
clemens-msupply: Yes would have to have a look how that best integrates with your work once we are putting it together...

src/util/sync/translation/item.rs
wlthomson: > You would basically have the same test case twice, wouldn't you? for this reason I find it easier to do it in one pass...

My understanding is that a unit test just focuses on one method or function and are don't need to be aware of the greater context except perhaps in providing a good coverage of inputs, i.e. a unit test is isolated and stateless (at least as much as possible), it passes if a given input returns the expected output. In comparison, integration tests are used to test that multiple methods (which are already tested in isolation to ensure they work individully) also behave as expected when used together. Typically unit tests are great in number so that they test as many function "branches" as possible, whereas integration tests are more sparse and designed for additional reassurance that the components which we know work as expected in isolation also behave as expected when combined. For that reason, I think it's quite OK for them to overlap in some instances, in fact I think the coverage provided by unit tests is complementary as it allows the integration tests to assume correctness of the functions they are testing in combination.

andreievg: I am happy with just one test. But shouldn't it be in RepositorySync ? A bit weird to me to import the consumer of the scoped method. 
Anyways also happy to keep it as is, and see how it looks further down the road. So not stopping approval of PR

wlthomson:
> 
> 
> > After another discussion with Will we realized that Andrei and me probably planted to use something more closer to a DAO pattern and that DAO and repository pattern are complimentary (e.g. [baeldung.com/java-dao-vs-repository](https://www.baeldung.com/java-dao-vs-repository)). For this reason I did some experimenting using a DAO layer between ORM and Repository where the Repository layer is mostly independent from Diesel and only uses the DAO layer
> 
> I am glad you didn't, because I would have disagreed, two layers of abstraction ORM, and DOA/Repo type is plenty for what we are trying to build.
> 
> Also can you please include me in these sorts of discussions.
> 
> @wlthomson I skimmed through your post, didn't really read it, since we already made a decision. Didn't want to be distracted (wanted to actually focus on getting some work done)
> 
> > Anyways, will get on with a proper review now
> 
> Would recommend that you submit a PR first, I can do the review for this one

There was no discussion, except for Clemens posting that article and asking me if it was representative of what I was trying to say in the meeting (it is). The problem is that this PR includes logic for inserting records into the db directly without the buffer, so there is a lot of overlap which will need resolving. If we prefer to do the PR anyway, I'll do that.

src/database/repository/diesel/sync.rs
andreievg: > we expect buffering to work with it.
@wlthomson have you read this design and diagram https://github.com/openmsupply/org-issues/discussions/32 ?

> And have you had a look at description of the issue you are working on ?

https://github.com/openmsupply/remote-server/issues/278

It just involves saving and writing to buffer (nothing more)

then this issue: 

https://github.com/openmsupply/remote-server/issues/279

will store incoming sync record in the buffer.

then `import_sync_records` would either get an array of records or look them up in db one at a time and integrate them. it's pretty simple, if you having issue conceptualising it i think it's best to just follow the specs in the issue and you'll see how it all fits together when it's implemented.

src/util/sync/translation/item.rs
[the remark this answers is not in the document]
andreievg: There is a `from` there, this translate also checks for name of the table, we might end up moving that though. And `from` doesn't take arguments, likely needed for complex translations, although that might be a seperate method.

> It also gives us conversion back from ItemRow to LegacyItemRow for free

No it doesn't (you have to implement seperate From<LegacyItemRow> for ItemRow and From<ItemRow> for LegacyItemRow)

Screen Shot 2021-09-09 at 5 19 13 PM

src/database/repository/diesel/sync.rs
andreievg: sync records are taken out of buffer, and integrated, pretty simple imo.

src/database/repository/diesel/sync.rs
wlthomson: I've read the issues, but I finding it hard to understand how this PR integrates with that design and how the sync buffer reading and writing is expected to be implemented (hence the relevance of the repo discussion).

I think it's best if I PR what I have so far and we can discuss there, as it might be more productive to get your direct thoughts on the code.

src/util/sync/translation/item.rs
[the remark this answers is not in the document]
wlthomson: Ah, yeah you're right. I misunderstood what was being said here: https://doc.rust-lang.org/rust-by-example/conversion/from_into.html.

src/database/repository/diesel/sync.rs
wlthomson: Is the buffer the sane as what you've called `central_queue` on the diagram?

src/database/repository/diesel/sync.rs
wlthomson: Does it? I thought it returned a reference to the same repository?

From `anymap::Map`:

```
/// Returns a reference to the value stored in the collection for the type `T`, if it exists.
#[inline]
pub fn get<T: IntoBox<A>>(&self) -> Option<&T> {
  self.raw.get(&TypeId::of::<T>())
  .map(|any| unsafe { any.downcast_ref_unchecked::<T>() })
}
```

Although, `AnyRepository` includes the `CloneAny` trait, so maybe you're right? I think the cloning of the pool may be implemented with reference counting under-the-hood (@chris-morgan)?

In any case, I'm wondering if it might be a good idea to take the pool out of the repos and inject it into the methods instead (like has been done for the `_tx` methods here? Also something for later, perhaps.

wlthomson:
> After another discussion with Will we realized that Andrei and me probably planted to use something more closer to a DAO pattern and that DAO and repository pattern are complimentary (e.g. [baeldung.com/java-dao-vs-repository](https://www.baeldung.com/java-dao-vs-repository)). For this reason I did some experimenting using a DAO layer between ORM and Repository where the Repository layer is mostly independent from Diesel and only uses the DAO layer
> 
> I am glad you didn't, because I would have disagreed, two layers of abstraction ORM, and DOA/Repo type is plenty for what we are trying to build.

That's fine, but it's unclear to me what a DAO/Repo is and what logic goes where?

> @wlthomson I skimmed through your post, didn't really read it, since we already made a decision. Didn't want to be distracted (wanted to actually focus on getting some work done)
> 
> Also can you please include me in these sorts of discussions.

I'm trying to understand what that decision was, and I'm asking because it's a blocker on the buffer work.

There was no big discussion, it consisted of Clemens posting that article and asking me if it was representative of what I was trying to say in the meeting and me saying it was. My question is just so I can have that decision in written form so that I know how to move forward with buffer translation logic. I don't really know what the decision was and it's unclear from this PR the scope of `SyncRepository` and whether it will also be used for e.g. reading and writing buffer records.

src/database/repository/diesel/sync.rs
chris-morgan: The use of a cloneable AnyMap seems to already be a historical artefact—it’s not actually doing any cloning; you can change AnyRepository from CloneAny to Any and it works fine.

src/database/repository/diesel/item.rs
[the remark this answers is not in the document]
clemens-msupply: Fixed in f8885b

migrations/postgres/00000105_create_item_table/up.sql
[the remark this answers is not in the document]
clemens-msupply: Removed item type in d8cc52

src/database/repository/diesel/item.rs
clemens-msupply: Slipped into f8885b

src/database/repository/diesel/sync.rs
wlthomson: EDIT: `central_queue` is equivalent to `central_buffer_queue`. Resolves most of my immediate confusion.

src/util/sync/translation/item.rs
wlthomson: Not a big deal, resolving as don't want to add unhelpful blockers.

src/database/repository/diesel/sync.rs
wlthomson: Thanks for the review @andreievg, resolving this now.

src/database/repository/diesel/sync.rs
wlthomson: > 
> 
> The use of a cloneable AnyMap seems to already be a historical artefact—it’s not actually doing any cloning; you can change the definition of AnyRepository from CloneAny to Any and it works fine (and so this should be done).

Cool, thanks. So, to clarify (for my own learning if nothing else), is this unrelated to the cloning @andreievg is referencing, or are you saying that `repository.get` in fact isn't doing any cloning and so there will be no functional difference in removing the trait?

chris-morgan: `registry.get()` doesn’t clone and never did; the reason it used to need CloneAny was because the registry needed to be Clone because we were using a registry per thread (so it cloned each item in the registry at thread creation time only), but now use just one system-wide. See https://github.com/openmsupply/remote-server/pull/219#discussion_r671918062 and 024719867e62aeebaace8cfba44b2df444eb3e93.

wlthomson: Ah, right. I can understand the confusion, I should have removed `CloneAny` in 0247198. Will make an issue.

src/database/repository/diesel/item.rs
wlthomson: Note: you probably already know this, but if you paste a commit hash before pushing the commits, github won't generate a link. When this happens to me, I edit the comment after pushing the relevant commits. The commit link should be generated after doing that 🤞.

andreievg (approved):
Thanks for changes and answers!
Just realised there is still a couple of unresolved comments, but they dont' look that important, up to you if you want to make the changes or not

clemens-msupply:
thanks for the review!

> That's fine, but it's unclear to me what a DAO/Repo is and what logic goes where?

A DAO/Repo is a minimal layer over the DB. This has the advantages:
- all low level db and orm code is in one place and doesn't leak into the rest of the code base
- You can (easier) do cool stuff like: swap DBs, access a DB over the network, request caching, logging... all in a transparent way.
If you put logic in it you water down these advantages and your Repositories just becomes "normal" code.

andreievg:
@wlthomson can you review changes requested please, good to merge soon

wlthomson (approved):

Issue #312, opened 2021-09-07
Remove mock repositories


wlthomson:
See https://github.com/openmsupply/remote-server/pull/303#issuecomment-914080963.

wlthomson:
While working on this, I came across something that reminded me of one of the key reasons that I put the mock repos in: foreign keys.

Using the mock repos, writing a test for the requisition gql query only requires inserting data used by the test (in this example, a requisition). Rewriting the test to use sqlite/pg requires inserting mock records into all the tables which requisition has a FK constraint on, which in turn requires inserting mock records for all the tables those tables have FK constraints on... and so on. As the schema grows, I imagine this could become quite an annoyance and a potential time sink. 

I can see the logic of having our tests as closely aligned to how the app will be used in production, however I don't think it actually really gives us that much here. Ideally, the gql resolvers will be using repo methods which will be being tested separately, and I'd hope we would design those tests to provide coverage solid enough to catch any database related bugs/regressions which would also cause gql tests to fail. This is the reasoning that led me to test the gql API with mock data rather than a live db.

I assume we still want to go farward with this, but I think it's worth flagging up. Since we probably still want to go ahead, my thinking is that we can create some fixtures to help with this, but I think we should be cautious of the tempting solution of creating methods to populate multiple database tables and reusing these in multiple tests. Hopefully we can find a way to reuse fixtures enough to reap the benefits of code reuse but not too much that we raise the risk of false positives higher than we are comfortable with (I think this is always going to be a trade-off when reusing fixtures). I don't have a strong opinion on what this would look like, but it's something we might want to think about.

Issue #283, opened 2021-08-31
[SYNC] - Integrate Central


andreievg:
# Description

Function that runs when all central records are pulled from the server, see https://github.com/openmsupply/org-issues/discussions/32 and main method https://github.com/openmsupply/remote-server/pull/273/files

### Scope

<ins>Includes </ins>
* Able to get records from central queue and process one at a time
* Able to translate and generate queries (storing them in array)
* Able to execute all queries in transaction
* Able to clear central_queue

<ins>Excludes</ins>

* Actual translation or generation of queries

### Tasks

- [ ] Break down tasks into digestible (by reviewer) units.
- [ ] MW job added

PR #315, opened 2021-09-09
#278 add central queue

### LINKED ISSUE #278: [SYNC] - Central queue

wlthomson:
Closes #278 (and includes reverted commits relevant to #279).


src/main.rs
```diff
@@ -1,7 +1,9 @@
 #![allow(where_clauses_object_safety)]
 
 use remote_server::{
-    database::{loader::get_loaders, repository::get_repositories},
+    database::{
```
wlthomson, line 4: I think some of this stuff overlaps with #283, can be removed and potentially cherry-picked in another branch?

andreievg (changes requested):
Thanks Will, a few changes please, and overall comment:

We should stick to scope of issue that's being implemented: https://github.com/openmsupply/remote-server/issues/278

Also if there are some question about naming etc, it's good to have a discussion in the issue. 

Sqlite doesn't support batch insert, diesel made a work around by doing them in a transaction, but I think that requires SqliteConnection vs pooled one, i'll have a look now

src/database/repository/diesel/sync_buffer.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub async fn insert_many(
+        &self,
+        sync_buffer_rows: Vec<SyncBufferRow>,
+    ) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::sync_buffer::dsl::*;
+        let connection = get_connection(&self.pool)?;
+        diesel::insert_into(sync_buffer)
+            .values(sync_buffer_rows)
+            .execute(&connection)
+            .expect("");
```
andreievg, line 41: just noticed, no panics please

andreievg (changes requested):

src/database/repository/mock/item.rs
```diff
@@ -46,7 +46,7 @@ impl ItemRepository {
         info!("Querying item record (item.id={})", id);
         match self.mock_data.lock().unwrap().get(&id.to_string()) {
             Some(DatabaseRow::Item(item)) => Ok(item.clone()),
-            _ => Err(RepositoryError {
+            _ => Err(RepositoryError::DBError {
```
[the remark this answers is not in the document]
wlthomson: Agree. Branched before agreed to close that PR. Have cherry-picked it out.

src/main.rs
wlthomson: Reverted changes to `main.rs`. As noted in ba3cefb3b6180588fb70e1235ffc0291c9d80394 reverted changes may still be useful later as a reference for #283.

src/database/schema/mod.rs
```diff
@@ -23,6 +24,7 @@ pub enum DatabaseRow {
     TransactLine(TransactLineRow),
     UserAccount(UserAccountRow),
     SyncOut(SyncOutRow),
+    SyncBuffer(SyncBufferRow),
```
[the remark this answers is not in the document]
wlthomson: True. Now it's there I don't think there is much utility in removing it in isolation at this point? Will remove all this in #312, anyway.

src/database/schema/sync_buffer.rs
```diff
@@ -0,0 +1,8 @@
+use super::diesel_schema::sync_buffer;
+
+#[derive(Clone, Queryable, Insertable, Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
wlthomson: Much nicer, but note that this is a bit of an outlier in that it's not involved in the graphql schema. Removing the `Clone` trait implementation works fine here (given that 'DatabaseRow' enum is removed), but I don't think it's as simple for the other types, unfortunately.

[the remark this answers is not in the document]
wlthomson: 83956184f3ddd4ab4d2b5068b6f91b9b6cefe6bf.

src/database/schema/mod.rs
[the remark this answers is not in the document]
wlthomson: Update: 83956184f3ddd4ab4d2b5068b6f91b9b6cefe6bf.

src/database/repository/diesel/sync_buffer.rs
wlthomson: d22257300a285b55e6ce6fa799e8ddd46fb28857.

src/main.rs
```diff
@@ -73,7 +77,7 @@ async fn main() -> std::io::Result<()> {
           sync_sender.schedule_send(Duration::from_secs(settings.sync.interval)).await;
         } => unreachable!("Sync receiver unexpectedly died!?"),
         () = async {
-          sync_receiver.listen().await;
+          sync_receiver.listen(&sync_connection, &sync_buffer_repository).await;
```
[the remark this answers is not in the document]
wlthomson: Removed this as out-of-scope, but commit history can be referenced for #279 if it remains helpful.

src/util/sync/mod.rs
```diff
@@ -65,60 +64,82 @@ impl SyncSenderActor {
...
 pub struct SyncReceiverActor {
-    connection: SyncConnection,
     receiver: MpscReceiver<()>,
 }
 
 #[allow(unused_assignments)]
 impl SyncReceiverActor {
     // Hacky method for pulling from sync_queue.
-    pub async fn pull_queued_records(&mut self) -> Vec<SyncQueueRecord> {
+    pub async fn pull_queued_records(
```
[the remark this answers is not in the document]
wlthomson: Yup, reverted (and as above can be referenced or cherry-picked from commit history if and when needed).

src/database/repository/diesel/central_sync_buffer.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    pub async fn insert_many(
+        &self,
+        central_sync_buffer_rows: &Vec<CentralSyncBufferRow>,
+    ) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::central_sync_buffer::dsl::*;
+        let connection = get_connection(&self.pool)?;
+        diesel::insert_into(central_sync_buffer)
+            .values(central_sync_buffer_rows)
+            .execute(&connection)?;
```
andreievg, line 31: this doesn't compile in sqlite, needs explicit type casting to connection (as there are some issues with pooled connection for sqlite, since queries are batched manually:

```rust
      let connection: &DBBackendConnection = &&get_connection(&self.pool)?;
        diesel::insert_into(sync_buffer)
            .values(sync_buffer_rows)
            .execute(connection)
            .expect("");
        Ok(())
```

src/database/repository/diesel/central_sync_buffer.rs
```diff
@@ -0,0 +1,50 @@
...
+        Ok(result)
+    }
+
+    pub async fn remove_all(&self) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::central_sync_buffer::dsl::*;
+        let connection = get_connection(&self.pool)?;
+        diesel::delete(central_sync_buffer).execute(&connection)?;
+        Ok(())
+    }
+}
```
andreievg, line 50: a couple of test would be great

andreievg (changes requested):
looks good, just one compilation error and a test, would be great

src/database/repository/diesel/central_sync_buffer.rs
andreievg: actually much cleaner to use fully qualified syntax for deref
```rust
let connection = &get_connection(&self.pool)?;
diesel::insert_into(sync_buffer)
            .values(sync_buffer_rows)
            .execute(Deref::deref(connection))
```

chris-morgan: For myself at least, I’d *never* write calls to `deref` or `deref_mut` manually; I’d prefer `&*connection`. But taking all into account, I believe this should work (though I haven’t tried it):

```rust
let connection = get_connection(&self.pool)?;
diesel::insert_into(sync_buffer)
            .values(sync_buffer_rows)
            .execute(&connection)
```

chris-morgan: Incidentally also, `Deref::deref` is not fully qualified; fully qualified is `<Type as Trait>::method`; `Type::method` and `Trait::method` are both partially qualified.

andreievg: Thanks @chris-morgan, `&**connection` works well (@wlthomson, can we use that one), I've spent too long trying to figure this out lol, it needs to be cast explicitly to SqliteConnection, there is some trickery that diesel does with sqlite for batched queries (manual batches them in a transaction, some unmet bound). I couldn't really figure it out, but `&connection` doesn't work, if you want to play around (
[checkdiesel.zip](https://github.com/openmsupply/remote-server/files/7136679/checkdiesel.zip)), 30 lines of code to demonstrate it.

Thanks for clarifying about fully qualified.

Can i ask, what is below called ? I can't remember, I read it somewhere but can't remember the name or can't thin of keyword to google it (been playing around with proc_macros, and it's useful to avoid importing traits)
```rust
RunQueryDsl::<DBBackendConnection>::execute(
            diesel::insert_into(sync_buffer).values(sync_buffer_rows),
            &connection,
        )?;
```
btw DBBackendConnection is either PgConnection or SqliteConnection based cfg feature flag.

chris-morgan: It’s a type inference thing. The referencing operator `&` *is* capable of dereferencing first when needed, but the problem here is loosely that it doesn’t think it needs to, because `PooledConnection<ConnectionManager<SqliteConnection>>` satisfies what `execute` is looking for in its type parameter `Conn` (a reference to some type that implements the `Connection` trait), and so when that fails in a *different* way (because the connection actually isn’t sufficient for it because of the lack of `BatchInsert`), it doesn’t try any other combinations (because that way lies madness and slow compilation). So you need to manually dereference it first (something that isn’t required terribly often).

----

As for `Type::method(receiver, arg)` versus `receiver.method(arg)`, that used to be called [universal function call syntax](https://doc.rust-lang.org/1.30.0/book/first-edition/ufcs.html), but is now called [fully qualified syntax](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name), crucially no longer including *function call* in the name (since you can use the syntax to take a reference to a method or to use an associated constant), but it seems to be fuzzy about what “*fully* qualified” means, so I’m seeking clarification in https://github.com/rust-lang/rust/issues/88799. (As a matter of simple fact, `Type::method` and `<Type>::method` are *not* fully qualified unless there’s a matching inherent method: otherwise it’ll look through trait implementations.)

----

On naming, incidentally, Rust convention would have `DbBackendConnection` (*c.f.* `PgConnection`, `SqliteConnection`, `RunQueryDsl`). Up to you lot whether you go that way or not.

andreievg: Thanks very much @chris-morgan, very good explanation

andreievg:
@wlthomson a couple of changes please, `test` and explicit deref coercion of connection in `insert_many` (so that feature sqlite compiles)

src/database/repository/diesel/central_sync_buffer.rs
wlthomson: 7ec0efd.

src/database/repository/diesel/central_sync_buffer.rs
wlthomson: 8e9903f.

wlthomson:
@andreievg Should be ready to go.

tests/repository/basic.rs
```diff
@@ -238,6 +239,42 @@ mod repository_basic_test {
         assert_eq!(item2, loaded_item);
     }
 
+    async fn central_sync_buffer_test(repo: &CentralSyncBufferRepository) {
+        let central_sync_buffer_row_a = CentralSyncBufferRow {
+            id: "1".to_string(),
+            cursor_id: 1,
+            table_name: "store".to_string(),
+            record_id: "store_a".to_string(),
+            data: "{ \"ID\": \"store_a\" }".to_string(),
```
andreievg, line 248: r#"{ "ID": "store_a" }"#

andreievg (approved):
Awesome, thanks Will ! There is one suggestion, but not required

tests/repository/basic.rs
wlthomson: a12eed8.

PR #319, opened 2021-09-10
281 simple translation full (Store)

clemens-msupply:
Question: should find_one_by_id be replaced by find_one_by_id_joined (keeping the name)? I think that is what we want here, right?

I am aware of test location and unused struct definitions concerns, will look into it later...

### DIFF
--- a/src/database/repository/diesel/store.rs
+++ b/src/database/repository/diesel/store.rs
@@ -1,14 +1,19 @@
-use super::DBBackendConnection;
+use super::{DBBackendConnection, DBConnection};
 
 use crate::database::{
     repository::{repository::get_connection, RepositoryError},
-    schema::StoreRow,
+    schema::{NameRow, StoreRow},
 };
 
 use diesel::{
     prelude::*,
     r2d2::{ConnectionManager, Pool},
 };
+
+pub struct Store {
+    pub id: String,
+    pub name: NameRow,
+}
 
 #[derive(Clone)]
 pub struct StoreRepository {
@@ -20,6 +25,27 @@ impl StoreRepository {
         StoreRepository { pool }
     }
 
+    #[cfg(feature = "postgres")]
+    pub fn upsert_one_tx(connection: &DBConnection, row: &StoreRow) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::store::dsl::*;
+        diesel::insert_into(store)
+            .values(row)
+            .on_conflict(id)
+            .do_update()
+            .set(row)
+            .execute(connection)?;
+        Ok(())
+    }
+
+    #[cfg(feature = "sqlite")]
+    pub fn upsert_one_tx(connection: &DBConnection, row: &StoreRow) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::store::dsl::*;
+        diesel::replace_into(store)
+            .values(row)
+            .execute(connection)?;
+        Ok(())
+    }
+
     pub async fn insert_one(&self, store_row: &StoreRow) -> Result<(), RepositoryError> {
         use crate::database::schema::diesel_schema::store::dsl::*;
         let connection = get_connection(&self.pool)?;
@@ -27,6 +53,22 @@ impl StoreRepository {
             .values(store_row)
             .execute(&connection)?;
         Ok(())
+    }
+
+    pub async fn find_one_by_id_joined(&self, store_id: &str) -> Result<Store, RepositoryError> {
+        use crate::database::schema::diesel_schema::{
+            name_table::dsl::name_table,
+            store::dsl::{id, store},
+        };
+        let connection = get_connection(&self.pool)?;
+        let result: (StoreRow, NameRow) = store
+            .inner_join(name_table)
+            .filter(id.eq(store_id))
+            .first(&connection)?;
+        Ok(Store {
+            id: result.0.id,
+            name: result.1,
+        })
     }
 
     pub async fn find_one_by_id(&self, store_id: &str) -> Result<StoreRow, RepositoryError> {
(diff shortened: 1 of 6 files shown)


src/database/repository/diesel/store.rs
```diff
@@ -29,6 +55,22 @@ impl StoreRepository {
         Ok(())
     }
 
+    pub async fn find_one_by_id_joined(&self, store_id: &str) -> Result<Store, RepositoryError> {
```
wlthomson, line 58: Really sorry, I really didn't want to bring this up again. I'll make this the last thing I say on the matter for a while, but in return, I'd really appreciate it if you guys just took the time to read it and think about it. It's fine if you still disagree afterwards, and we can talk about it another time, so there's no need to hold this PR up with a discussion.

It seems to me that repos are being used to group utility methods for common diesel queries (as opposed to building a layer between the db model and the domain model). I wonder whether this is an indication that the repo pattern isn't appropriate for us and whether we should just make our lives easier by getting rid of repos and replacing them with utility methods. Instead of having to use a repository registry, the gql resolvers could just import the functions they need (passing them a connection in the same way as the `tx` methods, thus removing the need for the repo state).

Just to explain what I mean by us not building a layer, we now have a data-access pattern with the following properties:

- sometimes the repo needs to only know about the data domain, sometimes the repo needs to know about the business domain, e.g. `find_one_by_id_joined` (returns a gql object) vs. `find_one_by_id` (returns a database row object).
- sometimes require the caller to have knowledge of the underlying storage mechanism, sometimes abstract the underlying storage mechanism from the caller, e.g. the `tx` methods (as they tale a connection param) vs. other methods (hide the connection from the caller).
- sometimes map to a table, sometimes map to a domain and hide the underlying tables, e.g.`StoreRepository` vs. `SyncRepository`.

Due to these multitude of use-cases, it seems to me like the direction we are going in has made the concept of a repo more of a grouping utility rather than an access layer? I can't really see what the concept of a repo is now giving us over just taking the repo functions and exposing them directly as helpers, e.g. `item_repo.find_item_by_id(item_id) => find_item_by_id(&connection, item_id)`? 

As promised, last thing I'll say on the matter for a while!

wlthomson (approved):
Approving to avoid blocking milestone progress (if you could read comment when you have time, would appreciate it).

andreievg:
> Question: should find_one_by_id be replaced by find_one_by_id_joined (keeping the name)? I think that is what we want here, right?

So far the only need is in the test that you've made in this PR. It's possible that we would need it further down the road (we should implement it then, imo)

If test in individual translations are evolving to a test in this PR, I would have to back pedal my 'back down' of comment in previous PR and strongly disagree that it's a good practice. This kind of test should live in `import_sync_records` scope if it's going to be testing other translations.

I actually I don't understand the resistance against just testing things in scope (was ok to compromise on this previously, but now that' it's evolving to testing database references, i want to put on the breaks) , i started making an example, and just decided to make a PR showing what i mean: https://github.com/openmsupply/remote-server/pull/320

src/util/sync/translation/store.rs
```diff
@@ -0,0 +1,132 @@
...
+    name: String,
+    code: String,
+    mwks_export_mode: String,
+    IS_HIS: bool,
+    sort_issues_by_status_spare: bool,
+    disabled: bool,
+    responsible_user_ID: String,
+    organisation_name: String,
+    address_1: String,
+    address_2: String,
```
andreievg, line 23: i really don't like this. mainly because it requires extra out of scope work

andreievg: would recommend removing. btw definitely wont' be using all the field form mSupply. and in upcoming pr: https://github.com/openmsupply/remote-server/pull/320, it's clear what fields are in sync record

src/util/sync/translation/mod.rs
```diff
@@ -44,6 +45,18 @@ fn do_translation(
 
         return Ok(());
     }
+    if let Some(row) = LegacyStoreRow::try_translate(sync_record)? {
+        // TODO: move this check up when fetching/validating/reordering the sync records?
+        // ignore stores without name
+        if row.name_id == "" {
```
andreievg, line 51: i think this should be in LegacyStoreRow, as per: https://github.com/openmsupply/remote-server/pull/320.
keep translation logic for each record in one place. 
also I've changed it to match by code (there are other ones like drug registration that have name_id but name actually doesn't exist

andreievg (commented):
Thanks Clemens, there is also name_store_join and master list related records. And I think there is `code` for `name`, store and `item`. I think it's ok to merge this one and make a seperate PR, maybe branched on testing PR i've made if you happy with it.

src/util/sync/translation/store.rs
clemens-msupply: Yes have a ticket to remove it once translation is done

src/database/repository/diesel/store.rs
clemens-msupply: Yes the repo pattern is a thin layer over the underlying storage, so yes it can be seen as a collection of utils functions.

Btw. to actually be able to make full use of the pattern it should be behind an interface/trait and use dynamic dispatch (what we are not doing). Main advantage of the Repo pattern for me:
  - All low level db and orm code is in one place and doesn't leak into the rest of the code base
  - You can (easier) do cool stuff like: swap DBs, access a DB over the network, request caching, logging... all in a transparent way.
  - mock testing

Some useful bits here and there:
https://stackoverflow.com/questions/8749153/why-use-repository-pattern-or-please-explain-it-to-me/11064420

clemens-msupply:
I removed `find_one_by_id_joined` since its not really needed right now.

@andreievg not quite sure about your PR, was it meant to be merged? LGTM should I merge it now?...

andreievg:
> @andreievg not quite sure about your PR, was it meant to be merged? LGTM should I merge it now?...

It's branched from this branch, so if you ok with merging it in this work that be good, or i can merged it into main straight after this is merged

andreievg:
I have to fix conflicts, will do it straight after standup

clemens-msupply:
ok thanks! yes please merge

PR #324, opened 2021-09-12
Move full legacy row type definitions into a doc file

clemens-msupply:
How about that?


andreievg:
> How about that?

Thank you

andreievg:
I think I just caused a conflict

andreievg (approved):

clemens-msupply:
no problem will resolve and merge

PR #325, opened 2021-09-13
Add additional columns to item, name and store

clemens-msupply:
Add code, is_customer and is_supplier to existing tables

### DIFF
--- a/src/util/sync/translation/name.rs
+++ b/src/util/sync/translation/name.rs
@@ -4,26 +4,14 @@ use crate::database::schema::NameRow;
 
 use serde::Deserialize;
 
+#[allow(non_snake_case)]
 #[derive(Deserialize)]
 pub struct LegacyNameRow {
-    #[serde(rename = "ID")]
-    id: String,
+    ID: String,
     name: String,
     code: String,
     customer: bool,
     supplier: bool,
-}
-
-impl From<&LegacyNameRow> for NameRow {
-    fn from(t: &LegacyNameRow) -> NameRow {
-        NameRow {
-            id: t.id.to_string(),
-            name: t.name.to_string(),
-            code: t.code.to_string(),
-            is_customer: t.customer,
-            is_supplier: t.supplier,
-        }
-    }
 }
 
 impl LegacyNameRow {
@@ -33,7 +21,13 @@ impl LegacyNameRow {
         }
         let data = serde_json::from_str::<LegacyNameRow>(&sync_record.data)
             .map_err(|_| "Deserialization Error".to_string())?;
-        Ok(Some(NameRow::from(&data)))
+        Ok(Some(NameRow {
+            id: data.ID.to_string(),
+            name: data.name.to_string(),
+            code: data.code.to_string(),
+            is_customer: data.customer,
+            is_supplier: data.supplier,
+        }))
     }
 }
 
--- a/src/util/sync/translation/name.rs
+++ b/src/util/sync/translation/name.rs
@@ -9,6 +9,9 @@ pub struct LegacyNameRow {
     #[serde(rename = "ID")]
     id: String,
     name: String,
+    code: String,
+    customer: bool,
+    supplier: bool,
 }
 
 impl From<&LegacyNameRow> for NameRow {
@@ -16,6 +19,9 @@ impl From<&LegacyNameRow> for NameRow {
         NameRow {
             id: t.id.to_string(),
             name: t.name.to_string(),
+            code: t.code.to_string(),
+            is_customer: t.customer,
+            is_supplier: t.supplier,
         }
     }
 }
--- a/src/util/sync/translation/test_data/item.rs
+++ b/src/util/sync/translation/test_data/item.rs
@@ -13,6 +13,7 @@ pub fn get_test_item_records() -> Vec<TestSyncRecord> {
         translated_record: TestSyncDataRecord::Item(Some(ItemRow {
             id: "8F252B5884B74888AAB73A0D42C09E7F".to_owned(),
             name: "Non stock items".to_owned(),
+            code: "NSI".to_owned(),
         })),
         identifier: "Non stock items",
         sync_record: SyncRecord {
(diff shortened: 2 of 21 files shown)


src/util/sync/translation/name.rs
```diff
@@ -9,13 +9,19 @@ pub struct LegacyNameRow {
     #[serde(rename = "ID")]
     id: String,
     name: String,
+    code: String,
+    customer: bool,
```
andreievg, line 13: i was hoping we can translate to is_customer and is_supplier here, would mean all our basic translations are in this struct definition. And the from becomes direct mapping. What do you think ?

src/util/sync/translation/test_data/item.rs
```diff
@@ -13,6 +13,7 @@ pub fn get_test_item_records() -> Vec<TestSyncRecord> {
         translated_record: TestSyncDataRecord::Item(Some(ItemRow {
             id: "8F252B5884B74888AAB73A0D42C09E7F".to_owned(),
```
wlthomson, line 14: Can we add these to and import them from `src/database/mocks`?

Also, it's a lot easiser when dealing with mock data to have identifiers which are memorable and not prone to typos (especially when dealing with relational data with FK constraints). Can we follow the existing pattern of having meaningful identifiers like `item_row_a` rather than random UUIDs? Or, if we want to restrict our mock data to valid UUIDs, something like e.g. `"AAAAAA...AA"`, `"AAAAAA...AB"` etc.? UUIDs are uniformly distributed, so AFAICS there is no advantage to having meaningless UUIDs over human-readable ones (i.e. we are not making our mock data any better by avoiding using meaningful UUIDs).

If later down the road we want to implement UUID validation logic, we should then design mock data with a range of valid and invalid UUIDs specifically for that purpose. That's a separate deal than what we are testing here (and again, still should't use random UUIDs, anyway), so I don't think there's a reason to avoid adopting helpful UUIDs in the general purpose mock data?

wlthomson (changes requested):
Mostly LGTM, just a minor re. the mock data and mirroring @andreievg re. being consistent with doing simple mappings on deserialisation.

src/util/sync/translation/test_data/item.rs
andreievg: @wlthomson you might have missed this PR btw: https://github.com/openmsupply/remote-server/pull/320

this data is coming directly from mSupply. I can see your point that there is a potential for typos (btw test shoulnd't pass if there is a typo), but there will be a lot of data here (like a record for every variation of enum), and copy paste like this might be the quickest way to do it. Also there is extra work of mocking UUIDS, why not just use actuall data ?

> If later down the road we want to implement UUID validation logic, we should then design mock data with a range of valid and invalid UUIDs specifically for that purpose

Are we going to validate UUIDs generated by central server ?

andreievg: > especially when dealing with relational data with FK constraints

Another reason to use real data is to make sure those FK dependencies are met (we've already found a few cases while using real data where store 'name_id' is null or 'name_id' is specified but name record exists, so had to put extra conditions in translations to record ingore those special mSupply stores (which of course would have been missed if we were not using real mSupply data)

src/util/sync/translation/name.rs
clemens-msupply: I think its better to stay closer to the original data struct to avoid confusion. We have to do one copy anyway to copy LegacyNameRow to NameRow.

andreievg: Hmm, but we are already translating fields with serde. Why do it in two places ?
I guess we can move it all to from, but then we would have to re-implement features that serde has out of the box (like enums conversions, defaults, error handling, validation, etc. (I guess we can use serde just for data type conversion, not field renaming, but to me, seeing it all in one place is less confusing)

I can't say I am a fan of duplication in from, there are ways to make it dry thought. And technically speaking we could actually put deserialise on Row rather then LegacyRow (but then it's hard to do more complex translations, where some fields need to be deserialised but don't go into Row, but are used in complex translation).

> We have to do one copy anyway to copy LegacyNameRow to NameRow.

Technically if we are translating in serde then it's not quite LegacyNameRow, but TranslatedNameRow. 

What do you think ?

clemens-msupply: We still need to convert to NameRow so why not stay explicit? We can still use serde to parse enums if needed... 

Is performance your concern?

andreievg: > Is performance your concern?

Not for sync

Have the same concern is you, code cleanliness (reducing confusion), in which case consistency is very important.

I guess I am ok to  moving renaming from serde and just use it for deserialising into correct data type, but not ok with having renaming in different places, so if we moving renaming then:
* make sure we have `from` for all LegacyRow -> Row (add it in LegacyItemRow and LegacyStoreRow), or move `from` to NameRow to `impl` of LegacyRow.
* remove renaming from derives

andreievg: what were the warnings that rename was fixing ?

andreievg: i guess I'll wait and see in a commit

src/util/sync/translation/test_data/item.rs
clemens-msupply: @wlthomson I think this can be resolved?

andreievg (approved):

src/util/sync/translation/test_data/item.rs
wlthomson: Yeah, can resolve. Might be worth making an issue to migrate the existing mock fixtures to the new way that test data is organised, though?

src/util/sync/translation/name.rs
wlthomson: Since I agreed with @andreievg initial requested change, posting to say I also don't mind where it is done as long as it is consistent. The main benefit of renaming with serde is to keep the backend structs neat, i.e avoid having structs with `camelCase` fields etc, which I think will cause warnings as rust is quite strict on naming conventions.

wlthomson (approved):

src/util/sync/translation/name.rs
clemens-msupply: I disabled the warnings for these "raw" data structs

PR #329, opened 2021-09-14
#328 Add logic for pulling records from central_records endpoint

### LINKED ISSUE #328: Add logic for pulling from central_records
Add logic for pulling records from `central_records` endpoint.

wlthomson:
Fixes #328.

### DIFF
--- a/src/util/sync/connection.rs
+++ b/src/util/sync/connection.rs
@@ -97,12 +97,12 @@ impl SyncConnection {
     // TODO: add return type.
     pub async fn acknowledge_records(
         &self,
-        records: &Vec<SyncQueueRecord>,
+        records: &Vec<RemoteSyncRecord>,
     ) -> Result<(), reqwest::Error> {
         // TODO: add error handling.
         let url = self.server.acknowledge_records_url();
 
-        let body: SyncQueueAcknowledgement = SyncQueueAcknowledgement {
+        let body: RemoteSyncAcknowledgement = RemoteSyncAcknowledgement {
             sync_ids: records
                 .into_iter()
                 .map(|record| record.sync_id.clone())
--- a/src/util/sync/connection.rs
+++ b/src/util/sync/connection.rs
@@ -1,7 +1,7 @@
 use crate::util::{
     settings::SyncSettings,
     sync::{
-        SyncCredentials, SyncQueueAcknowledgement, SyncQueueBatch, SyncQueueRecord, SyncServer,
+        RemoteSyncAcknowledgement, RemoteSyncBatch, RemoteSyncRecord, SyncCredentials, SyncServer,
     },
 };
 
@@ -34,12 +34,12 @@ impl SyncConnection {
         }
     }
 
-    // Initialize sync queue.
+    // Initialize remote sync queue.
     //
     // Should only be called on initial sync or when re-initializing an existing data file.
     //
     // TODO: add custom error to return type.
-    pub async fn initialize(&self) -> Result<SyncQueueBatch, reqwest::Error> {
+    pub async fn initialize(&self) -> Result<RemoteSyncBatch, reqwest::Error> {
         // TODO: add error handling.
         let url = self.server.initialize_url();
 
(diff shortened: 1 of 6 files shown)


src/util/sync/connection.rs
```diff
@@ -84,7 +85,41 @@ impl SyncConnection {
...
-        let sync_batch = response.json::<SyncQueueBatch>().await?;
+        let sync_batch = response.json::<RemoteSyncBatch>().await?;
+
+        Ok(sync_batch)
+    }
+
+    // Pull batch of records from central sync log.
+    //
+    // TODO: add custom error to return type.
+    pub async fn central_records(
```
andreievg, line 96: maybe pull_central_records for the name ?

wlthomson: This is already called by a method called `pull_central_records` in `sync/mod.rs`, so I think that might be a bit confusing. I went with the name of the endpoints for the connection methods just to make it a bit more transparent.

andreievg (changes requested):
Thanks Will, look good, i would assume there will be further PRs for:
* Error Handling
* Persistant storage in to central_buffer_queue, and cursor itself

A couple of comments:

* I think pull_central_records would save to central_buffer_queue for every loop iteration rather then return Vec (but that's in upcoming PR by the looks of it ?) 
* The 'data' bit of central buffer sync record would need to be deserialised and saved into a text field, i've played around, this looks like this simplest way ?
```rust

#[derive(Deserialize, Serialize, Debug)]
pub struct Record {
    id: String,
    #[serde(deserialize_with = "map_as_string")]
    data: String,
}

fn map_as_string<'de, D>(deserializer: D) -> Result<String, D::Error>
where
    D: Deserializer<'de>,
{
    use serde_json::*;
    to_string_pretty(&Value::deserialize(deserializer)?)
        .map_err(|err| serde::de::Error::custom(err.to_string()))
}

fn main() {
    let check: Record = serde_json::from_str(r#"{"id": "1", "data": {"one": "two"}}"#).unwrap();
    println!("{:#?}", check);

```

Deserialisation of CentralSyncRecordData

Only change request is test compilations for `sqlite` feature

wlthomson:
> 
> 
> Thanks Will, look good, i would assume there will be further PRs for:
> 
>     * Error Handling
> 
>     * Persistant storage in to central_buffer_queue, and cursor itself

Yup, tried to keep the scope of this relatively tight.
> 
> A couple of comments:
> 
>     * I think pull_central_records would save to central_buffer_queue for every loop iteration rather then return Vec (but that's in upcoming PR by the looks of it ?)

Yup, that's the idea!

>     * The 'data' bit of central buffer sync record would need to be deserialised and saved into a text field, i've played around, this looks like this simplest way ?
> 
> 
> ```rust
> #[derive(Deserialize, Serialize, Debug)]
> pub struct Record {
>     id: String,
>     #[serde(deserialize_with = "map_as_string")]
>     data: String,
> }
> 
> fn map_as_string<'de, D>(deserializer: D) -> Result<String, D::Error>
> where
>     D: Deserializer<'de>,
> {
>     use serde_json::*;
>     to_string_pretty(&Value::deserialize(deserializer)?)
>         .map_err(|err| serde::de::Error::custom(err.to_string()))
> }
> 
> fn main() {
>     let check: Record = serde_json::from_str(r#"{"id": "1", "data": {"one": "two"}}"#).unwrap();
>     println!("{:#?}", check);
> ```
> 
> Deserialisation of CentralSyncRecordData
> 
> Only change request is test compilations for `sqlite` feature

As you said: for a later PR (but will revisit this when I get to it 😄) .

src/util/sync/connection.rs
```diff
@@ -84,7 +85,41 @@ impl SyncConnection {
...
+        &self,
+        cursor: u32,
+        limit: u32,
+    ) -> Result<CentralSyncBatch, reqwest::Error> {
+        let url = self.server.central_records_url();
+
+        // TODO: add constants for query parameters.
+        let query = [
+            ("cursor", &cursor.to_string()),
+            ("limit", &limit.to_string()),
```
clemens-msupply, line 106: keep it simple or will these constants be used anywhere else?

clemens-msupply (approved):
LGTM

src/util/sync/connection.rs
wlthomson: Will probably be used more than once, but even if this wasn't the case I think I would still make constants for strings literals which are part of the API spec. It's not a huge deal but takes very little time and introduces no additional dependencies (we already have the endpoints imported as constants). It's just nice and tidy to have all API literals grouped together in a nice lexical scope and not spread across the codebase.

clemens-msupply: overengineering?

andreievg (approved):

src/util/sync/connection.rs
wlthomson: I'm struggling to see how assigning string literals as constants is overengineering, or how having constants for endpoints is any different from this? Overengineering implies added complexity or inefficiencies, I can't see either of those here... I don't quite see the benefit of not doing this except for saving a few minutes that have already been used by this discussion!

I think this is a trivial thing and not worth the time discussing, so let's just leave it as is, I'll remove the TODO in the next PR I do.

PR #330, opened 2021-09-14
281 master list translations

### LINKED ISSUE #281: [SYNC] - Simple Translation Full - M1

clemens-msupply:
Adds `master_list`, `master_list_line` and `master_list_name_join` tables and translations.

Closes #281

### DIFF
--- a/src/database/repository/diesel/master_list.rs
+++ b/src/database/repository/diesel/master_list.rs
@@ -10,7 +10,6 @@ use diesel::{
     r2d2::{ConnectionManager, Pool},
 };
 
-#[derive(Clone)]
 pub struct MasterListRepository {
     pool: Pool<ConnectionManager<DBBackendConnection>>,
 }
--- a/src/database/repository/diesel/master_list.rs
+++ b/src/database/repository/diesel/master_list.rs
@@ -1,0 +1,57 @@
+use super::{DBBackendConnection, DBConnection};
+
+use crate::database::{
+    repository::{repository::get_connection, RepositoryError},
+    schema::MasterListRow,
+};
+
+use diesel::{
+    prelude::*,
+    r2d2::{ConnectionManager, Pool},
+};
+
+#[derive(Clone)]
+pub struct MasterListRepository {
+    pool: Pool<ConnectionManager<DBBackendConnection>>,
+}
+
+impl MasterListRepository {
+    pub fn new(pool: Pool<ConnectionManager<DBBackendConnection>>) -> Self {
+        MasterListRepository { pool }
+    }
+
+    #[cfg(feature = "postgres")]
+    pub fn upsert_one_tx(
+        connection: &DBConnection,
+        row: &MasterListRow,
+    ) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::master_list::dsl::*;
+
+        diesel::insert_into(master_list)
+            .values(row)
+            .on_conflict(id)
+            .do_update()
+            .set(row)
+            .execute(connection)?;
+        Ok(())
+    }
+
+    #[cfg(feature = "sqlite")]
+    pub fn upsert_one_tx(
+        connection: &DBConnection,
+        row: &MasterListRow,
+    ) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::master_list::dsl::*;
+        diesel::replace_into(master_list)
+            .values(row)
+            .execute(connection)?;
+        Ok(())
+    }
+
+    pub async fn find_one_by_id(&self, id: &str) -> Result<MasterListRow, RepositoryError> {
+        use crate::database::schema::diesel_schema::master_list::dsl::*;
+        let connection = get_connection(&self.pool)?;
+        let result = master_list.filter(id.eq(id)).first(&connection)?;
+        Ok(result)
+    }
+}
--- a/src/database/repository/diesel/master_list_name_join.rs
+++ b/src/database/repository/diesel/master_list_name_join.rs
@@ -1,0 +1,57 @@
+use super::{DBBackendConnection, DBConnection};
+
+use crate::database::{
+    repository::{repository::get_connection, RepositoryError},
+    schema::MasterListNameJoinRow,
+};
+
+use diesel::{
+    prelude::*,
+    r2d2::{ConnectionManager, Pool},
+};
+
+#[derive(Clone)]
+pub struct MasterListNameJoinRepository {
+    pool: Pool<ConnectionManager<DBBackendConnection>>,
+}
+
+impl MasterListNameJoinRepository {
+    pub fn new(pool: Pool<ConnectionManager<DBBackendConnection>>) -> Self {
+        MasterListNameJoinRepository { pool }
+    }
+
+    #[cfg(feature = "postgres")]
+    pub fn upsert_one_tx(
+        connection: &DBConnection,
+        row: &MasterListNameJoinRow,
+    ) -> Result<(), RepositoryError> {
+        use crate::database::schema::diesel_schema::master_list_name_join::dsl::*;
+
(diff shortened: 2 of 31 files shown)


src/database/repository/diesel/master_list_name_join.rs
```diff
@@ -0,0 +1,57 @@
...
+    schema::MasterListNameJoinRow,
+};
+
+use diesel::{
+    prelude::*,
+    r2d2::{ConnectionManager, Pool},
+};
+
+#[derive(Clone)]
+pub struct MasterListNameJoinRepository {
```
wlthomson, line 14: This seems a bit weird to me... I think we've both agreed that the idea of repos is to abstract away database details. I would expect many-to-many relations between master lists and names to be abstracted behind `MasterListRepository` and `NameRepository`/`StoreRepository` etc. rather than having a repo specific to `MasterListNameJoin` (which is an artefact of how the many-to-many relationship is modelled relationally). 

I find it hard to see how repos are abstracting persisted storage details when we are directly exposing constructs which are specific to relational databases. It seems to me also that this approach is going to require more SQL queries, as opposed to doing the join in a single query (e.g. `store_repo.find_visible_items` vs. `master_list_repo.find_many_by_store_id` + `item_repo.find_many_by_master_list_id`). 

We will need to insert join records for sync, but I think that could be done relatively easily via `master_list_repo.join_master_list_to_name(master_list_id, name_id)` or `name_repo.join_name_to_master_list(name_id, master_list_id)` etc.

I think this comes back to the point about repos serving as mappers between the domain and the underlying storage (not just as structs for grouping CRUD methods over a given table), and shows why it's not just a theoretical discussion, but has practical importance.

wlthomson (approved):
LGTM.

I have some concerns about exposing a join table as a repository, but nothing major.

src/util/sync/translation/mod.rs
```diff
@@ -144,22 +180,24 @@ mod tests {
...
+        init_records.append(&mut get_test_master_list_records());
+        let mut upsert_records = Vec::new();
+        upsert_records.append(&mut get_test_item_upsert_records());
+        upsert_records.append(&mut get_test_name_upsert_records());
+        upsert_records.append(&mut get_test_master_list_upsert_records());
 
-        import_sync_records(&registry, &extract_sync_records(records))
+        let mut records = Vec::new();
+        records.append(&mut init_records);
+        records.append(&mut upsert_records);
```
andreievg, line 194: append moves records (thus the `mut`), now this test doesn't actually test upserts. (try changing upsert TestSyncDataRecord record)

we need to either use `extend` (which i think would require both clone and copy on TestSyncRecord, which in turn would need copy and clone on SyncRecord, which I am not a fan of, it's ok to have clone, but copy would mean we can by accident be copying SyncRecord in some code rather then passing by reference)

So i would prefer just derive clone on TestSyncRecord and TestSyncDataRecord

`upsert_records.append(&mut get_test_master_list_upsert_records());`

andreievg (changes requested):
look good, a few changes

I was going to add a panic test for upsert not matching, but thought it was too much, it looks like it would have helped (I guess we can think about it later), and maybe make our test strategy also check explicit panics in the future. (btw previously, in original test PR, I was manually doing this by changing upsert expected value to make sure tests work)

src/database/repository/diesel/master_list_name_join.rs
andreievg: In order to implement `join_master_list_to_name` you need to query for existence of join record, cannot upsert it in sql query just by knowing name_id and master_list_id (and query wouldn't work as it is in the same transaction, and name or master_list may not exist).

remote will never have a configuration for central data too, this only comes in sync.

I think what we have here is fine. Sync records should be integrated one at a time not as domain objects. And in item repository (or name repository), we can do a combined domain object including master list (if it's possible given the need to split domain objects into chunks due to data loaders).

We can have a discussion about this after M1

wlthomson: Sure, those were just examples. The point is that repositories are designed to hide the underlying persistence layer, not expose it directly. We want to expose it directly, which is fine, but I think if we want to go that route, repositories are redundant for us and just a source of unneeded complexity.

The question I'm interested in an answer to is what these repository structs are now giving us over just passing `Data<Pool>` and having utility methods which take connections. If the repos don't form a consistent abstraction layer (which they don't, the return types are quite varied now and to be fair never they never really did as I was returning row objects to start with) and they don't need to be stateful (which they don't, see the `tx` methods which take a connection param) then AFAICS they don't need to be objects and may as well just be modules with each module grouping common utility methods for a given table...we can then avoid all the issues with making respositories thread safe etc. and just worry about maintaining a pool and passing connections around to methods which wrap diesel queries.

Fine to revist this after M1.

src/database/repository/diesel/master_list.rs
```diff
@@ -0,0 +1,57 @@
...
+    repository::{repository::get_connection, RepositoryError},
+    schema::MasterListRow,
+};
+
+use diesel::{
+    prelude::*,
+    r2d2::{ConnectionManager, Pool},
+};
+
+#[derive(Clone)]
```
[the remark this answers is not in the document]
clemens-msupply: done in fc8d2d5

src/database/repository/diesel/master_list_name_join.rs
clemens-msupply: Yes, as mentioned earlier, without a trait and dynamic dispatch the repo pattern is only halve as useful. Anyway IMHO they are still provide a useful abstraction over the low level queries. And yes we don't really need a repo registry... (This would makes things even simpler regarding tx because we can just pass in a connection reference into the repos and do all operations on this reference, i.e. we don't need the static methods anymore and everything works with or without tx.)

Yes lets revise after M1

src/util/sync/translation/list_master.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub fn try_translate(sync_record: &SyncRecord) -> Result<Option<MasterListRow>, String> {
+        if sync_record.record_type != "list_master" {
+            return Ok(None);
+        }
+        let data = serde_json::from_str::<LegacyListMasterRow>(&sync_record.data)
+            .map_err(|_| "Deserialization Error".to_string())?;
+        Ok(Some(MasterListRow {
+            id: data.ID,
+            // There is no name in list_master use an empty name
+            name: "".to_string(),
```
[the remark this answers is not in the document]
clemens-msupply: thanks my bad fixed in 
88c0da4

src/util/sync/translation/mod.rs
```diff
@@ -122,15 +159,14 @@ mod tests {
         records.append(&mut get_test_name_records());
         records.append(&mut get_test_store_records());
         records.append(&mut get_test_item_records());
+        records.append(&mut get_test_master_list_records());
+        records.append(&mut get_test_master_list_line_records());
+        records.append(&mut &mut get_test_master_list_name_join_records());
```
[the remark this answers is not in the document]
clemens-msupply: done 4100bd6

src/util/sync/translation/mod.rs
clemens-msupply: Thanks good catch, I even briefly wondered why this has to be mut :) bd6d551

andreievg (approved):

Issue #286, opened 2021-08-31
Document All API specs for M1


andreievg:
Document all API requirements for M1, including queries and mutation.

PR #335, opened 2021-09-14
#286 update tranlsation in query api docs

andreievg:
Chargeable to 286

### Changes
* Use new translations
* Put implementation details in drop down and removed unnecessary mapping info
* Changed from `quantity` to explicit packSize and numberOfPack
* Types in GraphQL code blocks
* Moved available number of packs in lines to a new resolver (since it's not needed in supplier invoice type and requires use of join, which we've chosen no optimise with data loader, which requires a resolver to be efficient)

Sorry, this is probably unexpected change in some areas, hopefully it doesn't undo front end work (can't wait to get GraphQL api out there with introspection, so that this types are easy to change on front end)

### DIFF
--- a/docs/content/docs/api/mutations.md
+++ b/docs/content/docs/api/mutations.md
@@ -76,7 +76,7 @@ type InsertCustomerInvoiceInput = {
 
 Invoice lines that previously existed but are missing in `allInvoiceLines` list will be deleted. `UpsertCustomerInvoiceLineInput` can either be of [InsertCustomerInvoiceLineInput](/docs/api/mutations/#customer-invoice-line-insert) or [UpdateCustomerInvoiceLineInput](/docs/api/mutations/#customer-invoice-line-update). See below
 
-{TODO we can expand this query to also have `deletedInvoiceLines`, `partialInvoiceLines`, if and when needed}
+_{TODO we can expand this query to also have `deletedInvoiceLines`, `partialInvoiceLines`, if and when needed}_
 
 <details>
 <summary>IMPLEMENTATION DETAILS</summary>
@@ -91,7 +91,10 @@ All other fields are translated directly to snake case equivalent.
 `type` to be set as: `CUSTOMER_INVOICE`
 `store_id` to be set as current logged in store in session
 
+On status change `confirm_datetime`, `entry_datetime`, and `finalised_datetime` should be changed
+
 </details>
+&nbsp;
 
 ## CUSTOMER INVOICE LINE
 
@@ -129,7 +132,7 @@ Base table: `invoice_line`
 
 All fields are translated directly to snake case equivalent.
 
-`invoice_id` set as id of parent
+`invoice_id` set as id of parent 
 
 `stock_line` links on `stock_line.id` -> `invoice_line.stock_line_id`
 
@@ -142,6 +145,7 @@ All fields are translated directly to snake case equivalent.
 Validation of reduction to be checked against each `stock_line`, and reduction applied to `stock_line`. As per [InvoiceStatus implementation details](/docs/api/types/#enum-invoicestatus)
 
 </details>
+&nbsp;
 
 ## SUPPLIER INVOICE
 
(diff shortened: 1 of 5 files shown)


docs/content/docs/api/mutations.md
```diff
@@ -91,7 +91,10 @@ All other fields are translated directly to snake case equivalent.
 `type` to be set as: `CUSTOMER_INVOICE`
 `store_id` to be set as current logged in store in session
 
+On status change `confirm_datetime`, `entry_datetime`, and `finalised_datetime` should be changed
```
mark-prins, line 94: I'm a little confused by this. I thought `entry_datetime` was the time that the invoice was entered / created. In which case it wouldn't change when the status changes.

The `confirm` and `finalised` times would only change if those specific statuses were changed - the line implies that the `confirm_datetime` changes when the status changes to `finalised` for example.

Perhaps:
```
On status change the datetime fields are set as appropriate:
- `confirm_datetime` is set when the status is changed to `confirmed`
- `finalised_datetime` is set when the status is changed to `finalised`
```

docs/content/docs/api/queries.md
```diff
@@ -21,145 +21,147 @@ Full list of GraphQL queries is provided here is for:
...
 }
 ```
+`destination` if type is `CUSTOMER_INVOICE`, linked name record represents destination of stock movement
 
-`Base Table`: transaction
-
-<ins>Field Mapping</ins>
-
-**id** [String](/docs/api/types/#string) -> `id`
+`source` if type is `SUPPLIER_INVOICE`, linked name record represents source of stock movement
```
mark-prins, line 55: there's a push to rename these to 'outbound shipment' and 'inbound shipment' - this is to make it clearer for users, so is more of a UI concern and I sort of like having the flexibility of display terms having some separation from internal naming. We're already doing this with mSupply ( name, transact ) so I think we can cope.

just mentioning it because this is not mapping of a structural name to a display term. Here we have `invoice` with type of `SUPPLIER_INVOICE` mapping to `Inbound Shipment` - using `inbound` and `outbound` as the types is, I think, much simpler.

mark-prins (approved):
LGTM! thanks Andrei for keeping on top of all this

docs/content/docs/api/mutations.md
andreievg: thanks Mark, this is more explicit and less confusion, i've made the change in downstream branch and update this one, so should be here now

docs/content/docs/api/queries.md
andreievg: hmm, i wasn't aware of this, I would be concerned with three layers of mapping mSupply -> omSupply -> API. I think some alignment here would be beneficial, I think @craigdrown had some reserve on naming transactions as shipments. Btw not everyone will be dispensing to patients, it's a common practice in a few countries to sum up their consumption (to patients) at the end of the day an make one (transaction) invoice, this doesn't quite sounds like 'shipment'. Stock movement seems like a more appropriate term, but apparently there is more information on transaction then just stock movement.

Btw, google gives me this for 'invoice': a list of goods sent or services provided, with a statement of the sum due for these; a bill.

And this for shipment: `a quantity of goods shipped; a consignment` 

The former I think describes our schema better. I wonder if the later really describes our process better at user level ? idk

andreievg: ohh, and one other thing, it also good to know the pain points that are pushing us to make naming changes. i.e. we changed to invoice from transaction because transaction is a common database term which we quite often confuse

andreievg: And there is also the concept of service fee in an invoice..

andreievg: One more comment, sorry I feel quite strong about this: I think it should be fine to present to the user different terms (even can have simplified english translations), but i think it's quite problematic mixing term in our code (between front and back end)

mark-prins: I don't believe the discussion over terminology is finished - though I did think that 'outbound / inbound shipments' were settled. I think that's outside of this discussion anyway. agree that the front end / display terms are best to be distinct from internal naming. in which case - as you were. we currently have `transaction` rather than `invoice` in the mock API, so perhaps that already should change.

PR #337, opened 2021-09-15
#336 Persist central sync cursor

### LINKED ISSUE #336: Persist central sync cursor
Add logic and table for persisting central sync cursor. 

For now, just going to go with a standalone table for the cursor. Can refactor this out once we have a design for the preferences table.

wlthomson:
Closes #336.

Will need to be revised when we decide on a way to store preferences/settings etc. as having a whole table for a single integer value is a bit overkill 😆 .


src/util/sync/mod.rs
```diff
@@ -98,18 +112,26 @@ impl SyncReceiverActor {
...
-                    records.push(central_sync_buffer_row);
+                    central_sync_buffer_repository
+                        .insert_one(&central_sync_buffer_row)
+                        .await
+                        .expect("Failed to insert central sync record into sync buffer");
+
+                    cursor += 1;
+
+                    central_sync_cursor_repository
+                        .update_cursor(cursor)
```
wlthomson, line 132: Would be nice to be able to do `++cursor` here... oh well.

clemens-msupply, line 132: I think the cursor should be updated in one transaction together with the central_sync_buffer_row. Otherwise if the server crashes you could have the case that you try to insert the same row twice which would fail (because we don't us upsert)...

However, since the cursor is also stored in the sync_buffer row, can't you just query there for the latest cursor? This means you also don't need the cursor table.

andreievg: Was planning to drop all records in buffer table after successful integration

src/util/sync/mod.rs
```diff
@@ -79,14 +82,25 @@ pub struct SyncReceiverActor {
...
+                0
+            });
+
         // Arbitrary batch size.
         const BATCH_SIZE: u32 = 500;
 
-        let mut records: Vec<CentralSyncBufferRow> = Vec::new();
-
         loop {
             info!("Sending central sync request...");
```
andreievg, line 105: not sure if this comment is accurate, 'Requesting central sync data'

src/util/sync/mod.rs
```diff
@@ -98,18 +112,26 @@ impl SyncReceiverActor {
 
             if let Some(central_sync_records) = sync_batch.data {
                 for central_sync_record in central_sync_records {
-                    cursor += 1;
-
                     let central_sync_buffer_row = CentralSyncBufferRow {
                         id: central_sync_record.id.to_string(),
-                        cursor_id: cursor as i32,
+                        cursor_id: (cursor + 1) as i32,
```
andreievg, line 117: this i actually the ID of CentralSyncRecord (and is not necessarily cursor + 1). this also mean that cursor += 1 is not accurate

src/util/sync/mod.rs
andreievg: agree about transaction

src/util/sync/mod.rs
```diff
@@ -98,18 +112,26 @@ impl SyncReceiverActor {
...
                             .expect("Failed to stringify central sync record data"),
                     };
 
-                    records.push(central_sync_buffer_row);
+                    central_sync_buffer_repository
+                        .insert_one(&central_sync_buffer_row)
+                        .await
+                        .expect("Failed to insert central sync record into sync buffer");
+
+                    cursor += 1;
```
andreievg, line 129: as per comment above, need to use ID of incoming sync record

andreievg (changes requested):
I think I missed a PR.

I think you should deserialize directly to CentralSyncBufferRow

andreievg:
One more question, have you tried running this against mSupply ? I am going to have a go once requested changes are done btw, looking forward to seeing those sweet sync records : )

src/util/sync/mod.rs
```diff
@@ -98,18 +112,26 @@ impl SyncReceiverActor {
...
+                        cursor_id: (cursor + 1) as i32,
                         table_name: central_sync_record.table_name,
                         record_id: central_sync_record.record_id,
                         data: serde_json::to_string(&central_sync_record.data)
                             .expect("Failed to stringify central sync record data"),
                     };
 
-                    records.push(central_sync_buffer_row);
+                    central_sync_buffer_repository
+                        .insert_one(&central_sync_buffer_row)
```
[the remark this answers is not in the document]
wlthomson: > 
> 
> One more question, have you tried running this against mSupply ? I am going to have a go once requested changes are done btw, looking forward to seeing those sweet sync records : )



> 
> 
> One more question, have you tried running this against mSupply ? I am going to have a go once requested changes are done btw, looking forward to seeing those sweet sync records : )

Yup, tested working against latest mSupply sync v5!

src/util/sync/mod.rs
wlthomson: So, the `cursor_id` is not necessarily related to the ordering of the records/position of the cursor?

src/util/sync/mod.rs
wlthomson: Good point about the transaction, there's not really any proper error handling here and that would be a nice way to do it.

Re. the sync buffer row `cursor_id`, I also thought about this but I thought it was a bit of extra complexity in terms of retaining the cursor position after the buffer has been processed (and, I assume, cleared). It look's like from @andreievg comment that I made an assumption about the relationship between the `cursor_id` on the record and the cursor maintained by the API consumer.

src/util/sync/mod.rs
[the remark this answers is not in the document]
wlthomson: Yeah can be done with insert many, but it seemed to me like this was the only place where the buffer would be having records inserted... I was going to actually remove `insert_many` for that reason, but decided against it. 

I suppose if we wanted to, we could just use `insert_many` across the board in place of `insert_one`. It is somewhat nicer in terms of self-documenting code to have `insert_one` ( `insert_one` makes it clear that records should be processed sequentially, which might be important in some instances), and it's also a bit fussy to have to wrap the argument in a vector... but it's a pretty minor thing that I'm not really worried about either way.

wlthomson:
> 
> 
> I think I missed a PR.
> 
> I think you should deserialize directly to CentralSyncBufferRow

Hm, is there a reason in particular for why you think this? 

In terms of code architecture, it seems cleaner and less complex to me to keep deserialisation as a separate step from translation. It also seems like some pretty tight coupling to directly map between the serialised API data to the representation of the data in the database. I didn't do it yet, but I was thinking of having the repo take `CentralSyncRecords` directly and for the repo to handle that mapping so that the sync logic doesn't even need to know how that data is stored internally. I wasn't sure, but I was also thinking of implementing `from` for converting betwen the `CentralSyncRecord` and `CentralSyncBufferRow`.

If you want to avoid further discussion on this and you're sure that deserialising directly is the way to go, then I can do that here or we can leave it for a future PR. Up to you how you want to do it?

src/util/sync/mod.rs
andreievg: check out central queue api, you'll see gaps

andreievg:
> Hm, is there a reason in particular for why you think this?

It removes a lot of code (and would make it simpler and more readable). 

from:
```rust
       if let Some(central_sync_records) = sync_batch.data {
                for central_sync_record in central_sync_records {
                    let central_sync_buffer_row = CentralSyncBufferRow {
                        id: central_sync_record.id.to_string(),
                        cursor_id: (cursor + 1) as i32,
                        table_name: central_sync_record.table_name,
                        record_id: central_sync_record.record_id,
                        data: serde_json::to_string(&central_sync_record.data)
                            .expect("Failed to stringify central sync record data"),
                    };

                    central_sync_buffer_repository
                        .insert_one(&central_sync_buffer_row)
                        .await
                        .expect("Failed to insert central sync record into sync buffer");

                    cursor += 1;

                    central_sync_cursor_repository
                        .update_cursor(cursor)
                        .await
                        .expect("Failed to update central sync cursor");
                }
            }
```
to:
```rust
  if let Some(central_sync_records) = sync_batch.data {
                central_sync_buffer_repository.insert_main(&central_sync_records)?;
                cursor = central_sync_records.last().unwrap().sequence_id
            }
```

> In terms of code architecture, it seems cleaner and less complex to me to keep deserialisation as a separate step from translation. It also seems like some pretty tight coupling to directly map between the serialised API data to the representation of the data in the database. I didn't do it yet, but I was thinking of having the repo take CentralSyncRecords directly and for the repo to handle that mapping so that the sync logic doesn't even need to know how that data is stored internally. I wasn't sure, but I was also thinking of implementing from for converting betwen the CentralSyncRecord and CentralSyncBufferRow.

I don't see a need to over-engineer a simple task tbh, what would be a use case for the things you mention above ?

> it seems cleaner  

To me the above from and to examples is much simpler, and simpler architecture in my opinion is always cleaner.

> separate step from translation

If we can deserialise and translate directly into 'domain object', which in this case is maps straight into database type, then why wouldn't we do it ? Also you mentioned it's less complex to keep deserialisation as a seperate step, but you seem to be doing it in CentralSyncRecord deserialisation.

(unfortunately for actual sync record data translation we can't translate directly to a database type, so in that case it's better to have all mapping logic in a different place)

> If you want to avoid further discussion on this and you're sure that deserialising directly is the way to go, then I can do that here or we can leave it for a future PR. Up to you how you want to do it?

Here would be great

wlthomson:
```
  if let Some(central_sync_records) = sync_batch.data {
                central_sync_buffer_repository.insert_main(&central_sync_records)?;
                cursor = central_sync_records.last().unwrap().sequence_id
            }
```

Sorry, don't quite understand what's being suggested... AFAICS this is moving translation logic to a method on the repository, which is what I was also suggesting with:

> I was thinking of having the repo take CentralSyncRecords directly and for the repo to handle that mapping so that the sync logic doesn't even need to know how that data is stored internally.

The current approach would also look like this if the mapping was moved into the repo. Without seeing the suggested deserialisation implementation, I'm not really sure what's being suggested here.

clemens-msupply:
In general I would agree with Will so that incoming API data is decoupled from what we store in the DB, e.g. to easily handle API changes. But I think in this case we simply "buffer" the incoming data in the table and its ok to do it without the intermediate step.

src/database/schema/central_sync_buffer.rs
```diff
@@ -1,11 +1,24 @@
...
 pub struct CentralSyncBufferRow {
+    #[serde(rename = "ID")]
     pub id: String,
+    #[serde(rename = "ID")]
     pub cursor_id: i32,
+    #[serde(rename = "tableName")]
     pub table_name: String,
+    #[serde(rename = "recordId")]
     pub record_id: String,
     pub data: String,
```
andreievg, line 17: I suspect that this wouldn't quite deserialise and would require change as per: https://github.com/openmsupply/remote-server/pull/329#pullrequestreview-753397180

src/database/repository/diesel/central_sync_buffer.rs
```diff
@@ -19,15 +19,35 @@ impl CentralSyncBufferRepository {
...
+    pub async fn insert_one_and_update_cursor(
         &self,
-        central_sync_buffer_rows: &Vec<CentralSyncBufferRow>,
+        central_sync_buffer_row: &CentralSyncBufferRow,
+    ) -> Result<(), RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+        let cursor = central_sync_buffer_row.cursor_id as u32;
+        connection.transaction(|| {
+            CentralSyncBufferRepository::insert_one_tx(&connection, central_sync_buffer_row)?;
+            CentralSyncCursorRepository::update_cursor_tx(&connection, cursor)?;
```
andreievg, line 30: i think would have been better as insert_many, but i think that's just stylistic opinion. not a change request

andreievg (changes requested):
Thanks for changes, looks good, just a small request left (i think deserialisation might be slightly broken

src/database/schema/central_sync_buffer.rs
wlthomson: Yup, you suspect correctly... I spent a few hours reimplementing that pretty much identically before reading this (ah well, I learned a lot).

src/database/repository/diesel/central_sync_buffer.rs
wlthomson: I'm guessing you mean the same logic and passing `insert_many` a singleton vector? If you mean batch inserting the records, it would have to be done carefully to avoid resyncing records in the case where connectivity is lost?

wlthomson:
Should now be working with latest mSupply (`feature/sync-v5'):

```
sqlite> select * from central_sync_buffer;
1|store|9EDD3F83C3D64C22A3CC9C98CF4967C4|{
  "ID": "9EDD3F83C3D64C22A3CC9C98CF4967C4",
  "IS_HIS": false,
  "address_1": "",
  "address_2": "",
  "address_3": "",
  "address_4": "",
  "address_5": "",
  "code": "DRG",
  "created_date": "0000-00-00",
  "custom_data": null,
  "disabled": false,
  "logo": "[object Picture]",
  "mwks_export_mode": "",
  "name": "Drug Registration",
  "name_ID": "9A3F71AA4C6D48649ADBC4B2966C5B9D",
  "organisation_name": "",
  "phone": "",
  "postal_zip_code": "",
  "responsible_user_ID": "",
  "sort_issues_by_status_spare": false,
  "spare_user_1": "",
  "spare_user_10": "",
  "spare_user_11": "",
  "spare_user_12": "",
  "spare_user_13": "",
  "spare_user_14": "",
  "spare_user_15": "",
  "spare_user_16": "",
  "spare_user_2": "",
  "spare_user_3": "",
  "spare_user_4": "",
  "spare_user_5": "",
  "spare_user_6": "",
  "spare_user_7": "",
  "spare_user_8": "",
  "spare_user_9": "",
  "store_mode": "drug_registration",
  "sync_id_remote_site": 1,
  "tags": ""
}
```

P.s. it would be fun to write a deserializer for 4D picture objects 😆.

src/util/sync/mod.rs
wlthomson: I didn't observe any gaps, but I'm using the test data file. Are there any docs/specs I can read to understand the cursor_id better?

src/util/sync/mod.rs
wlthomson: 4a1734c5673487f5da7793a6a4530318342e6303.

andreievg (approved):
Awesome, thanks Will.

wlthomson:
Ready to merge!

configuration/base.yaml
```diff
@@ -2,8 +2,8 @@ server:
   port: 8000
 sync:
   port: 2048
-  username: "username"
```
wlthomson, line 5: This snuck into vcs during the merge. I guess no harm, but can revert if it's a bit messy.

wlthomson: 1274faf.

Issue #304, opened 2021-09-06
[API] - Items, names, pagination


andreievg:
MW parent job: `[OMSBEA] omSupply Back End - API`

### Description

* Add items and names graphQL query (as per docs)
* Add pagination

### Scope

<ins>Includes</ins>
* Fields mentioned in the docs
* Extra schema fields if they dont' exists yet (and correct translations)
* totalCount

<ins>Excludes</ins>
* Filters or sorting

### Tasks

- [ ] Break down tasks into digestible (by reviewer) units.
- [ ] MW job added

PR #338, opened 2021-09-16
#304 part1 - names query and pagination

andreievg:
From issues 304 (part 1)

### Changes

* Removed name query
* Added names query and NameList query (to have totalCount and nodes)
* Added pagination input object
* Added new repository NameQueryRepository (since our 'domain' object will be a join between name and name_store_join)
* Added repository specific tests (although they do use another repository to insert data, not that happy about it, but should be ok?)


src/database/repository/diesel/name_query.rs
```diff
@@ -0,0 +1,156 @@
...
+            .limit(pagination.first())
+            .load::<NameRow>(&*connection)?
+            .into_iter()
+            .map(NameRow::into)
+            .collect())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
clemens-msupply, line 55: Plan to much the other repo tests over soonish, just in case you plan to do it as well and we do it twice...

src/server/service/graphql/schema/queries/pagination.rs
```diff
@@ -0,0 +1,38 @@
...
+pub trait PaginationOption {
+    fn first(&self) -> i64;
+    fn offset(&self) -> i64;
+}
+
+pub const DEFAULT_PAGE_SIZE: i64 = 100;
+
+impl PaginationOption for Option<Pagination> {
+    fn first(&self) -> i64 {
+        match match self {
```
clemens-msupply, line 18: its that equivalent to a `.map`?

clemens-msupply (commented):
Can't say too much about graphql but rest looks good to me.

src/database/repository/diesel/mod.rs
```diff
@@ -91,6 +93,7 @@ pub async fn get_repositories(settings: &Settings) -> RepositoryMap {
     repositories.insert(ItemRepository::new(pool.clone()));
     repositories.insert(ItemLineRepository::new(pool.clone()));
     repositories.insert(NameRepository::new(pool.clone()));
+    repositories.insert(NameQueryRepository::new(pool.clone()));
```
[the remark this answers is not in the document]
andreievg: It returns a new type (name joined with name_store_join), you can see that in upstream pr. Don't really want to start another dialog of repositories lol, but it seems that's a different domain object (specific to query)

src/database/repository/diesel/name_query.rs
andreievg: i'll only be touching things in scope, so will also do it for `items`. There are some repositories that are not needed for M1, and beyond M1 those ones that are not used yet might change. If we move tests, maybe it's a good idea to just do repos that are in M1

src/server/service/graphql/schema/queries/pagination.rs
andreievg: i saw this in serde repo:
```rust
// double match
let result = match match start {
 match1
} {
  match2
}
// single match
let result1 = match start { match1 }
let result = match result1 {match2 }
}
```

if you think this will trip up devs in the future, i'll change to single match. (i thought it was quite neat)

andreievg:
> Can't say too much about graphql but rest looks good to me.

Thanks for review, I've manually tested graphql api, seems to work correctly (i'll make an issue for graphql tests)

src/server/service/graphql/schema/queries/pagination.rs
clemens-msupply: I actually meant and_then:

image

would that work as well?

clemens-msupply (approved):

src/server/service/graphql/schema/queries/pagination.rs
andreievg: much better, thank you: 1d13158d386a796a9ff7f8844f4fd64f6baaf500

src/database/repository/diesel/name_query.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn new(pool: Pool<ConnectionManager<DBBackendConnection>>) -> NameQueryRepository {
+        NameQueryRepository { pool }
+    }
+
+    pub fn count(&self) -> Result<i64, RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+        Ok(name_table.count().get_result(&*connection)?)
+    }
+
+    pub fn all(&self, pagination: &Option<Pagination>) -> Result<Vec<NameQuery>, RepositoryError> {
```
wlthomson, line 42: I think I'm going to surprise everyone and say... 🎉!

I think this repo is a step in the right direction! I have a few nits, but I'm not going to focus on them as I like significantly more about this than I dislike! 

This repository decouples the graphql objects from the underlying db objects (woo!), meaning anybody using the repository only needs to know about the graphql domain and doesn't need to know or care about anything to do with the database (all persistence layer details are hidden from the consumer). in the event that schema details are changed, only the repository logic needs to be updated to "map" the updated schema to the existing graphql type(s). This provides us with a nice separation of concerns and a sort of "firewall" between layers. I think this is a big positive for both software robustness and ease-of-developement.

In my ideal world we would get to a place where all repos have methods which take and return domain objects (mainly thinking of gql types, but also other domains such as sync api structs), and the schema types were not used anywhere except for in repos, which would be be resposible for a particular service or domain object, with the database tables only an implementation detail. I could see a single `SyncRepository` repository being sufficient for all sync concerns (i.e. buffering records, cursor management, saving translated buffered records). I could also see us moving towards conceptualising the graphql types as the main "reference types", rather than the db tables, e.g. instead of talking about `name` and `name_store_join` etc., we talk about the `Name` type (with `name` and related joins being implementation details). The benefit of this is that we would be speaking a common language with the front-end team.

As discussed, no need to discuss this here, just documenting for posterity.

src/database/repository/diesel/name_query.rs
```diff
@@ -0,0 +1,156 @@
...
+        let connection = get_connection(&self.pool)?;
+        Ok(name_table
+            .offset(pagination.offset())
+            .limit(pagination.first())
+            .load::<NameRow>(&*connection)?
+            .into_iter()
+            .map(NameRow::into)
```
wlthomson, line 49:
```
.map(NameRow::into)
```

So clean, really like this pattern.

src/database/repository/diesel/name_query.rs
```diff
@@ -0,0 +1,156 @@
...
+            service::graphql::schema::{
+                queries::pagination::{Pagination, DEFAULT_PAGE_SIZE},
+                types::NameQuery,
+            },
+        },
+        util::test_db,
+    };
+    use std::convert::TryFrom;
+
+    fn data() -> (Vec<NameRow>, Vec<NameQuery>) {
```
wlthomson, line 72: Is the plan to extract this data out into fixtures?

src/database/repository/diesel/name_query.rs
```diff
@@ -0,0 +1,156 @@
...
+            name_row_respository.insert_one(&row).await.unwrap();
+        }
+
+        let default_page_size = usize::try_from(DEFAULT_PAGE_SIZE).unwrap();
+
+        // Test
+        let repository = registry.get::<NameQueryRepository>();
+
+        // .count()
+        assert_eq!(
```
wlthomson, line 114: I think it would be nice to get to a place where these "subtests" are broken out into individual tests. I know it would be inefficent in terms of setting up and tearing down the db, but I think we could quite easily make helpers for that, and I suspect once the codebase gets sufficiently large it might actually make tests faster due to leveraging concurrency (one method with all tests sequentially versus all tests at once in parallel). This is actually the approach suggested in zero2production, so I'm borrowing the author's authority to argue for it!

Another "design decision" which can be left for the next sprint discussion.

src/server/service/graphql/schema/queries/mod.rs
```diff
@@ -6,12 +6,18 @@ use crate::database::schema::{
...
 use crate::server::service::graphql::ContextExt;
 
 use async_graphql::{Context, Object};
 
+use self::pagination::Pagination;
+
+use super::types::NameList;
+
+pub mod pagination;
+
```
wlthomson, line 20: I think the pagination import can be made a bit cleaner by moving it after the module declaration?

```suggestion
use super::types::NameList;

pub mod pagination;

use pagination::Pagination;
```

src/server/service/graphql/schema/queries/mod.rs
```diff
@@ -21,19 +27,12 @@ impl Queries {
...
-        let name_row: NameRow = name_repository
-            .find_one_by_id(&id)
-            .await
-            .unwrap_or_else(|_| panic!("Failed to get name {}", id));
-
-        Name { name_row }
+        _ctx: &Context<'_>,
+        #[graphql(desc = "pagination (first and offset)")] page: Option<Pagination>,
+    ) -> NameList {
+        NameList { pagination: page }
```
wlthomson, line 35: Personally, I would prefer the call to the repo being lifted up here and `NameList` being a dumb wrapper over `vec<Name>`... both ways work 🤷‍♂️.

src/server/service/graphql/schema/queries/pagination.rs
```diff
@@ -0,0 +1,30 @@
...
+pub trait PaginationOption {
+    fn first(&self) -> i64;
+    fn offset(&self) -> i64;
+}
+
+pub const DEFAULT_PAGE_SIZE: u32 = 100;
+
+impl PaginationOption for Option<Pagination> {
+    fn first(&self) -> i64 {
+        self.as_ref()
```
wlthomson, line 18: What's the `as_ref` for?

wlthomson (approved):
LGTM :).

src/server/service/graphql/schema/types/name.rs
```diff
@@ -0,0 +1,31 @@
+use crate::database::repository::NameQueryRepository;
+use crate::server::service::graphql::{schema::queries::pagination::Pagination, ContextExt};
+use async_graphql::{Context, Object, SimpleObject};
+
+#[derive(SimpleObject, PartialEq, Debug)]
+#[graphql(name = "Name")]
+pub struct NameQuery {
```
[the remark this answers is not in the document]
andreievg: Mainly to differentiate that it's API specific shape

src/database/repository/diesel/name_query.rs
andreievg: Not sure about t his one tbh, since this test is specific to pagination

src/database/repository/diesel/name_query.rs
andreievg: Hmm, have to see what the benefit of this is, if it's to narrow down on what is failing, then we can add a string as last parameter of assert_eq (although when test fails it tell you which line it's on, so not sure how much benefit it is)

src/server/service/graphql/schema/queries/mod.rs
andreievg: 760c2a9ca0d2910110daf368df17169b67ab19d7

src/server/service/graphql/schema/queries/mod.rs
andreievg: I think i did it this way to reduce end size of root query file (and to keep it to a higher level of abstraction)

src/server/service/graphql/schema/queries/pagination.rs
andreievg: to avoid moving Option

PR #340, opened 2021-09-16
304 2 add iscustomer and issupplier to names query

andreievg:
### Changes

* Added name_store_join table
* In NameQuery, left join name_store_join

Didn't add any further test as I think it's a little bit out of scope to be refactoring API test atm, and name_store_join test would require adding stores and adding name_store_join (with helper methods, which will never be used in remote application). Did a manual graphql tests, seems good, but found a bug in diesel (i think)


src/database/repository/diesel/name_query.rs
```diff
@@ -2,7 +2,12 @@ use super::{get_connection, DBBackendConnection};
...
         repository::RepositoryError,
-        schema::{diesel_schema::name_table::dsl::*, NameRow},
+        schema::{
+            diesel_schema::{
+                name_store_join::dsl::name_store_join, name_table::dsl as name_table_dsl,
+            },
+            NameRow, NameStoreJoinRow,
+        },
     },
     server::service::graphql::schema::{
```
clemens-msupply, line 12: Think I now understand the Query repos a bit better (was too tired yesterday)

Don't want to start another discussion so just as a comment: Think we can have a cleaner cut here on what goes into repository and what builds on top of a repository, e.g. here the graphql schema gets imported into the repository but I think the repositories should be DB only without dependencies on the rest of the code. For example, we could have a low level NameStoreJoinRepository and the `.map(NameQuery::from).collect())` part could just be done in the code that needs it...

src/database/repository/diesel/name_query.rs
```diff
@@ -36,19 +55,38 @@ impl NameQueryRepository {
...
+    // .offset(pagination.offset())
+    //  .limit(pagination.first())).to_string());
+
+    // which results in
+
+    // SELECT "name"."id", "name"."name", "name"."code", "name"."is_customer", "name"."is_supplier", "name_store_join"."id", "name_store_join"."name_id", "name_store_join"."store_id", "name_store_join"."name_is_customer", "name_store_join"."name_is_supplier" FROM ("name" LEFT OUTER JOIN "name_store_join" ON "name_store_join"."name_id" = "name"."id");
+
+    // this is NOT equivelant to
+
+    // SELECT "name"."id", "name"."name", "name"."code", "name"."is_customer", "name"."is_supplier", "name_store_join"."id", "name_store_join"."name_id", "name_store_join"."store_id", "name_store_join"."name_is_customer", "name_store_join"."name_is_supplier"  from "name" left outer join name_store_join on name_store_join.id = "name".id;
```
clemens-msupply, line 86: Think my SQL is not good enough here. Are the parenthesis the problem? otherwise the Diesel query looks good to me. In your query, is `name_store_join on name_store_join.id = "name".id` a typo?

src/database/repository/diesel/name_query.rs
andreievg: i kind of thought that `domain object` was whatever consumer wants it (not mean to map directly to data types and their shapes, and repository needs to know about this shape), in this case consumer is API. You'll see upstream there is an ItemQuery, which require a very particular join

src/database/repository/diesel/name_query.rs
andreievg: Yes, i complete confused this (thanks for noting), 3f1ccea4114bdcce09ebfcfc5d41c385bfd1c4e9,

We would either need to aggregate with group by or subquery for this sort of join. Will have to see what the options are

src/database/repository/diesel/name_query.rs
clemens-msupply: Ok right I see the problem. I mean you could define ItemQuery in repository but then you still have GraphQL macros in it...

A shame that you can't apply a macro to a struct from a different module files (like impl a trait from a struct from a different module) this would keep it nicely separated. Just thinking out loud...

src/database/schema/name_store_join.rs
```diff
@@ -0,0 +1,11 @@
+use super::diesel_schema::name_store_join;
+
+#[derive(Queryable, Insertable, Debug, PartialEq, Eq, AsChangeset)]
+#[table_name = "name_store_join"]
+pub struct NameStoreJoinRow {
+    pub id: String,
+    pub name_id: String,
+    pub store_id: String,
+    pub name_is_customer: bool,
+    pub name_is_supplier: bool,
```
wlthomson, line 10: Are these mutuall exclusive? Is it possible for them both to have the same value?

src/server/service/graphql/schema/types/name.rs
```diff
@@ -8,9 +8,9 @@ pub struct NameQuery {
     pub id: String,
     pub name: String,
     pub code: String,
-    // will come from name_store_join not the Name record itself
-    // pub is_customer: bool,
-    // pub is_supplier: bool,
+    // Below are from name_store_join
```
wlthomson, line 11: I reckon keep the schema decoupled from the database? The repos can handle the mapping and hopefully will be self-documenting enough (but if not, no opposition to the comment in the repo).

wlthomson (approved):
LGTM!

src/database/repository/diesel/name_query.rs
```diff
@@ -15,12 +20,24 @@ use diesel::{
     r2d2::{ConnectionManager, Pool},
 };
 
-impl From<NameRow> for NameQuery {
-    fn from(name_row: NameRow) -> Self {
+type NameAndNameStoreJoin = (NameRow, Option<NameStoreJoinRow>);
```
wlthomson, line 23: I'm no so sure of the value of defining a type for this... I think it's actually clearer to not alias the tuple and just use it directly 🤷. The tuple is already relatively nice and tidy, and for this to actually represent what it aliases, i think it would need to be something like: `NameAndNameStoreJoinLeftJoinRow` 😆!

src/database/repository/diesel/name_query.rs
andreievg: Yeah that would have been great

src/database/schema/name_store_join.rs
andreievg: just following API specs

src/server/service/graphql/schema/types/name.rs
andreievg: it's just that NameRow also has is_customer and is_supplier, which is different from NameQuery, wanted to be explicit

src/database/repository/diesel/name_query.rs
andreievg: if you add another join, then have to change in three places, rather then one

PR #341, opened 2021-09-16
#304 part3 add items query

andreievg:
### Changes
* Remove item query (and others that were relying on it)
* Improve testing slightly with helper to get everything required for test and setup test
* Postgres pagination tests were failing so had to add order by id (and improve them slightly)
* Added ItemQuery type
* Added ItemQueryRepository (made master list tables joinable)

`note`: I've added test that relies on other repositories to make sure ItemQuery is working (it needed data for quite a few tables), not a fan of having 400 line method, and not a fan of such a big PR but 60% of it is test data, so should be easy enough to review. Btw pagination testing is basically copy paste form name

Tested GraphQL, seems to work good, but that issue with diesel will cause duplicate items if more then one master list has the same item. Thus in the tests i made sure there is only one master list -> item relation (although there are 2 master list in the tests)

stil coming -> stock lines on item


src/database/repository/diesel/item_query.rs
```diff
@@ -0,0 +1,342 @@
...
+            schema::{ItemRow, MasterListLineRow, MasterListNameJoinRow, MasterListRow, NameRow},
+        },
+        server::service::graphql::schema::{
+            queries::pagination::{Pagination, DEFAULT_PAGE_SIZE},
+            types::ItemQuery,
+        },
+        util::test_db,
+    };
+    // TODO this is very repetative, although it's ok for tests to be 'wet' I think we can do better (and stil have readable tests)
+    fn data() -> (Vec<ItemRow>, Vec<ItemQuery>) {
```
clemens-msupply, line 100: I think its ok. There is certainly a point where it gets too messy and you need to refactor but with test it often doesn't get that far and a basic solution is ok.

From my experience I often only realize that there is logic that needs testing while doing the development, having a simple test setup makes it much easier for me to focus on these tests (and not forget about them while designing a nice test).

src/database/repository/diesel/item_query.rs
```diff
@@ -0,0 +1,342 @@
...
+                offset: Some(150),
+                first: Some(90),
+            }))
+            .unwrap();
+        assert_eq!(result.len(), queries.len() - 150);
+        assert_eq!(result.last().unwrap(), queries.last().unwrap());
+    }
+
+    // TODO not sure where this fits, seems like this unit test has a lot of dependencies
+    // I think test snapshot-like functionality is need ?
```
clemens-msupply, line 182: I would be ok with it since it does the job.

Think this kind of test and DB snapshot are complimentary. With the former you can guarantee that you have a valid DB state while with a DB snapshot you could have a snapshot that couldn't be reached in a real system (e.g. because something changed since you last took the snapshots). Think snapshot are great to test real world error cases while with this test you have a very robust way to test the basic functionality.

src/util/test_db.rs
```diff
@@ -1,7 +1,13 @@
...
+    database::repository::{get_repositories, DBBackendConnection},
+    server::data::{RepositoryMap, RepositoryRegistry},
+};
+
 use super::settings::{DatabaseSettings, ServerSettings, Settings, SyncSettings};
 
+use diesel::r2d2::ConnectionManager;
 use diesel_migrations::{find_migrations_directory, mark_migrations_in_directory};
-
+use r2d2::{Pool, PooledConnection};
```
clemens-msupply, line 10: Any reason why you use the r2d2 directly?

src/util/test_db.rs
```diff
@@ -97,3 +103,30 @@ pub fn get_test_settings(db_name: &str) -> Settings {
         },
     }
 }
+
+pub async fn setup_all(
```
clemens-msupply, line 107: Think a doc comment would be helpful here

clemens-msupply (approved):
LGTM

There is only one comment about a missing doc comment that is not really critical and kind of option for this reason I already approved this PR. For the future, is that a ok for me to this or should I select "Request changes".

src/util/test_db.rs
andreievg: I always use auto import (never type them in), so it's something rust analyzer did: b32be634bc62ff11ac4ba13c7eae5de0c9b6280d

src/util/test_db.rs
andreievg: Like this: 9bd4c175880ceb47c9ed4e29438826789445349a ?

andreievg:
> There is only one comment about a missing doc comment that is not really critical and kind of option for this reason I already approved this PR. For the future, is it ok for me to do so or should I select "Request changes".

up to you, i always check all of the comments, and try to reply or at least thumbs up, and I'll go over them again before mergin

src/util/test_db.rs
clemens-msupply: Same happened to me before and Will complained :)

chris-morgan: dependencies.r2d2 should be removed from Cargo.toml.

andreievg: 555a11363ef608d8e5280fd83ec1dcf4b8389aa2
goneski

src/server/service/graphql/schema/types/mod.rs
```diff
@@ -1,22 +1,27 @@
...
     },
     server::service::graphql::ContextExt,
 };
 
 use async_graphql::{dataloader::DataLoader, Context, Enum, InputObject, Object};
 
+// M1 speced API is moved to their own files
+// Types defined here are prototype types and should be removed before M1 release to avoid confusion (for consumers and devs)
 pub mod name;
 pub use self::name::*;
```
wlthomson, line 20: Pretty sure `self` is redundant here?

src/util/test_db.rs
```diff
@@ -97,3 +102,35 @@ pub fn get_test_settings(db_name: &str) -> Settings {
...
+    let pool = Pool::new(connection_manager).expect("Failed to connect to database");
+
+    (
+        pool.clone(),
+        match all_repositories {
+            true => get_repositories(&settings).await,
+            false => RepositoryMap::new(),
+        },
+        pool.get().unwrap(),
+    )
```
wlthomson, line 135: This looks a bit overcooked to me (I think having to bundle together three values into the return is a bit of a indicator that this method may be doing too many things). I think it would be fine to take `settings` as a param (this is already behind a helper) and shortcut before the tuple and just return `pool`. As far as I can see that would reduce most of the boilerplate, and would also make this method a bit cleaner/clearer?

wlthomson (approved):
LGTM.

src/server/service/graphql/schema/types/mod.rs
andreievg: really ? i get a compilation error if it's removed

src/util/test_db.rs
andreievg: This method was meant to provide all of the things that a test need in one call, i.e.:
Screen Shot 2021-09-20 at 8 46 25 AM

Agree it's not that clean, but it's small and easy enough to read. happy to hear your suggestions

PR #342, opened 2021-09-16
304 4 add stock line to items

andreievg:
Sorry for big diff, i should have done renaming first. the [last commit](https://github.com/openmsupply/remote-server/commit/dc3676b633b472b8f6cc6c78feef33f8b1a8b3b5) is the main change (under 200 lines)

### Changes
* Renamed item line to stock line
* Added schema changes with new translations
* Removed old loaded and implemented loaded for StockLineRepository

Didn't know what to test, i did quite a few manual tests with GraphQL, all working fine.

(hindsight, best not too get far ahead with prototype, just show how things can work, otherwise quite a lot of changes needed when requirements are clarified)

this query should be satisfied now (including pagination).
```graphql
query {
  items {
    nodes {
      id: String!
      code: String!
      name: String!
      # item visibility in current logged in store
      isVisible: Boolean!
      # batches available in current logged in store
      availableBatches {
        nodes {
          packSize: Number!
          costPricePerPack: Float!
          sellPricePerPack: Float!
          availableNumberOfPacks: Number!
          totalNumberOfPacks: Number!
          batch: String
          expiryDate: Date
        }
      }
    }
  }
}
```


src/server/service/graphql/schema/types/stock_line.rs
```diff
@@ -0,0 +1,65 @@
+use async_graphql::{Context, Object, SimpleObject};
+use std::convert::TryInto;
+
+use crate::database::schema::StockLineRow;
+
+#[derive(SimpleObject, PartialEq, Debug)]
+#[graphql(name = "StockLine")]
+pub struct StockLineQuery {
```
andreievg, line 8: i thought of just deriving this on just deriving this on StockLineRow, but it seemed like the wrong place to do it ?

You can see the from clause is just a direct mapping from StockLineRow

src/database/repository/diesel/mod.rs
```diff
@@ -114,6 +125,8 @@ pub async fn get_repositories(settings: &Settings) -> RepositoryMap {
     repositories.insert(MasterListRepository::new(pool.clone()));
     repositories.insert(MasterListLineRepository::new(pool.clone()));
     repositories.insert(MasterListNameJoinRepository::new(pool.clone()));
+    
+    repositories.insert(DataLoader::new(StockLineRepository::new(pool.clone())));
```
andreievg, line 129: I can't see the need for having two seperate holders, one for repository and one for loaders, i think we should just combine them into one (maybe database_context ?)

src/server/service/graphql/schema/types/item.rs
```diff
@@ -12,6 +17,19 @@ pub struct ItemQuery {
     pub is_visible: bool,
 }
 
+#[ComplexObject]
+impl ItemQuery {
+    async fn available_batches(&self, ctx: &Context<'_>) -> StockLineList {
+        let repository = ctx.get_repository::<DataLoader<StockLineRepository>>();
```
andreievg, line 23: as per another comment, i think this should just be `get()` or `database_context`

src/server/service/graphql/schema/types/stock_line.rs
clemens-msupply: Would be really cool if you could do something like this:

``` rust
#[derive(...)] 
type StockLineQuery = StockLineRow;
```

src/database/repository/diesel/mod.rs
clemens-msupply: Could the loaders also be created on the stack on demand, e.g. `let loader = ItemLoader::new(repositories)`?

src/database/repository/mod.rs
```diff
@@ -11,6 +11,9 @@ pub enum RepositoryError {
     /// Foreign key constraint is violated
     #[error("foreign key constraint is violated")]
     ForeignKeyViolation,
+    /// Actix thred pool canceled
+    #[error("actix thread pool canceled")]
+    ActixThredPoolCanceled,
```
clemens-msupply, line 16: thred -> thread (in comment as well)

Is that a repository error though? Chris had some ideas about errors: https://github.com/openmsupply/remote-server/discussions/301

clemens-msupply (changes requested):

src/database/repository/diesel/mod.rs
andreievg: Hmm, is this for testing repositories?  i am not sure, i think it's async graphql specific. Also it has to be one common loader so that resolvers know that there is an async operation waiting for the operation to be fetched (i.e. when loads of items are requested, each individual resolver for stockline will be called and some magic know for them to wait for main query to finish)

src/database/repository/mod.rs
andreievg: 36a7384fbe942bbe32a3d67ef0bb80b1a31f0621

it has to map to repository error, diesel::result::Error doesn't work in the loader, there is 0 extra information in the Canceled enum of BlockedErrror btw, maybe it's good to also say where it happened (which loader ?), but it would require mapping directly (not using from), can do that if you think it's better.

clemens-msupply: The anyhow crate that Chris mentioned might be useful here, they had a feature like that. I don't have the full picture yet though how it would all fit together...

Another thought, maybe simply remove Actix from the error name to make the error more generic?

src/database/repository/diesel/mod.rs
wlthomson: The reason they're separate is just to keep it easy to remove the loaders if we decided to go with a different pattern. I agree that they should be combined: repositories probably shouldn't be used directly by graphql if we are using loaders.

andreievg: > I agree that they should be combined: repositories probably shouldn't be used directly by graphql if we are using loaders.

How do you pass a filter or sort or pagination InputObject to loaders ? (I think it's fine to use repositories in GraphQL)

src/database/repository/mod.rs
andreievg: cbf4fd258fc5dda004a8dabef56415195dcdacf0

hmm, requires a deeper thought

src/server/service/graphql/schema/types/stock_line.rs
andreievg: technically it's possible to make a macro like this (probably declarative function-like macro vs procedural, the former replaces code vs later adds to it). I think we would need to specify the source filename of StockLineRow definition (and use `syn:File` to parse that file and find StockLineRow).

This is all very deep lol, but i had the same question so thought of checking out how it could be done.

The call would be like (this is just out of interest, no way suggesting we should do it)
```rust
duplicate_struct!({
        prefix -> "#[derive(PartialEq, Debug)]"
        to -> StockLineQuery
        from -> StockLineRow
        fromLocation -> "src/database/schema/stock_line.rs",
        deriveFrom -> true
     });
```

src/database/repository/diesel/mod.rs
wlthomson: That was phrased a bit badly, I mean't to say we shouldn't call repositories directly in the instances where we are using a dataloader (i.e. the data loader should call the repository and batch the results). I agree with using both repositories and data loaders in gql, exposed via a unified registry. Does that clarify?

src/server/service/graphql/schema/types/item.rs
wlthomson: I'd be happy just with `ctx.get`!

wlthomson (approved):

src/server/service/graphql/schema/types/stock_line.rs
clemens-msupply: haha that would work but I guess you still can't pass a StockLineRow straight into a method that expect a StockLineQuery? or would there be a `from` implementation in the macro? but `from` would still unnecessarily copy the structs at runtime, wouldn't it?

andreievg:
@clemens-msupply, I think i've address pending change requests

clemens-msupply (approved):

src/database/repository/diesel/mod.rs
andreievg: what do you mean by 'call' ?
Btw we can also make DataLoader instance a field in a repository, i kind of think they are integral part of repository, i guess we can discuss this at the start of m2

PR #347, opened 2021-09-17
Use a "SyncSession" to integrate all records in a single tx

clemens-msupply:
Instead of putting more logic into the SyncRepository I introduced a SyncSession and all records integrated in the same SyncSession are stored in a single transaction.

Let me know what you think...

### DIFF
--- a/src/util/sync/mod.rs
+++ b/src/util/sync/mod.rs
@@ -15,7 +15,10 @@ pub use remote::{
 pub use server::SyncServer;
 
 use crate::{
-    database::{repository::CentralSyncBufferRepository, schema::CentralSyncBufferRow},
+    database::{
+        repository::{CentralSyncBufferRepository, SyncRepository},
+        schema::CentralSyncBufferRow,
+    },
     server::data::RepositoryRegistry,
 };
 
@@ -172,6 +175,11 @@ impl SyncReceiverActor {
     ) -> Result<(), String> {
         let central_sync_buffer_repository: &CentralSyncBufferRepository =
             repositories.get::<CentralSyncBufferRepository>();
+        let sync_session = repositories
+            .get::<SyncRepository>()
+            .new_sync_session()
+            .await
+            .unwrap();
         for table_name in TRANSLATION_RECORDS {
             let buffer_rows = central_sync_buffer_repository
                 .get_sync_entries(table_name)
@@ -186,7 +194,7 @@ impl SyncReceiverActor {
                     data: row.data,
                 })
                 .collect();
-            import_sync_records(repositories, &records).await?;
+            import_sync_records(&sync_session, repositories, &records).await?;
         }
         central_sync_buffer_repository
             .remove_all()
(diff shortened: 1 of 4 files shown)


andreievg:
does 'savepoint' between transactions mean that rollback would be up to the save point or for the whole transaction ? https://docs.diesel.rs/diesel/connection/trait.Connection.html#provided-methods

src/util/sync/mod.rs
```diff
@@ -172,6 +175,11 @@ impl SyncReceiverActor {
     ) -> Result<(), String> {
         let central_sync_buffer_repository: &CentralSyncBufferRepository =
             repositories.get::<CentralSyncBufferRepository>();
+        let sync_session = repositories
```
wlthomson, line 178: Sorry if I'm missing some context, but could you explain the reasoning for lifting the transaction management up to the sync actor, rather than keeping it the responsibility of the repo?

Is the motivation here to make iteration over the tables and the integration of those records a single atomic action? If so, my confusion is why the repo has responsibility for ensuring record integration is atomic for all records (in a given table), but not responsibility for ensuring table integration is atomic for all tables? Could be wrong, but it seems to me that the need to lift the session up to the sync actor is due to the fact that translation logic is kept outside of the sync repo? If this is the case, it might be an indication that it would be tidier to let the sync actor only have to worry about sync domain objects and let the repo handle the translation behind the scenes?

This turned into a bigger architectural question, so we can revisit this at a later sprint meeting if that is the best place to discuss!

wlthomson (approved):
Have some architectural concerns about repo abstractions being "leaked" into the sync actor, but approving as these can be tackled at a later sprint meeting.

clemens-msupply:
> does 'savepoint' between transactions mean that rollback would be up to the save point or for the whole transaction ? https://docs.diesel.rs/diesel/connection/trait.Connection.html#provided-methods

It would only be for the savepoint transaction but I guess if you don't handle the error in the parent transaction block you effectively rollback the whole transaction. Does that make sense?

src/util/sync/mod.rs
clemens-msupply: There is also a discussion here regarding tx:
https://github.com/openmsupply/remote-server/discussions/300

I think what you really want is a controller, service, repository architecture (apparently called Repository-Service Pattern)? A repository should really just abstract over the basic IO without additional logic but the service has the logic... I wouldn't mind refactoring the part out into Service and it would make it also easier to test (even though at the moment its not too big so that it could stay in SyncReceiverActor).

andreievg:
> It would only be for the savepoint transaction but I guess if you don't handle the error in the parent transaction block you effectively rollback the whole transaction. Does that make sense?

Just want to confirm you've tested that it rolls back the whole transaction if something fails, rather then to the nearest savepoint ?

clemens-msupply:
Thanks for the heads up, just checked this PR not even address the issue, it stills runs it in multiple tx will create a new PR...

I haven't planned to use savepoints/nested transactions...

PR #352, opened 2021-09-19
Use CentralSyncBufferRow instead of SyncRecord

clemens-msupply:


src/util/sync/mod.rs
```diff
@@ -29,7 +29,7 @@ use tokio::{
     time::{self, Duration, Interval},
 };
 
-use self::translation::{import_sync_records, SyncRecord, SyncType, TRANSLATION_RECORDS};
+use self::translation::{import_sync_records, TRANSLATION_RECORDS};
```
wlthomson, line 32:
```suggestion
use translation::{import_sync_records, TRANSLATION_RECORDS};
```

`translation` module already declared above so should be OK?

src/util/sync/translation/test_data/mod.rs
```diff
@@ -62,6 +59,14 @@ fn from_option_to_db_result<T>(option: Option<T>) -> Result<T, RepositoryError>
     }
 }
 
+#[allow(dead_code)]
+pub fn extract_sync_buffer_rows(records: &Vec<TestSyncRecord>) -> Vec<CentralSyncBufferRow> {
+    records
+        .into_iter()
+        .map(|test_record| test_record.central_sync_buffer_row.clone())
```
wlthomson, line 66: Prob can pass by value here and avoid the need for the clone? Seems like in most situations you won't need the `TestSyncRecord` objects after this?

wlthomson (approved):
LGTM.

src/util/sync/translation/test_data/mod.rs
wlthomson: Is this function going to see a lot of use? It's wrapping a pretty small expression, so unless it's being used in a fair few places I think it might be fine to just inline these?

andreievg: It's used in 3 places, if you examine the code it's used in you'll see that clone is required as 'CentralSyncBufferRows' are consumed, moved, while records need to be used after that (btw nothing wrong with cloning in tests)

andreievg (approved):

andreievg:
> LGTM... Would be nice to have an issue though. I know a minor change but think anything touching this many files should really have an associated issue. Unless the issue is going to take longer than the fix, I think it's reasonable to create one.

I've also been making multiple PR's for the same issue (although I've made checkboxes for issues that need to be done). So idk, in other project we tend to stick with one PR per Issue (and extend epics to have multiple issues), personally I'll try to add issues for each PR i do and see how much overhead that is

src/util/sync/mod.rs
clemens-msupply: thanks, I removed it

wlthomson:
> 
> 
> > LGTM... Would be nice to have an issue though. I know a minor change but think anything touching this many files should really have an associated issue. Unless the issue is going to take longer than the fix, I think it's reasonable to create one.
> 
> I've also been making multiple PR's for the same issue (although I've made checkboxes for issues that need to be done). So idk, in other project we tend to stick with one PR per Issue (and extend epics to have multiple issues), personally I'll try to add issues for each PR i do and see how much overhead that is

Ah I didn't realise it was for an existing issue. I'm not gonna force it if it's more work, but would be good to reference the issue in the title/description (ideally with a number or something if it's a multi-part PR)?

PR #353, opened 2021-09-19
Fix sync transaction

clemens-msupply:
Sync was actually spread over multiple transactions...

Not 100% happy about the current API but I tried to keep it simple, e.g. SyncSession does the transaction and is passed as an argument to the SyncRepository import method...

### DIFF
--- a/src/database/repository/diesel/sync.rs
+++ b/src/database/repository/diesel/sync.rs
@@ -9,7 +9,10 @@ use crate::database::{
     schema::{ItemRow, MasterListLineRow, MasterListNameJoinRow, MasterListRow, NameRow, StoreRow},
 };
 
+use std::future::Future;
+
 use diesel::{
+    connection::TransactionManager,
     prelude::*,
     r2d2::{ConnectionManager, Pool},
 };
@@ -35,6 +38,33 @@ pub struct SyncSession {
     tx: DBConnection,
 }
 
+impl SyncSession {
+    pub async fn transaction<'a, F, Fut>(&'a self, f: F) -> Result<(), String>
+    where
+        F: FnOnce(&'a SyncSession) -> Fut,
+        Fut: Future<Output = Result<(), String>>,
+    {
+        let transaction_manager = self.tx.transaction_manager();
+        transaction_manager
+            .begin_transaction(&self.tx)
+            .map_err(|_| "Failed to start tx".to_string())?;
+        match f(&self).await {
+            Ok(value) => {
+                transaction_manager
+                    .commit_transaction(&self.tx)
+                    .map_err(|_| "Failed to end tx".to_string())?;
+                Ok(value)
+            }
+            Err(e) => {
+                transaction_manager
+                    .rollback_transaction(&self.tx)
+                    .map_err(|_| "Failed to rollback tx".to_string())?;
+                Err(e)
+            }
+        }
+    }
+}
+
 impl SyncRepository {
     pub fn new(pool: Pool<ConnectionManager<DBBackendConnection>>) -> SyncRepository {
         SyncRepository { pool }
@@ -56,30 +86,28 @@ impl SyncRepository {
         integration_records: &IntegrationRecord,
     ) -> Result<(), RepositoryError> {
         let tx = &session.tx;
-        tx.transaction(|| {
-            for record in &integration_records.upserts {
-                match &record {
-                    IntegrationUpsertRecord::Name(record) => {
-                        NameRepository::upsert_one_tx(&tx, record)?
-                    }
-                    IntegrationUpsertRecord::Item(record) => {
-                        ItemRepository::upsert_one_tx(&tx, record)?
-                    }
-                    IntegrationUpsertRecord::Store(record) => {
-                        StoreRepository::upsert_one_tx(&tx, record)?
-                    }
-                    IntegrationUpsertRecord::MasterList(record) => {
-                        MasterListRepository::upsert_one_tx(&tx, record)?
-                    }
-                    IntegrationUpsertRecord::MasterListLine(record) => {
-                        MasterListLineRepository::upsert_one_tx(&tx, record)?
-                    }
-                    IntegrationUpsertRecord::MasterListNameJoin(record) => {
-                        MasterListNameJoinRepository::upsert_one_tx(&tx, record)?
-                    }
+        for record in &integration_records.upserts {
+            match &record {
+                IntegrationUpsertRecord::Name(record) => {
+                    NameRepository::upsert_one_tx(&tx, record)?
+                }
+                IntegrationUpsertRecord::Item(record) => {
+                    ItemRepository::upsert_one_tx(&tx, record)?
+                }
+                IntegrationUpsertRecord::Store(record) => {
+                    StoreRepository::upsert_one_tx(&tx, record)?
+                }
+                IntegrationUpsertRecord::MasterList(record) => {
+                    MasterListRepository::upsert_one_tx(&tx, record)?
+                }
+                IntegrationUpsertRecord::MasterListLine(record) => {
+                    MasterListLineRepository::upsert_one_tx(&tx, record)?
+                }
+                IntegrationUpsertRecord::MasterListNameJoin(record) => {
+                    MasterListNameJoinRepository::upsert_one_tx(&tx, record)?
                 }
             }
-            Ok(())
-        })
+        }
+        Ok(())
     }
 }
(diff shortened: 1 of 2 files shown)


src/util/sync/mod.rs
```diff
@@ -175,27 +175,35 @@ impl SyncReceiverActor {
...
+                    let records = buffer_rows
+                        .into_iter()
+                        .map(|row| SyncRecord {
+                            record_id: row.record_id,
+                            sync_type: SyncType::Insert,
+                            record_type: row.table_name,
+                            data: row.data,
+                        })
+                        .collect();
+                    import_sync_records(sync_session, repositories, &records).await?;
```
wlthomson, line 200: Passing a sync session around like this is pretty confusing! Correct me if wrong (I'm sure I am somewhere), but I think what's happening here is:

The `sync_repo` is used to start a `session`, and a transaction is then started on that `session`, with ownership of the `session` object moved to the transaction block. Within the transaction block, buffered data is fetched from the `central_sync_buffer_repository` for each table (but without using the `session`?), and the `session` is passed by reference to `import_sync_records` (a method defined in another module) which then translates the buffer records into integration records, gets a reference to the `sync_repo` again and finally passes the session created earlier by to the `integrate_record` method on the sync repo.

Putting aside the question of the final architecture until M2... if we are trying to be pragmatic, could we not make our lives and code simpler by putting all this into the `sync_repository` and doing away with all the complexities and couplings that arise from "lifting up" the repo `session` and having to pass it back and forwards between the repo and the business logic? I'm not saying we should design our repos to include translation logic as a general principle, just that I don't understand what we are gaining from avoiding it, or whether it's worth it for how much more complicated it seems to be making our code? I'm fully open to refactoring later to pull the logic up into a service if thats the way we want to go.

As agreed, going to approve but noting this here for later discussion.

src/database/repository/diesel/sync.rs
```diff
@@ -35,6 +38,33 @@ pub struct SyncSession {
     tx: DBConnection,
 }
 
+impl SyncSession {
+    pub async fn transaction<'a, F, Fut>(&'a self, f: F) -> Result<(), String>
+    where
+        F: FnOnce(&'a SyncSession) -> Fut,
+        Fut: Future<Output = Result<(), String>>,
```
wlthomson, line 45: Does the `Fut` need to be here? Can you do:

```
F: FnOnce(&'a SyncSession) -> Future<Output = Result<(), String>>,
```

wlthomson (approved):
Approving with some architectural concerns which can be revisited later.

andreievg:
I don't quite understand what the benefit we are getting from this complexity, is it possible that we are overcooking it ?

As oppose to:

```
integrate_central_records 
  for each table name
     get translated IntegrationRecords for table name subset of buffer
   end

   integrate all IntegrationRecords in a transaction (db client abstracted here)
```

Also on the issue o async vs sync, is the reason to use async because we might have async db client ? Does that mean all our methods will need to be async (in the whole code base) ? 

I am not sure what to do either approve and refactor in M2, or just leave it sitting here for a bit and refactor in this PR in the future. Either way i think it's more complex then it needs to be (or am i missing something?)

andreievg (changes requested):
just saw that Will approved, marking as change request to hold of the merge (want to see what you think about my comment)

src/util/sync/mod.rs
wlthomson: Just read @andreievg review. I may have been incorrect to think that the complexity was a result of trying to keep the translation separated? If translation can be kept out of the "session" and decoupled from the repo then that seems OK, too. Main question is whether this can be simplified and if we can avoid the repo losing ownership of its "session".

src/database/repository/diesel/sync.rs
clemens-msupply: Probably, I googled how to do it and had the same question... :)

src/util/sync/mod.rs
clemens-msupply: As said not 100% happy about the API but it shows the main idea and will probably change in the future so I didn't want to spend more time on it.

What I am trying to do here is to reuse a relatively simple "low level" SyncRepository on a higher level and not put logic into it.

With the concept of a tx/DbContext we can get rid of the whole SyncRepository which should make it much cleaner, to be discussed...

clemens-msupply:
@andreievg yes that was my initial idea but my concern is memory consumption, i.e. can I load all sync data into memory (including mobile)? Sorry if I interpreted it wrong in the last PR. This solution splits the buffer read and repo write into smaller batches (which could be split even further if needed...)

src/util/sync/mod.rs
wlthomson: Fair enough 👍.

andreievg:
> @andreievg yes that was my initial idea but my concern is memory consumption, i.e. can I load all sync data into memory (including mobile)? Sorry if I interpreted it wrong in the last PR. This solution splits the buffer read and repo write into smaller batches (which could be split even further if needed...)

I see, I did some maths before writing the discussion and suggesting transaction. I think, this sort of optimisation needs to have some benchmarking attached and some real examples of real data volumes.

Sorry I am quite strong on this as I've seen a lot of early optimisations in other project that really reduce readability of code (most of them not needed, and not backed up by benchmark or statistics of actual data volumes).

From my understanding the difference here is number of `IntegrationUpsertRecord` in memory, and i didn't think it was going to be a problem (since at the point of having too much memory usage in `IntegrationUpserRecord` you will have too much bandwidth usage for syncing those records, which will be another issue).

I did a benchmark: https://github.com/openmsupply/remote-server/compare/checking-integration-upsert-record-size
Here is a built version of the server (needs to run from within repo root)
[remote_server.zip](https://github.com/openmsupply/remote-server/files/7194888/remote_server.zip)

This endpoint allows you to set number of records you want to generate:

Screen Shot 2021-09-20 at 9 45 16 PM

### When starting:

Screen Shot 2021-09-20 at 9 39 26 PM

### 10000
Screen Shot 2021-09-20 at 9 40 49 PM

### 100000

Screen Shot 2021-09-20 at 9 42 36 PM

### 1000000

Screen Shot 2021-09-20 at 9 43 57 PM

### 10000000

Screen Shot 2021-09-20 at 9 45 04 PM

Of course the record i am keeping in memory is small, it has 3 fields, but we can do the maths for larger records. 

Anyways the point being, hardware specs are likely to be proportionate to bandwidth and internet capacity of the site. If we use a cloud server for a huge site, then internet would be fast and it's ok to sync gigabytes of data, and at this point hardware would have gigabytes of ram. And if we have a remote site with limited internet connectivity, it will likely have limited hardware capacity, and limited internet connectivity (in which case we just can't sync gigabytes off initialisation data).

Also i didn't see any indication that sqlite or postgres transaction won't be able to support this huge transaction

src/util/sync/mod.rs
andreievg: I don't think there is anything wrong with having a repository (or helper) that takes Vec<IntegrationRecord> and integrates them all using associated methods from other repositories in one transaction. To me Vec<IntegrationRecord> is a domain object, domain = sync object = data shape, repo = abstraction over DB that works with domain objects.

clemens-msupply:
thanks I created a new PR https://github.com/openmsupply/remote-server/pull/354

src/util/sync/mod.rs
wlthomson: This implies that repositories are working with domain objects at different layers, which breaks both the "aggregate root" approach that I initially wanted, as well as the service-repository pattern that @clemens-msupply has suggested (which I am also OK with). In both patterns, repos work together to form a layer of abstraction, with each repo being responsible for persisting an entity (Single Responsibility Principle in SOLID). In the aggregate root approach, this entity is domain object and not a table. In the service-repository pattern proposed by @clemens-msupply, seems like this entity is a table. If repos need to be combined, then that indicates that the repos are not being used consistently at the same abstraction layer, which means we are using repos inconsistently and not isolating our layers (i.e. we are not forming a consistent abstraction between our business logic and data access logic). 

I'm yet to see an example of repositories where the repos call each other :(. I google searched "should a repository call another repository", and these were the top results:

- https://stackoverflow.com/questions/60364118/ddd-can-a-repository-access-another-repository
- https://stackoverflow.com/questions/1364461/should-a-repository-call-another-repository-or-should-a-repository-call-a-service

The answers are varied, but all seem to be consistently answering the question with "no". However, if you have some examples from stackoverflow or blogs contradicting this, I would be open to it!

clemens-msupply: Don't think a domain object is restricted to a single table (but it could).

Yes I would agree a Repository should best not call each other (this should go into a Service). In general, the repository should also be as small as possible (number of methods). The reason is that if you want to do an alternative repository implementation, e.g. for logging, caching, etc., you only have to cover a small amount of methods...

PR #354, opened 2021-09-20
Integerate all records in one transaction

clemens-msupply:


src/util/sync/mod.rs
```diff
@@ -179,18 +180,26 @@ impl SyncReceiverActor {
...
+            let mut buffer_rows = central_sync_buffer_repository
                 .get_sync_entries(table_name)
                 .await
                 .map_err(|_| "Failed to read central sync entries".to_string())?;
-            import_sync_records(&sync_session, repositories, &buffer_rows).await?;
+            records.append(&mut buffer_rows);
         }
+
+        // write all entries to the DB
+        let sync_session = repositories
```
wlthomson, line 195: Nice, already a lot easier to follow :). I have a suggestion which I think might serve to make this a bit cleaner and still keep translation logic out of the repo:

```
Initialise integration_records list
For each table
  Get buffer rows
  Translate buffer rows into integration records
  Append integration records to integration records list
End for each

// note that no mutations have occurred yet, hence no session yet

sync_repo.integrate_records(integration_records) // all mutations done atomically with a transaction inside `sync_repo`
```

where `integrate_records` does the same as what it does now, but also removes all the buffer records when done (I would also get the buffer rows via `sync_repo` not `central_sync_buffer_repo` (this is moving in the direction of a later refactor of `sync_repo` to `sync_service`).

Note that this removes the need for the sync actor to handle "sessions" at all, and leaves that to the `sync_repo` (again, potentially later to be the `sync_service`).

@clemens-msupply Do you see what I'm going for with this? Is there anything I've missed that would make this approach not work?

wlthomson (commented):
Some "general feedback without explicit approval" 😃. Will approve on resolving the discussion.

image

src/util/sync/mod.rs
clemens-msupply: sorry my bad the sync_session was already outdated...

central_sync_service -> yes but maybe later?

clemens-msupply: Removed in df71bd3

wlthomson (approved):

andreievg (approved):
thank you

Issue #356, opened 2021-09-21
Add transaction and transaction line translations


andreievg:
Part of: https://github.com/openmsupply/remote-server/issues/349
Change schema to invoice and invoice_line

Issue #294, opened 2021-09-02
Rename transact tables


wlthomson:
Looking at the spreadsheet, it looks like we've made a decision to go with `transaction` over `transact`. Not much of a hassle to rename now, so may as well get it out of the way sooner rather than later?

PR #357, opened 2021-09-21
356 invoice schema changers

andreievg:
Fixes: #356 
Fixes: #294 

### Changes

* Change from transact to invoice (with all translations needed in M1)
* Change from transact_line to invoice_line (with all translations need in M1)

Not happy with type field, we were hoping there wouldn't be anything restricting us using field we want (as there is always mechanism that allows us to escape keywords): https://github.com/openmsupply/org-issues/discussions/36
But diesel doesn't account for this, so it's less pleasant syntax then what I hoped

We can branch from this branch for mutation and invoice query work.

I didn't use `git mv`, as it's already huge amount of work to just rename. (rest assured I didn't change any logic)


andreievg:
@wlthomson, this one plays into your position on keywords (I was naive to expect all rust frameworks to respect rust keyword rules).

Basically what we have now is:

in schema definition:
```rust
  #[sql_name = "type"]
     type_ -> Invoice_type,
```
in data struct
```rust
 #[column_name = "type_"]
    pub r#type: InvoiceRowType,
```
and when using dsl
```rust
    let result = invoice
            .filter(
                type_
                    .eq(InvoiceRowType::CustomerInvoice)
                    .and(store_id.eq(store)),
            )
```

That's still not enough for me to budge on my position, but enough to acknowledge that I should have thought beyond rust language itself and considered frameworks we'll be using. (btw I did check async-graphql and serde which were fine with #r in structure they auto derive).

Anyways if Clemens holds your view, let's change.

migrations/postgres/00000107_create_invoice_table/up.sql
```diff
@@ -0,0 +1,27 @@
...
+    name_id TEXT NOT NULL REFERENCES name(id),
+    -- For customer invoices, the id of the issuing store.
+    -- For supplier invoices, the id of the receiving store.
+    store_id TEXT NOT NULL REFERENCES store (id),
+    invoice_number INTEGER NOT NULL,
+    type invoice_type NOT NULL,
+    status invoice_status NOT NULL,
+    comment TEXT,
+    their_reference TEXT,
+    entry_datetime TEXT NOT NULL,
```
clemens-msupply, line 25: Think we can use the Postgres date time here and Diesel should transparently do the right thing for sqlite (I can try that out for the invoice query).

andreievg: Do you want me to change it here ? Or as part of your check ? 
Would be great to know, since it also affects the mutation logic

src/server/service/graphql/schema/types/mod.rs
```diff
@@ -186,131 +180,105 @@ impl RequisitionLine {
...
     }
 
     pub async fn invoice_number(&self) -> i32 {
-        self.transact_row.invoice_number
+        self.invoice_row.invoice_number
     }
 
-    pub async fn type_of(&self) -> TransactType {
-        self.transact_row.type_of.clone().into()
+    pub async fn r#type(&self) -> InvoiceType {
```
clemens-msupply, line 221: Regarding your keyword comment, I don't really mind either style, just a bit of feedback: When I first saw this syntax a couple of days ago I found it very odd (and that you can do it for function names) but I kind of guessed what it does...

clemens-msupply (approved):
Looks good to me (mostly skimmed through it)

migrations/postgres/00000107_create_invoice_table/up.sql
clemens-msupply: We can change it later if it works. Will try it out now...

wlthomson (approved):
Nice, think this covers my question here https://github.com/openmsupply/remote-server/issues/294#issuecomment-917377812 (and closes the issue)?

Re. `type`, it seems like using rust keywords incurs additional costs than just SQL keywords (in terms of code clarity and conciseness/niceness? I'm assuming that this problem doesn't extend to SQL keywords, so I don't think there's any reason to change the decision to use and escape those? It's your call on rust keywords, up to you and whether you think the benefits of using the word "type" outweigh the additional costs 🤷‍♂️ (there aren't many rust keywords, so I think it's **probably** going to be a decision limited to `type`).

If you decide to avoid rust keywords, you could rename the field with a redundant table specifier e.g. `invoice_type`. If you want to keep them, it still might be a good idea to use something like that in place of where you currently have `type_` (just to make it a bit clearer, the underscore is easy to miss)? Whatever decision you go with, would be good to have a comment somewhere (if you change the field, maybe on the migration, if you don't, maybe on the diesel schema mapping)?

andreievg:
> Re. type, it seems like using rust keywords incurs additional costs than just SQL keywords (in terms of code clarity and conciseness/niceness? I'm assuming that this problem doesn't extend to SQL keywords, so I don't think there's any reason to change the decision to use and escape those? It's your call on rust keywords, up to you and whether you think the benefits of using the word "type" outweigh the additional costs 🤷‍♂️ (there aren't many rust keywords, so I think it's probably going to be a decision limited to type).

> If you decide to avoid rust keywords, you could rename the field with a redundant table specifier e.g. invoice_type. If you want to keep them, it still might be a good idea to use something like that in place of where you currently have type_ (just to make it a bit clearer, the underscore is easy to miss)? Whatever decision you go with, would be good to have a comment somewhere (if you change the field, maybe on the migration, if you don't, maybe on the diesel schema mapping)?

Cool I'll mark this as a point to revisit first thing after M1

andreievg:
> Nice, think this covers my question here #294 (comment) (and closes the issue)?

Added to the main issue comment

PR #358, opened 2021-09-21
356 invoice datetime

clemens-msupply:
Tested for pg and sqlite


andreievg (approved):
One thing to be aware of, is in API we need to add UTC timezone

PR #359, opened 2021-09-22
api error handling example

andreievg:
After and idea from @joshxg  (that somewhat extends on the idea form @wlthomson [of including error in query](https://github.com/openmsupply/org-issues/discussions/25#discussioncomment-1257711), and [this article](https://blog.logrocket.com/handling-graphql-errors-like-a-champ-with-unions-and-interfaces/)) and some research into how unions work and interfaces work in async graphql, i think we have a pretty robust solution, shown in this PR.

`... on ` graphQL syntax is type narrowing of sorts. 

To make sure front end gets an error alway, an interface is used `Error`, you can see in below example and output that OffsetError is actually missing from in the third query, but it's still present in response (not fully but it's there). 

This also allows for type introspection to be translate into generated types in front end (otherwise errors would have to be either self types or another root query with all of the error)

```graphql
query names {
  # Just one offset error
  error1: names(page: { offset: -1 }) {
    ... on NameList {
      totalCount
      nodes {
        id
        name
      }
    }
    ... on NameListError {
      errors {
        ... on Error {
          __typename
          description
        }
        ... on FirstError {
          first
          minFirst
          maxFirst
        }
        ... on OffsetError {
          offset
        }
      }
    }
  }
  # Both offset and first error
  error2: names(page: { first: 300, offset: -1 }) {
    ... on NameList {
      totalCount
      nodes {
        id
        name
      }
    }
    ... on NameListError {
      errors {
        ... on Error {
          __typename
          description
        }
        ... on FirstError {
          first
          minFirst
          maxFirst
        }
        ... on OffsetError {
          offset
        }
      }
    }
  }
  # Both errors but offset error not specified in query
  error3: names(page: { first: 300, offset: -1 }) {
    ... on NameList {
      totalCount
      nodes {
        id
        name
      }
    }
    ... on NameListError {
      errors {
        ... on Error {
          __typename
          description
        }
        ... on FirstError {
          first
          minFirst
          maxFirst
        }
      }
    }
  }
}
```
result

```JSON
{
  "data": {
    "error1": {
      "errors": [
        {
          "__typename": "OffsetError",
          "description": "Offset must not be negative",
          "offset": -1
        }
      ]
    },
    "error2": {
      "errors": [
        {
          "__typename": "OffsetError",
          "description": "Offset must not be negative",
          "offset": -1
        },
        {
          "__typename": "FirstError",
          "description": "First is too large",
          "first": 300,
          "maxFirst": 200,
          "minFirst": 1
        }
      ]
    },
    "error3": {
      "errors": [
        {
          "__typename": "OffsetError",
          "description": "Offset must not be negative"
        },
        {
          "__typename": "FirstError",
          "description": "First is too large",
          "first": 300,
          "maxFirst": 200,
          "minFirst": 1
        }
      ]
    }
  }
}
```

### DIFF
--- a/src/database/repository/diesel/mod.rs
+++ b/src/database/repository/diesel/mod.rs
@@ -37,7 +37,7 @@ pub use requisition::RequisitionRepository;
 pub use requisition_line::RequisitionLineRepository;
 pub use stock_line::StockLineRepository;
 pub use store::StoreRepository;
-pub use sync::{IntegrationRecord, IntegrationUpsertRecord, SyncRepository, SyncSession};
+pub use sync::{IntegrationRecord, IntegrationUpsertRecord, SyncRepository};
 pub use user_account::UserAccountRepository;
 
 use diesel::{
--- a/src/main.rs
+++ b/src/main.rs
@@ -72,9 +72,6 @@ async fn main() -> std::io::Result<()> {
     tokio::select! {
         result = http_server => result,
         () = async {
-          sync_sender.schedule_send(Duration::from_secs(settings.sync.interval)).await;
-        } => unreachable!("Sync receiver unexpectedly died!?"),
-        () = async {
           sync_receiver.listen(repository_registry_sync_data).await;
         } => unreachable!("Sync scheduler unexpectedly died!?"),
     }
--- a/src/server/service/graphql/schema/queries/mod.rs
+++ b/src/server/service/graphql/schema/queries/mod.rs
@@ -7,8 +7,8 @@ use crate::database::schema::{InvoiceLineRow, InvoiceRow, RequisitionRow, StoreR
 use crate::server::service::graphql::schema::types::{Invoice, InvoiceLine, Requisition, Store};
 use crate::server::service::graphql::ContextExt;
 
-use super::types::{ItemList, NameList};
-use async_graphql::{Context, Object};
+use super::types::{ItemList, NameListWithError};
+use async_graphql::*;
 use pagination::Pagination;
 pub struct Queries;
 
@@ -23,8 +23,8 @@ impl Queries {
         &self,
         _ctx: &Context<'_>,
         #[graphql(desc = "pagination (first and offset)")] page: Option<Pagination>,
-    ) -> NameList {
-        NameList { pagination: page }
+    ) -> NameListWithError {
+        NameListWithError::new(page)
     }
 
     pub async fn items(
--- a/src/server/service/graphql/schema/queries/pagination.rs
+++ b/src/server/service/graphql/schema/queries/pagination.rs
@@ -3,7 +3,7 @@ use async_graphql::InputObject;
 #[derive(InputObject)]
 pub struct Pagination {
     pub first: Option<u32>,
-    pub offset: Option<u32>,
+    pub offset: Option<i32>,
 }
 
 pub trait PaginationOption {
@@ -12,6 +12,7 @@ pub trait PaginationOption {
 }
 
 pub const DEFAULT_PAGE_SIZE: u32 = 100;
+pub const MAX_PAGE_SIZE: u32 = 200;
 
 impl PaginationOption for Option<Pagination> {
     fn first(&self) -> i64 {
--- a/src/server/service/graphql/schema/types/name.rs
+++ b/src/server/service/graphql/schema/types/name.rs
@@ -61,11 +61,11 @@ pub enum Error {
     OffsetError(OffsetError),
 }
 
-pub struct NameListError(pub Error);
+pub struct NameListError(pub Vec<Error>);
 
 #[Object]
 impl NameListError {
-    async fn error(&self) -> &Error {
+    async fn errors(&self) -> &Vec<Error> {
         &self.0
     }
 }
@@ -79,16 +79,22 @@ pub enum NameListWithError {
 impl NameListWithError {
     pub fn new(pagination: Option<Pagination>) -> NameListWithError {
         match NameListWithError::check_error(&pagination) {
-            Some(error) => NameListWithError::NameListError(NameListError(error)),
+            Some(errors) => NameListWithError::NameListError(NameListError(errors)),
             None => NameListWithError::NameList(NameList { pagination }),
         }
     }
 
-    pub fn check_error(pagination: &impl PaginationOption) -> Option<Error> {
+    pub fn check_error(pagination: &impl PaginationOption) -> Option<Vec<Error>> {
+        let mut result = Vec::new();
         if pagination.offset() < 0 {
-            Some(Error::OffsetError(OffsetError(pagination.offset())))
-        } else if pagination.first() < 1 || pagination.first() > MAX_PAGE_SIZE.into() {
-            Some(Error::FirstError(FirstError(pagination.first())))
+            result.push(Error::OffsetError(OffsetError(pagination.offset())))
+        }
+        if pagination.first() < 1 || pagination.first() > MAX_PAGE_SIZE.into() {
+            result.push(Error::FirstError(FirstError(pagination.first())))
+        }
+
+        if result.len() > 0 {
+            Some(result)
         } else {
             None
         }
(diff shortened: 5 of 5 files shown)


andreievg:
btw this is just to show the semantics of error unions and interfaces, i think could do better job with actually determining error etc..

joshxg:
Thats awesome! The introspection and codegen for error types, typesafety and explicit handling of errors is all really positive thanks @wlthomson and @andreievg !

And for clarity - any errors to do with graphql errors (type checking, parsing) and server errors (yeah, right!) will still use the regular `errors` field and a gql error shape?

Two things i'm still thinking through:
1. Auth and permission errors: I think permission errors will be the same as what you have. Just a new `PermissionError` interface, or something. Auth errors - do they come under the 'regular gql' or this new error system? (Maybe we need some vocab to differentiate? 😁 )
2. Mutations - I have some brain storming below

- Maybe we put the `id` at the same level of the error/s 🤔 ? So `{ id, errors: ... }` (better example below?). 
- With this, I'm wondering about when an invoice and lines have errors in a mutation. I'm thinking that there's only the errors for the invoice and not the lines without going more complicated. However, I'm unsure if this is really much of a problem, practically. Errors on the invoice itself should be rare enough - we have validation client side, so realistically error returns on an invoice will either be because of conflicts (someone deleted it, oops!) or a bug on the frontend. In both cases, just seeing the invoice errors, and not the line errors until you save a second time doesn't seem too bad to me (again, rare).
- I have some examples of what I think this all looks like for mutations below.. please point out my foolishness where I am wrong!

```
mutation updateInvoice($updateInvoicePatch: UpdateInvoicePatch) {
  updateInvoice(updateInvoicePatch: UpdateInvoicePatch) {
    ... on InvoiceUpdatePatch {
      id
      status
      etc
    }
    ... on InvoiceUpdateError {
      errors {
        ... on Error {
          __typename
          id
          description
        }
        ... on ForeignKeyError {
          __typename
          id
          code
          field
        }
      }
    }
  }
}
```

Because it would be very useful to have a query like this for a mutation:

```
mutation updateInvoice($updateInvoicePatch: UpdateInvoicePatch) {
  updateInvoice(updateInvoicePatch: UpdateInvoicePatch) {
    ... on UpdateInvoicePatch {
      id
      status

      lines {
        ... on UpdateInvoiceLinePatch {
          id
          quantity
        }

        ... on UpdateInvoiceLineError {
          id
          error {
            ... on QuantityError {
              code
              quantity
            }

            ... on OtherError {
              code
            }
          }
        }
      }
    }
    ... on UpdateInvoiceError {
      id
      errors {
        ... on Error {
          __typename
          description
        }
        ... on ForeignKeyError {
          __typename
          code
          field
        }
      }
    }
  }
}

```

andreievg:
> Auth and permission errors: I think permission errors will be the same as what you have. Just a new PermissionError interface, or something. Auth errors - do they come under the 'regular gql' or this new error system? (Maybe we need some vocab to differentiate?

I think consistency is quite important. Every error that we explicitly handle, including auth errors have the same mechanism ? 

> Maybe we put the id at the same level of the error/s 🤔 ? So { id, errors: ... } (better example below?).

Sure, if that helps consumer.

> With this, I'm wondering about when an invoice and lines have errors in a mutation. I'm thinking that there's only the errors for the invoice and not the lines without going more complicated. However, I'm unsure if this is really much of a problem, practically. Errors on the invoice itself should be rare enough - we have validation client side, so realistically error returns on an invoice will either be because of conflicts (someone deleted it, oops!) or a bug on the frontend. In both cases, just seeing the invoice errors, and not the line errors until you save a second time doesn't seem too bad to me (again, rare).

Hmm, I think the complication arrises from sending the full shape of input back and merging errors where they fit in that shape ? I think evaluating errors and sending back only errors for the same level works even if there is one shape for both types of errors (just leave lines empty if there is an error with actual response, and leave top level 'errors' empty if there is an error with lines ?). Happy with either way although i kind of like the vary last shape below (but it means putting id inside error, but it means you just look in one place for errors). 

Just FYI InputObject can't be the same as Output Object in async graphql

And one more note, I think it's important to differentiate between fully successful update and failed update, especially since you have to dig deep to find if there was an error.

Typical GraphQL mutation return the type it's mutating, so how does this sound (this also helps with knowing straight away if there is an error or not):

```graphql
mutation updateInvoice($updateInvoicePatch: UpdateInvoicePatch) {
  updateInvoice(updateInvoicePatch: $updateInvoicePatch) {
    ... on Invoice {
      ...InvoiceFragment
    }
    ... on UpdateInvoiceError {
      id
      errors {
        # Interface
        ... on InvoiceError {
          __typename
          description
        }
        # Interface Impl
        ... on ForeignKeyError {
          field
        }
      }
      lines {
        id
        errors {
          ... on InvoiceLineError {
            __typename
            description
          }
          # Interface Impl
          ... on QuantityError {
            quantity
          }
          # Interface Impl
          ... on ForeignKeyError {
            field
          }
        }
      }
    }
  }
}
```

Do we really need the full info of input/patch for mutation in the error, would id be sufficient ? 

Since we have __typename, does code become description ?

Personal favourite (but happy to go with whatever)
```graphql
mutation updateInvoice($updateInvoicePatch: UpdateInvoicePatch) {
  updateInvoice(updateInvoicePatch: $updateInvoicePatch) {
    ... on Invoice {
      ...InvoiceFragment
    }
    ... on UpdateInvoiceError {
      errors {
        # Interface
        ... on InvoiceError {
          id
          __typename
          description
        }
        # Interface Impl
        ... on ForeignKeyError {
          fieldId
          field
        }
      }
    }
    ... on UpdateInvoiceLineError {
      errors {
        # Interface
        ... on InvoiceLineError {
          id
          __typename
          description
        }
        # Interface Impl
        ... on ForeignKeyError {
          fieldId
          field
        }
        # Interface Impl
        ... on QuantiyError {
          quantity
        }
      }
    }
  }
}
```

joshxg:
> I think consistency is quite important. Every error that we explicitly handle, including auth errors have the same mechanism ?

Hmm OK that makes sense I am just wondering about the response from an auth error for, say, an invoice mutation

-------

Shape:

I'm not understanding this part, but we can work off the concrete examples, I think.

> I think evaluating errors and sending back only errors for the same level works even if there is one shape for both types of errors (just leave lines empty if there is an error with actual response, and leave top level 'errors' empty if there is an error with lines ?).

---------

I really like your first shape and it would be great for the client if it's not too difficult.If we just went with this, I would be so very happy!

> Do we really need the full info of input/patch for mutation in the error, would id be sufficient ?

Just the ID is sufficient I think. You seem to have everything needed in the error so far (i.e. for batch reduction) [The only case where I think *maybe* we might need a lot more info, I think, would be if for example, we showed a UI which popped up and said "choose which one you want".. kind of like a git diff - but this is just a random idea and can be sorted out later]

Just to make sure, I want to ask about your second shape:

You have an `UpdateInvoiceError ` and `UpdateInvoiceLineError` - but in the example above, there can be an `errors` at both the `Invoice` level and the `lines` level. So this makes me curious about the response if there is both an invoice error (say, the other party ID is invalid) as well as an invoice line error (say, the quantity is invalid) in your second example?

> Since we have __typename, does code become description ?

I have a bit of a crazy idea? (another one?!)

What if we have a `description` or `message` which is a message that the user is able to actually view in the UI?

1. These won't always need to be used, but when introducing new errors, they'll be automatically handled and shown in the UI.
2. We will have localisation JSON files on the server, already..
3. Could save bundle size
4. When we add new errors, don't need to always update the frontend, with this system (The record will have the error attached, with at least a message..) (this is 1 isn't it?!)
5. Seems by using the `DateTime` in the API we need to start a session with the client with their timezone and serve the client `DateTime`s in their timezone, so we can also keep their desired language in the session?
6. The API is heading towards quite a deeply coupled to the client, so why not? :)

andreievg:
> I really like your first shape and it would be great for the client if it's not too difficult.If we just went with this, I would be so very happy!

Sold, when doing work on supplier invoice it felt 'nicer' then the second option

andreievg:
doing PR clean up, so closing, will know it's here for reference

Issue #307, opened 2021-09-06
[API] - Invoices


andreievg:
MW parent job: `[OMSBEA] omSupply Back End - API`

### Description

Invoices query (as per docs)

`note` to be done after `invoice`, `sorting`, `filtering` and `pagination`

### Scope

<ins>Includes</ins>
* Add query for all fields in invoices
* Add filtering sorting and pagination from existing mechanisms

<ins>Excludes</ins>
* Related table filtering
* Multi sort

### Tasks

- [ ] Break down tasks into digestible (by reviewer) units.
- [ ] MW job added

clemens-msupply:
Filtering and sorting still missing...

PR #360, opened 2021-09-23
307 invoices query

clemens-msupply:
First part of the invoices query (query + pagination). Sorry I swapped tickets halfway and saw to late hat #307 should be done later.

Test is very basic/stupid, but wanted to figure out how to verify the graphql results in a better way first... Test can probably also do with some utils method to make the setup a bit easier


src/database/repository/diesel/invoice_query.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub type InvoiceQueryJoin = (InvoiceRow, Option<NameRow>, Option<StoreRow>);
+
+impl InvoiceQueryRepository {
+    pub fn new(pool: Pool<ConnectionManager<DBBackendConnection>>) -> Self {
+        InvoiceQueryRepository { pool }
+    }
+
+    /// Calculates invoice line stats for a given invoice
+    pub async fn stats(
```
wlthomson, line 33: Should this maybe be `get_invoice_line_stats`? Something which indicates that the method is a getter and the stats are for the lines? Maybe this is obvious, but are there any drawbacks to making the method identifier a bit more specific?

src/database/repository/diesel/invoice_query.rs
```diff
@@ -0,0 +1,60 @@
...
+        &self,
+        invoice_ids: &[String],
+    ) -> Result<Vec<InvoiceLineStatsRow>, RepositoryError> {
+        use crate::database::schema::diesel_schema::invoice_line_stats::dsl as invoice_line_stats_dsl;
+        let connection = get_connection(&self.pool)?;
+
+        Ok(invoice_line_stats_dsl::invoice_line_stats
```
wlthomson, line 40: Could maybe include this qualifier in the import? Something like:

```
use crate::database::schema::diesel_schema::invoice_line_stats::dsl::invoice_line_stats;
```

A bit of a weirdly long import... but probably still a bit cleaner?

src/database/repository/diesel/invoice_query.rs
```diff
@@ -0,0 +1,60 @@
+use crate::{
+    database::{
+        repository::RepositoryError,
+        schema::{
+            diesel_schema::{
+                invoice::dsl as invoice_dsl, name_table::dsl as name_dsl, store::dsl as store_dsl,
```
wlthomson, line 6: Is there a reason you've lifted these imports up here? I don't see a problem with it, but it seems to be breaking the pattern you established in the other diesel repos?

src/database/schema/invoice_line.rs
```diff
@@ -12,6 +13,16 @@ pub struct InvoiceLineRow {
     pub pack_size: i32,
     pub cost_price_per_pack: f64,
     pub sell_price_per_pack: f64,
+    pub total_after_tax: f64,
     pub available_number_of_packs: i32,
     pub total_number_of_packs: i32,
 }
+
+/// Row for the invoice_line_stats VIEW table (needed because of a Diesel limitation to query
+/// aggregates and columns at the same time).
```
wlthomson, line 22: Would be helpful maybe to include a link to an issue/doc explaining this limitation?

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+use chrono::{DateTime, Utc};
+
+#[derive(SimpleObject, PartialEq, Debug)]
+pub struct InvoicesPricing {
+    #[graphql(name = "totalAfterTax")]
+    total_after_tax: f64,
+}
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug)]
+pub enum GraphQLInvoiceType {
```
wlthomson, line 20: Is there a need for the `GraphQL` specifier here, given that these types are all behind the `graphql` module?

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+    #[graphql(name = "confirmDatetime")]
+    confirm_datetime: Option<String>,
+    #[graphql(name = "finalisedDatetime")]
+    finalised_datetime: Option<String>,
+}
+
+#[ComplexObject]
+impl InvoicesNode {
+    async fn pricing(&self, ctx: &Context<'_>) -> InvoicesPricing {
+        let loader = ctx.get_loader::<DataLoader<InvoiceLineStatsLoader>>();
```
wlthomson, line 84: Since you have the parent invoice, could you not do a query here to get all the lines for that invoice and then sum their `total_after_tax`? The query would be a single SQL query, and the summing should be quite cheap to do? Could just expose this as a method on the `invoice` repository? 

Have I missed something?

wlthomson (changes requested):
Trying to not start any design discussions here, so attempting to avoid any comments re. general patterns that have already been established in previous PRs! 

Main question is about the need for the view etc. and whether the calculation could be done using existing repositories... good chance I am missing some context!

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        let result = loader
+            .load_one(self.id.to_string())
+            .await
+            .ok()
+            .flatten()
+            .map_or(
+                InvoiceLineStatsRow {
```
andreievg, line 92: I think it's neater to implement `from <Option<InvoiceLineStatsRow>> for InvoicePricing` with default to zero.

then it's 
```rust
let result = loader
            .load_one(self.id.to_string())
            .await.unwrap().into()
```

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+        InvoicesNode {
+            id: invoice_row.id,
+            other_party_name,
+            other_party_id,
+            status: GraphQLInvoiceStatus::from(invoice_row.status),
+            invoice_type: GraphQLInvoiceType::from(invoice_row.r#type),
+            invoice_number: invoice_row.invoice_number,
+            their_reference: invoice_row.their_reference,
+            comment: invoice_row.comment,
+            entry_datetime: DateTime::<Utc>::from_utc(invoice_row.entry_datetime, Utc).to_rfc3339(),
```
andreievg, line 120: I think use `from`, and also for the option below

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+        InvoicesPricing {
+            total_after_tax: result.total_after_tax,
+        }
+    }
+}
+
+impl From<InvoiceQueryJoin> for InvoicesNode {
+    fn from((invoice_row, name_row, _store_row): InvoiceQueryJoin) -> Self {
+        // TODO return error if name is not present (None)?
+        let (other_party_id, other_party_name) =
```
andreievg, line 108: i think it should be inner_join rather then left_join (since both store_id and name_id are not optional on invoice), then no need for Options in InvoiceQueryJoin

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+}
+
+pub struct InvoicesList {
+    pub pagination: Option<Pagination>,
+}
+
+#[Object]
+impl InvoicesList {
```
andreievg, line 136: lists have totalCount and  nodes

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+                |v| v,
+            );
+
+        InvoicesPricing {
+            total_after_tax: result.total_after_tax,
+        }
+    }
+}
+
+impl From<InvoiceQueryJoin> for InvoicesNode {
```
andreievg, line 105: hmm, i wonder why repository shouldn't just return the domain object (being InvoiceNode), it sure saves a lot of repetition.

Is it because of the mapping of status, types and dates ?

src/database/repository/diesel/invoice_query.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    /// Gets all invoices
+    pub fn all(
+        &self,
+        pagination: &Option<Pagination>,
+    ) -> Result<Vec<InvoiceQueryJoin>, RepositoryError> {
+        let connection = get_connection(&self.pool)?;
+
+        Ok(invoice_dsl::invoice
+            .left_join(name_dsl::name_table)
```
andreievg, line 53: i think inner_join for both

src/database/loader/invoice_line.rs
```diff
@@ -29,3 +29,23 @@ impl Loader<String> for InvoiceLineLoader {
...
+#[async_trait::async_trait]
+impl Loader<String> for InvoiceLineStatsLoader {
+    type Value = InvoiceLineStatsRow;
+    type Error = RepositoryError;
+
+    async fn load(&self, keys: &[String]) -> Result<HashMap<String, Self::Value>, Self::Error> {
+        Ok(self
+            .invoice_query_repository
+            .stats(keys)
+            .await?
```
andreievg, line 46: Can this really be non blocking ? I never checked, i thought it has to be. Going to do some test on this, if it can be non blocking then there was no point for me doing block and move in stock line repository

andreievg (changes requested):

andreievg:
@wlthomson 
> Trying to not start any design discussions here, so attempting to avoid any comments re. general patterns that have already been established in previous PRs!

I also agree, a little bit of re-inventing

src/database/repository/diesel/invoice_query.rs
clemens-msupply: The problem is there are imports like `id` or `item` that often clash with function arguments. Found having an *_dsl much cleaner especially if you have multiple dsl.

src/database/repository/diesel/invoice_query.rs
clemens-msupply: No reason for it (probably rust-analyser). In this case I use the dsls a couple of times so having them at the top is probably better.

src/database/schema/invoice_line.rs
clemens-msupply: If I could have found and issue/docs :) only got a type error that suggested that it's not possible. I still hope that Diesel can do it somehow.

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: Was just to have a clearer name InvoiceType vs InvoiceRowStatus. Renamed in 96042bb.

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: Yes I think I can just load all the lines, that would be easier and I don't need the View, thanks. Think I spend too much time with this Diesel problem to solve it with a single query :)...

src/database/loader/invoice_line.rs
```diff
@@ -29,3 +29,23 @@ impl Loader<String> for InvoiceLineLoader {
...
+pub struct InvoiceLineStatsLoader {
+    pub invoice_query_repository: InvoiceQueryRepository,
+}
+
+#[async_trait::async_trait]
+impl Loader<String> for InvoiceLineStatsLoader {
+    type Value = InvoiceLineStatsRow;
+    type Error = RepositoryError;
+
+    async fn load(&self, keys: &[String]) -> Result<HashMap<String, Self::Value>, Self::Error> {
```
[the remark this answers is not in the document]
clemens-msupply: Think if I load all lines to get the sum I can avoid the extra View and this whole loader...

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: is that graphql? or our convention?

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+}
+
+#[derive(SimpleObject, PartialEq, Debug)]
+#[graphql(complex)]
+#[graphql(name = "Invoice")]
+pub struct InvoicesNode {
+    id: String,
+    #[graphql(name = "otherPartyName")]
```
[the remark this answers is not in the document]
clemens-msupply: thanks done in 62fa927

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+            InvoiceRowType::CustomerInvoice => GraphQLInvoiceType::CustomerInvoice,
+            InvoiceRowType::SupplierInvoice => GraphQLInvoiceType::SupplierInvoice,
+        }
+    }
+}
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug)]
+pub enum GraphQLInvoiceStatus {
+    #[graphql(name = "DRAFT")]
+    Draft,
```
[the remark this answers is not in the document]
clemens-msupply: thanks 62fa927

src/database/loader/invoice_line.rs
andreievg: i've checked, this work on problems, i mean only one query fired up 👍

clemens-msupply: Its only because all the other repo methods are async, no technical reason...

src/server/service/graphql/schema/types/invoices.rs
```diff
@@ -0,0 +1,146 @@
...
+            InvoiceRowStatus::Confirmed => GraphQLInvoiceStatus::Confirmed,
+            InvoiceRowStatus::Finalised => GraphQLInvoiceStatus::Finalised,
+        }
+    }
+}
+
+#[derive(SimpleObject, PartialEq, Debug)]
+#[graphql(complex)]
+#[graphql(name = "Invoice")]
+pub struct InvoicesNode {
```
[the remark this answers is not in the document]
clemens-msupply: done 89391ba Thought InvoiceNode might clash with the invoice query but this will not be the case...

src/database/repository/diesel/invoice_query.rs
wlthomson: I think the motivation for keeping them in the methods is just good SOLID in terms of tables only being imported in methods where they are used. Agree that it's not a big deal.

clemens-msupply: I think it just comes from the Diesel examples where they include the whole dsl. If you would do that at the top you suddenly have variables like `id` defined in the whole file... But as said I found including the whole dsl very annoying because it often clashes with local variables.

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: thanks done in 31d284b

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: One reason is that I don't wanted to put GraphQL references in the lower repository layer. I might be wrong but the InvoiceNode object seems very GraphQL specific, having a more general "domain object" is more powerful because it can be used for other purposes as well.

What do you mean by repetition? I think I only need to transform it once don't I? Its really just keeping the layers separated otherwise its the same, or do I miss something?

src/database/repository/diesel/invoice_query.rs
clemens-msupply: done 31d284b

src/server/service/graphql/schema/types/invoices.rs
andreievg: We are moving from sql at this point. And reinventing a solid (and very fast mechanism). We would have to group them manually too. I prefer we use the 'power' of database engine (it's heavily optimised), rather than using it just as a data store.
This query is a list query, if we are grabbing x number of invoices, which typically have large number of lines. To be conservative, say it's 100 lines on each invoice and pagination limit is 1000, that means we will get 100000 lines that we have group and sum (we not even using indexes at this point, database engines used indexes for grouping)

src/database/loader/invoice_line.rs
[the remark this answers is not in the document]
andreievg: I strongly disagree, i've made a comment

src/server/service/graphql/schema/types/invoices.rs
andreievg: it's a common convention for pagination in grahpql (and what we've adopted, described in our api docs and implemented for names and items

src/server/service/graphql/schema/types/invoices.rs
andreievg: > One reason is that I don't wanted to put GraphQL references in the lower repository layer. 

Does this mean repository deals with datatype and shapes of database only ? I thought they deal with domain objects. 

> I might be wrong but the InvoiceNode object seems very GraphQL specific, having a more general "domain object" is more powerful because it can be used for other purposes as well.

I wanted to use InvoiceNode in business logic to validate InputObject of mutation against it. 
Just wondering, once an object auto derives some traits of GraphQL does it automatically become GraphQL object ?

> What do you mean by repetition? I think I only need to transform it once don't I? Its really just keeping the layers separated otherwise its the same, or do I miss something?

I thought sharing object between layers is common (rather than translating object at every layer boundary). 

Lastly it feel weird to me that we are getting database datatype object in GraphQL and translating it there, rather then repository giving us the object that we need.

Looking forward to getting clarity on all of this next week

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: Thanks I had the same concern this morning and was unsure about how much data we are talking about... So I leave the VIEW in (as long as there isn't a better way to do it in Diesel).

src/database/loader/invoice_line.rs
```diff
@@ -29,3 +29,23 @@ impl Loader<String> for InvoiceLineLoader {
...
+        Ok(self
+            .invoice_query_repository
+            .stats(keys)
+            .await?
+            .iter()
+            .map(|stats: &InvoiceLineStatsRow| (stats.invoice_id.clone(), stats.clone()))
```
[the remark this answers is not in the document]
clemens-msupply: done 0911440

src/database/loader/invoice_line.rs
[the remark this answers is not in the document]
clemens-msupply: done 0911440

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: thanks done in 6bec9e6

andreievg (approved):
thanks for changes

src/server/service/graphql/schema/types/invoices.rs
wlthomson: Sorry, I'm a bit list now!

We can still have a single SQL query that groups and sums all the lines for a given invoice, all that is fine... I still don't understand why we need a view and a new loader etc., can't we just do this with diesel in a method on the invoice repo (if we really want, we can execute raw SQL in the repo method) and avoid all this extra stuff?

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: I think how data types are shared should be in one direction, from the lower level up to the higher level. Ideally the lower level code shouldn't depend on definitions from the higher level, e.g. because there could be different APIs (graphql, rest,...) using the same low level code.

I just looked at the name_query repo and there is a circular module dependency, i.e. NameQuery is imported into the repository. To solve this we could move NameQuery into the repository module but then there are still graphql macros in respository (which is weird to me). So either way their is a bit of a circular dependency...

Yes agree having tuples of database row types as domain objects is a bit strange but the good think about it is that we don't have to transform them into an intermediate domain objects. For this reason I think its a good compromise while having a very flexible multi purpose domain object. I think having a simple struct instead of tuples would work as well and would be not too much overhead to manage.

Doing the tuple transformation into a GraphQL object in the graphql module has the advantage that everything graphql related is done in one place. Wouldn't you validate the InputObject in the graphql module anyway? then it shouldn't be a problem(?)

Either way, if we want to avoid having intermediate domain objects we either have to expose api objects in `repository` or DB objects in the API. IMHO the later is the cleaner and more flexible way to do it.

src/server/service/graphql/schema/types/invoices.rs
clemens-msupply: sorry think the discussion got lost in the chat...

Thats the simple query I liked to do:
```sql
SELECT invoice.id, SUM(invoice_line.total_after_tax) as total_after_tax FROM invoice LEFT JOIN invoice_line ON (invoice_line.invoice_id = invoice.id)
GROUP BY invoice.id;
```
but Diesel doesn't like to have an aggregate type AND a non-aggregate type in the same select. Happy if you have a solution (maybe I am just missing the simple solution here)

wlthomson: Hm... played around a bit and got a bit further by using sql for the aggregating. I ended up with something like this:

```
invoice_dsl::invoice
  .select((
    invoice_dsl::id,
    sql::<Integer>("sum(cost_price_per_pack) as total_after_tax"),
  ))
  .left_join(
     invoice_line_dsl::invoice_line.on(invoice_dsl::id.eq(invoice_line_dsl::invoice_id)),
   )
   .group_by(invoice_dsl::id)
   .load(&connection)?;
```

Does that solve any of the problems you experienced?

clemens-msupply:
I pushed an implementation using Diesel without Views + test. Find this the more straight forward way to do it for this case but not opposed to Views either...

wlthomson (approved):

PR #361, opened 2021-09-23
#343 sync error handling mechanism

### LINKED ISSUE #343: [ERROR HANDLING] - Sync - Error handling Mechanism
MW parent job: `[OMSBEE] omSupply: Back End - Error Handling`

### Description

No panics or unwraps in error handling.

I would assume that SyncActor will actual call just one method (which has the content of current SyncActor), and will analyse the result (probably just log it for now). In the future I would assume scheduler would be told to run next sync at different intervals based on error: i.e. connection error -> retry soon (or if username and password not correct stop further scheduling etc..), no error -> normal interval, critical error -> ... )

I like the idea of one Error type with nested enums (and nested enums holding original error <- this kind of couples our DB to our error though?) 

### Table

none

### Scope

<ins>Includes</ins>
* Propagating all errors in sync up
* Match and log errors for now

<ins>Excludes</ins>
* Doing anything more then logging and matching with final caught error

### Tasks

- [ ] MW job added

wlthomson:
Closes #343.

Sorry for the delay, spent a bit of time trying a few different approaches/crates and was quite tricky to find a good way to organise the error hierarchy (mostly due to the logic being spread around a bit... once we get around to defining our layers and repo vs. service reponsibilities, I think this will become a lot easier). I think this will end up more of a ` 3` than a `2` :(.

Think this does the job, but could probably be improved. Haven't updated tests yet (which I expect won't be a trivial task if we want to test all errors are handled as intended), as want to get some feedback before going any deeper.


src/util/sync/synchroniser.rs
```diff
@@ -0,0 +1,278 @@
...
+    #[error("Failed to pull central sync records")]
+    PullCentralSyncRecordsError { source: SyncConnectionError },
+    #[error("Failed to update central sync buffer records")]
+    UpdateCentralSyncBufferRecordsError { source: RepositoryError },
+    #[error("Failed to get central sync cursor record")]
+    GetCentralSyncCursorRecordError { source: RepositoryError },
+    #[error("Failed to get central sync buffer records")]
+    GetCentralSyncBufferRecordsError { source: RepositoryError },
+    #[error("Failed to import central sync buffer records")]
+    ImportCentralSyncRecordsError { source: SyncImportError },
```
clemens-msupply, line 26: Just a comment: I had very unhelpful diesel errors during sync from my local mSupply server and I was thinking of giving the RepositoryError some context (maybe using anyerror) to find out where the problem was... Alternatively we could do logging for every part of the import...

src/util/sync/synchroniser.rs
```diff
@@ -0,0 +1,278 @@
...
+        &self,
+        registry: &RepositoryRegistry,
+    ) -> Result<(), CentralSyncError> {
+        let central_sync_buffer_repository: &CentralSyncBufferRepository =
+            registry.get::<CentralSyncBufferRepository>();
+        let sync_session = registry
+            .get::<SyncRepository>()
+            .new_sync_session()
+            .await
+            .unwrap();
```
clemens-msupply, line 162: One unwrap here. The sync_session is also removed in main so probably need to be resolved when merging...

src/util/sync/translation/mod.rs
```diff
@@ -23,27 +23,65 @@ use self::{
...
+    #[error("Failed to translate item record")]
+    ItemTranslationError { source: serde_json::Error },
+    #[error("Failed to translate store record")]
+    StoreTranslationError { source: serde_json::Error },
+    #[error("Failed to translate list master record")]
+    ListMasterTranslationError { source: serde_json::Error },
+    #[error("Failed to translate list master line record")]
+    ListMasterLineTranslationError { source: serde_json::Error },
+    #[error("Failed to translate list master name join record")]
+    ListMasterNameJoinTranslationError { source: serde_json::Error },
```
clemens-msupply, line 41: Every new translation would need its own error... Could there maybe be one error for all of them? or a `SimpleTranslationError { translation: String, source:...}` Not sure how that works with the thiserror message though...

clemens-msupply (commented):
Looks good to me, just some smaller comments...

src/util/sync/connection.rs
```diff
@@ -10,6 +10,59 @@ use reqwest::{
...
+    #[error("Sync server responded with status 401 Unauthorized")]
+    UnauthorisedError { source: reqwest::Error },
+    #[error("Sync server responded with status 404 Not found")]
+    NotFoundError { source: reqwest::Error },
+    #[error("Sync server responded with status 405 Method Not Allowed")]
+    MethodNotAllowedError { source: reqwest::Error },
+    #[error("Sync server responded with status 500 Internal Server Error")]
+    InternalServerError { source: reqwest::Error },
+    #[error("Unknown sync connection error")]
+    UnknownError { source: reqwest::Error },
```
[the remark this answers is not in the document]
wlthomson: Yeah, I can see the thinking as this is currently just reflecting the underlying reqwest error. However, that's just the case now, The nice thing about wrapping it is that it encapsulates the errors and makes our lives easier down the road if and when we want to swap out or upgrade `reqwest`, or add additional non-reqwest errors. If and when that happens,  we only need to make updates here, rather than in the error handling spread throughout the rest of the code. I also found it frustrating that the `reqwest::Error` is a struct not a enum - every time you drill down you have to examine the object methods. By doing all that in the `from`, we expose the error declaratively (consistent with our other errors).

src/util/sync/synchroniser.rs
wlthomson: We could update the `RepositoryError` to use the pattern I've done here (e.g. add a `source` field with the original diesel error)?

src/util/sync/synchroniser.rs
wlthomson: Nice spot, I left that `unwrap` to be removed in the merge 👍.

src/util/sync/synchroniser.rs
clemens-msupply: The original diesel error is unfortunately not helpful (have to double check though if I can't get more out of it). Meant more like an error detailt like msg: "Foreign key violation when importing 'items"..."

src/util/sync/translation/mod.rs
wlthomson: Yeah, I thought about that, but then I questioned what we would get from generalising these errors into a single error... and I couldn't really see any benefits? 

My thinking here was to try to maintain a nice consistent pattern for error handling. My thinking was that if you have a single error, you have to drill down into the serde error to find out what data type was involved, or add another field to the error which contains that information. By adding a field to the error, we have to populate that field every time the error is thrown (here only once, but good to consider how patterns will work in the general case, too), and we lose a lot of the simplicity of the recursive pattern (it's nice to be able to "drill down", with each error representing all the info about the error that the current method knows, and the `source` representing the next level down). By keeping the errors "flat" (i.e. representing the error variations declaratively) we also minimise the amount of work the caller has to do: by adding a field, the caller has to match twice in order to handle the different variations (on the error and then on the field)... we might only need to log here, but we probably should take into consideration that often the caller will want to handle different errors differently, not just log the message.

Are there benefits to the other approach/drawbacks to this one that I've missed?

src/util/sync/connection.rs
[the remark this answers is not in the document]
chris-morgan: Just to mention it (without engaging further or judging suitability in any way): another option would be a struct with a *kind* field like std::io::ErrorKind.

src/util/sync/synchroniser.rs
wlthomson: Ah, maybe something to visit in a separate issue then? If you like the idea here of going declarative instead of nested fields, another approach might also be to wrap the diesel issues in our own lower-level `DatabaseError`? e.g. `RepositoryError { source: DatabaseError }`?

src/util/sync/connection.rs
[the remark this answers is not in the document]
wlthomson: Hm, that could also work! Am I right that that approach is equal in expressive power? Is it mostly a stylistic decision?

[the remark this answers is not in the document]
chris-morgan: Yeah, it’s basically just observing that all the variants have the same fields and choosing therefore to extract the discriminant out and treat it separately from the associated data. You could do much the same with your `SyncTranslationError`, though there I’d be tempted to just use `record: &'static str` instead of defining an enum, depending on whether you will ever actually match on them, and whether you want to extend it in plugins or anything (in which case such an enum won’t be suitable).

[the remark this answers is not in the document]
wlthomson: Ah yup, sounds like you are saying the same thing as here https://github.com/openmsupply/remote-server/pull/361#discussion_r714448378? I suppose I could just do something like:

```
TranslationError {
  record: &'static str,
  source: serde_json::Error
}
```

As in my response to @clemens-msupply, I'm just wondering if this sacrifices a bit of flexibility in terms of matching (but maybe matching isn't really a concern here).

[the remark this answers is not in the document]
clemens-msupply: Yes, you probably don't need this level of detail in the higher level other than for debugging and maybe logging... e.g. you usually don't handle an internal server error anywhere else than directly at the api call

src/util/sync/translation/mod.rs
clemens-msupply: I think its just simplicity, we might not even match on the translation errors or only on a very small sub set. If we do eventually need to match on a special errors it should be easy to add the required error or use an error kind? We probably should just try it out and then we see to pros and cons and what works for us...

src/util/sync/synchroniser.rs
clemens-msupply: Yes separate issue. Don't know a solution yet...

src/util/sync/connection.rs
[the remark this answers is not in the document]
andreievg: I actually like this, except would have gone for

```rust
pub struct SyncConnectionError {
   source: ...
   error: enum
}
```

One of the beauties of exceptions is you can handle them in one place, rather than error handling scattered throughout your code, leading to hard to follow branching.

At higher level we do want to know if it's UnauthorisedError or just plain Connection error or if it's InternalServer error, we can handle all of those in one place (and request different rescheduler or stop sync all together)

src/util/sync/connection.rs
```diff
@@ -90,14 +137,37 @@ impl SyncConnection {
         Ok(sync_batch)
     }
 
+    // Acknowledge successful integration of records from sync queue.
+    pub async fn acknowledge_remote_records(
```
andreievg, line 141: was hoping we could leave remote sync stuff until it's in spec, we are likely to change a few things from the prototype (like acknowledgement is done at the start of batch sync operation, not the end) and leaving it avoids double handling.

src/util/sync/mod.rs
```diff
@@ -1,300 +1,19 @@
...
 pub use central::CentralSyncBatch;
-pub use connection::SyncConnection;
+pub use connection::{SyncConnection, SyncConnectionError};
 pub use credentials::SyncCredentials;
 pub use remote::{
     RemoteSyncAcknowledgement, RemoteSyncBatch, RemoteSyncRecord, RemoteSyncRecordAction,
     RemoteSyncRecordData,
 };
 pub use server::SyncServer;
-
```
andreievg, line 16: could use `git mv` (to actors.rs), would avoid big diff

src/util/sync/synchroniser.rs
andreievg: Just fyi, it's a good idea to run postgres in `log_statements = all` mode and have log file in terminal with `tail -f`, not sure about sqlite though, i think there must be a way to stream all queries on server ?

src/util/sync/translation/mod.rs
```diff
@@ -23,27 +23,65 @@ use self::{
...
 /// Translates sync records into the local DB schema.
 /// Translated records are added to integration_records.
 fn do_translation(
     sync_record: &CentralSyncBufferRow,
     integration_records: &mut IntegrationRecord,
-) -> Result<(), String> {
-    if let Some(row) = LegacyNameRow::try_translate(sync_record)? {
+) -> Result<(), SyncTranslationError> {
+    if let Some(row) = LegacyNameRow::try_translate(sync_record)
+        .map_err(|source| SyncTranslationError::NameTranslationError { source })?
```
andreievg, line 65: Nit:

```
use self::SynctranslationError::*;
...
.map_err(|source| NameTranslationError { source })?
...
```

andreievg: Should this error mapping be in try_translate though ?

src/util/sync/connection.rs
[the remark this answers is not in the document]
andreievg: forget about that shape, that's not inline with thiserror

andreievg (approved):
Thanks !

Another Nit:

```rust
pub enum SyncTranslationError {
    #[error("{0}")]
    NameTranslationError(#[from] NameTranslationError)

// ...

 if let Some(row) = LegacyNameRow::try_translate(sync_record)? {

// ...
// in LagacyNameRow
#[derive(Error, Debug)]
#[error("Failed to translate name record")]
pub struct NameTranslationError(#[from] serde_json::Error);
// ...

  ) -> Result<Option<NameRow>, NameTranslationError> {
        if sync_record.table_name != "name" {
            return Ok(None);
        }
        let data = serde_json::from_str::<LegacyNameRow>(&sync_record.data)?;


```

I think we can leave out #[error] in this case on SyncTranslationError entirely and just `#[derive(Display)]`


Also was wondering why not use units in enum vs struct like syntax for individual errors which just hold source ?

src/util/sync/translation/mod.rs
wlthomson: Hm, yeah good point. I think it should be!

wlthomson:
> 
> 
> Thanks !
> 
> Another Nit:
> 
> ```rust
> pub enum SyncTranslationError {
>     #[error("{0}")]
>     NameTranslationError(#[from] NameTranslationError)
> 
> // ...
> 
>  if let Some(row) = LegacyNameRow::try_translate(sync_record)? {
> 
> // ...
> // in LagacyNameRow
> #[derive(Error, Debug)]
> #[error("Failed to translate name record")]
> pub struct NameTranslationError(#[from] serde_json::Error);
> // ...
> 
>   ) -> Result<Option<NameRow>, NameTranslationError> {
>         if sync_record.table_name != "name" {
>             return Ok(None);
>         }
>         let data = serde_json::from_str::<LegacyNameRow>(&sync_record.data)?;
> ```
> 
> I think we can leave out #[error] in this case on SyncTranslationError entirely and just `#[derive(Display)]`
> 
> Also was wondering why not use units in enum vs struct like syntax for individual errors which just hold source ?

No reason, they are equal in expressive power. I just liked having a common interface for all errors 🤷 . There are a few ways of achieving the same outcome, mostly stylistic preference, it seems.

src/util/sync/connection.rs
wlthomson: Hm, OK will remove!

andreievg:
>> Also was wondering why not use units in enum vs struct like syntax for individual errors which just hold source ?

> No reason, they are equal in expressive power. I just liked having a common interface for all errors 🤷 . There are a few ways of achieving the same outcome, mostly stylistic preference, it seems.

Sweet, sounds good, although a would like to discuss this next week. Above style suggest we never need to use tuple like enum variant and will always use struct, or use them based on individual stylist preference, which maybe slightly inconsistent (In general in rust code i've looked at i've never seen one key -> value named fields, it's alway been a tuple)

src/util/sync/translation/mod.rs
wlthomson: beb775f.

src/util/sync/connection.rs
wlthomson: a34190f.

wlthomson:
Tried a few different ways to introduce additional tests, but eventually realised testing here is going to overlap a bit with some of the unfinalised design decisions, so going to park them for now until we have agreement on how we want to test with the sync server and to what extent we want to test error states. Initial thoughts are that we might want to mock test against an ideal sync server as well as do some e2e testing against an actual live sync server. There are a few more complexities here, but will write them up in the design discussion.

wlthomson:
Still stuck on this 😢! Wrangling with some failing tests, will investigate further tomorrow!

Issue #374, opened 2021-09-26
Make it possible to do granular mutations on full invoice in repository


andreievg:
Will be required for: https://github.com/openmsupply/remote-server/issues/372. So in effort to reduce size of next PR, breaking it down further

PR #375, opened 2021-09-26
374 Granular mutations on for full invoice shape

andreievg:
Fixes: #374 

At first I thought we can use existing FullInvoiceShape, but then realised it would be to difficult and mess with 'read' part of invoice too much. So decided to go with different approach, which i kind of like, and i think we can make it a standard for mutations across all repos, i.e just one method 'mutate' in repo which accepts Mutations<Type>.

### DIFF
--- a/src/business/supplier_invoice/insert.rs
+++ b/src/business/supplier_invoice/insert.rs
@@ -17,7 +17,7 @@ use crate::{
 
 use super::{
     check_invoice_insert, check_other_party_insert, current_date_time, current_store_id,
-    get_new_insert_lines, FullInvoice, InsertSupplierInvoiceError,
+    get_insert_line_and_batches, FullInvoiceMutation, InsertSupplierInvoiceError, Mutations,
 };
 
 impl From<RepositoryError> for InsertSupplierInvoiceError {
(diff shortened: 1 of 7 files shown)


clemens-msupply (commented):
Just some initial comments:

DB Related:

The insert call, when checking if the invoice already exists: doing this out of the write tx doesn't help much, i.e. if the UI does the call twice both request may pass the check but one write will fail (hope that makes sense). Better would be to report the already existing invoice when writing the invoice.

Note: Postgress has on default only Read Committed isolation level that could cause weird inconsistencies if you read, insert and deletes lines concurrently (even in the same transaction). Just something we should be aware of... Sqlite is fine since it only has Serlizable isolation level.

Error handling:

I found it quite difficult to follow what is checked, mostly because there are many method just transforming errors.

Some of the intermediate errors can be removed, e.g. CheckOtherParty, or what is their intended purpose?

Comments and sometimes better method names would help to understand what the checks are doing.

There are actually not too many checks and they would probably easily fit into one method, e.g. `validate_invoice_{line}()`. This would make it very easy to follow. I would only start splitting the validation logic up on demand.

clemens-msupply:
The problem with one big mutate is that you need to implement it for every Repo/service that needs to do something in a transaction. Furthermore, the reads are done in another transaction which means you write something on assumptions that might have changed in the meantime. We should just use DB transactions; we can make Repositories work with them transparently which would remove a lot of code already and is very flexible.

PR #377, opened 2021-09-27
Implement the invoice GraphQL endpoint

clemens-msupply:
Still to come:
- Unified InvoiceNode for invoices and invoice
- Schema changes Decimal -> Double, expiry date

### DIFF
--- a/src/server/service/graphql/schema/types/invoice_query.rs
+++ b/src/server/service/graphql/schema/types/invoice_query.rs
@@ -1,0 +1,147 @@
+use crate::{
+    database::{
+        repository::{InvoiceLineQueryJoin, InvoiceLineQueryRepository, InvoiceQueryJoin},
+        schema::{InvoiceRowStatus, InvoiceRowType},
+    },
+    server::service::graphql::ContextExt,
+};
+
+use async_graphql::{Context, Enum, Object, SimpleObject};
+use chrono::{DateTime, Utc};
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug)]
+pub enum InvoiceType {
+    CustomerInvoice,
+    SupplierInvoice,
+}
+
+impl From<InvoiceRowType> for InvoiceType {
+    fn from(row: InvoiceRowType) -> InvoiceType {
+        match row {
+            InvoiceRowType::CustomerInvoice => InvoiceType::CustomerInvoice,
+            InvoiceRowType::SupplierInvoice => InvoiceType::SupplierInvoice,
+        }
+    }
+}
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, Debug)]
+pub enum InvoiceStatus {
+    Draft,
+    Confirmed,
+    Finalised,
+}
+
+impl From<InvoiceRowStatus> for InvoiceStatus {
+    fn from(row: InvoiceRowStatus) -> InvoiceStatus {
+        match row {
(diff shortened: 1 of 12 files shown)


src/server/service/graphql/schema/types/invoice_query.rs
```diff
@@ -0,0 +1,147 @@
...
+        InvoiceQuery {
+            id: invoice_row.id.to_owned(),
+            other_party_name: name_row.name,
+            other_party_id: name_row.id,
+            status: InvoiceStatus::from(invoice_row.status),
+            invoice_type: InvoiceType::from(invoice_row.r#type),
+            invoice_number: invoice_row.invoice_number,
+            their_reference: invoice_row.their_reference,
+            comment: invoice_row.comment,
+            entry_datetime: DateTime::<Utc>::from_utc(invoice_row.entry_datetime, Utc).to_rfc3339(),
```
andreievg, line 71: I think we should use DateTime<Utc> for datetime GraphQL types (as this exposes the right type introspection), and we can just remove to_rfc3339). Also, maybe a helper method ?

src/server/service/graphql/schema/types/invoice_query.rs
```diff
@@ -0,0 +1,147 @@
...
+            item_id: item.id,
+            item_name: item.name,
+            item_code: item.code,
+            pack_size: invoice_line.pack_size,
+            number_of_packs: invoice_number_of_packs,
+            cost_price_per_pack: invoice_line.cost_price_per_pack,
+            sell_price_per_pack: invoice_line.sell_price_per_pack,
+            batch: invoice_line.batch,
+            expiry_date: invoice_line.expiry_date,
+            stock_line: StockLine {
```
andreievg, line 135: hmm, i guess this is ok for now, but the reason made this a new resolver (vs availableNumberOfPacks in InvoiceLineNode), was that it's not always required by front end

andreievg (changes requested):
Hi, thanks, just one request, really (datetime in graphql schema)

src/server/service/graphql/schema/types/invoice_query.rs
clemens-msupply: Thanks wasn't aware about GraphQL supporting it. Added in 3ec204e

src/server/service/graphql/schema/types/invoice_query.rs
clemens-msupply: yes I added a TODO

src/util/datetime.rs
```diff
@@ -0,0 +1,5 @@
+use chrono::{DateTime, NaiveDateTime, Utc};
+
+pub fn naive_date_time_to_utc(datatime: NaiveDateTime) -> DateTime<Utc> {
+    DateTime::<Utc>::from_utc(datatime, Utc)
```
wlthomson, line 4: I like the idea of having a 'DateTimeService' for this sort of thing (later). Would be nice for utility functions to also be part of the domain model!

wlthomson (approved):

andreievg (approved):
thanks

src/util/datetime.rs
clemens-msupply: In this case I would say its so simple that its fine to just have it as a utils function. But I guess if we have more time related functions, e.g. get_store_timezone, it would makes sense.

PR #378, opened 2021-09-27
308 schema updates

clemens-msupply:
- Use TIMESTAMP for the expiry_date (stock_line and invoice_line)
- Use DOUBLE instead of DECIMAL


src/database/schema/stock_line.rs
```diff
@@ -12,5 +14,5 @@ pub struct StockLineRow {
     pub sell_price_per_pack: f64,
     pub available_number_of_packs: i32,
     pub total_number_of_packs: i32,
-    pub expiry_date: Option<String>,
+    pub expiry_date: Option<NaiveDateTime>,
```
andreievg, line 17: I actually made this NaiveDate (both in db and graphql schema), when I was doing supplier invoice, it translates right away without `from_utc`. I still think this should be NaiveDate (`Date` in postgres, `Text` in sqlite, `Date` in dsl), you can ask Craig if you want (but the discussion about it needing to be Craig was with him)

andreievg (changes requested):

andreievg (approved):

PR #379, opened 2021-09-27
Unify invoice query return value (invoice and invoices)

clemens-msupply:


src/server/service/graphql/schema/types/invoice_query.rs
```diff
@@ -54,12 +64,36 @@ pub struct InvoiceQuery {
     entry_datetime: String,
     confirm_datetime: Option<String>,
     finalised_datetime: Option<String>,
-    lines: InvoiceQueryLines,
+    lines: InvoiceLines,
+}
+
+#[ComplexObject]
+impl InvoiceNode {
+    async fn pricing(&self, ctx: &Context<'_>) -> InvoiceLinePricing {
```
andreievg, line 72: This is quite interesting, `stats`, is unsafe in two ways (raw query + group_by no compile time type checking), and we are completely ignoring any errors with `ok()`. So if we have type mismatch or query is wrong it would just show 0.0 ? 

I would prefer unwrap for now

andreievg (changes requested):
InvoiceLines are now N+1 for Invoices Query ?

src/server/service/graphql/schema/types/invoice_query.rs
clemens-msupply: The loader's load_one() returns a Option anyway so we also have to handle this case, i.e. what to do when there are no lines, return an Option or return a default? I went with the default in: d181c12

andreievg (approved):

andreievg (changes requested):
oops i approved, but realised the N + 1 thing (in first PR review comment)

clemens-msupply:
Sorry forgot about your n+1 comment. Done in 
431cd52

andreievg (approved):
thanks

PR #381, opened 2021-09-28
Store sqlite test db files in test_output directory

clemens-msupply:
This helps to keep the project directory clean.

### DIFF
--- a/src/util/test_db.rs
+++ b/src/util/test_db.rs
@@ -58,8 +58,12 @@ pub async fn setup(db_settings: &DatabaseSettings) {
     const MIGRATION_PATH: &str = "sqlite";
 
     let db_path = format!("./{}.sqlite", db_settings.database_name);
-
     fs::remove_file(&db_path).ok();
+
+    // create parent dirs
+    let path = std::path::Path::new(&db_path);
+    let prefix = path.parent().unwrap();
+    fs::create_dir_all(prefix).unwrap();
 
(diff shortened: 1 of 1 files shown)


andreievg (approved):

src/util/test_db.rs
```diff
@@ -75,6 +79,17 @@ pub async fn setup(db_settings: &DatabaseSettings) {
...
 
+#[cfg(feature = "postgres")]
+fn make_test_db_name(base_name: String) -> String {
+    base_name
+}
+
+#[cfg(feature = "sqlite")]
+fn make_test_db_name(base_name: String) -> String {
+    // store all test db files in a test directory
+    format!("test_output/{}", base_name)
```
wlthomson, line 90: Could we clean-up the directory after the tests are all finished?

wlthomson (approved):
LGTM.

Would also prob be good to add this to `.gitignore`?

src/util/test_db.rs
clemens-msupply: Had the same thought and I am still not sure about it. I think its actually nice to keep the sqlite files for debugging, e.g. when something went wrong you can straight look into the db and you don't have to disable the deletion first... However, you could also just rerun the test and set a breakpoint at the correct line...

clemens-msupply: *.sqlite is already in .gitignore...

PR #383, opened 2021-09-30
#382 Use NaiveDate type for the expiry date

### LINKED ISSUE #382: Use NaiveDate for expiry date

clemens-msupply:
Closes #382


src/database/mock/invoice_line.rs
```diff
@@ -9,7 +9,7 @@ pub fn mock_customer_invoice_a_invoice_lines() -> Vec<InvoiceLineRow> {
         item_id: String::from("item_a"),
         stock_line_id: Some(String::from("item_a_line_a")),
         batch: Some(String::from("item_a_line_a")),
-        expiry_date: Some(Utc::now().naive_utc()),
+        expiry_date: Some(Utc::now().naive_utc().date()),
```
wlthomson, line 12: We probably want our test data to be identical across different tests, otherwise our test results may not be replicable. I would suggest just choosing an arbitrary date for this?

src/database/repository/tests.rs
```diff
@@ -186,7 +186,7 @@ mod repository_test {
                 invoice_id: invoice_1().id.to_string(),
                 stock_line_id: None,
                 batch: Some("".to_string()),
-                expiry_date: Some(NaiveDateTime::from_timestamp(4000, 0)),
+                expiry_date: Some(NaiveDateTime::from_timestamp(2 * 24 * 60 * 60, 0).date()),
```
wlthomson, line 189: Why the timestamp? Could perhaps simplify with something like `from_ymd`?

migrations/sqlite/00000105_create_stock_line_table/up.sql
```diff
@@ -3,7 +3,7 @@ CREATE TABLE stock_line (
     item_id TEXT NOT NULL REFERENCES item(id),
     store_id TEXT NOT NULL REFERENCES store(id),
     batch TEXT,
-    expiry_date TIMESTAMP,
+    expiry_date DATE,
```
wlthomson, line 6: Does this work? I thought SQLite didn't have an inbuilt type for dates or times (https://www.sqlite.org/datatype3.html)?

wlthomson (changes requested):

src/database/repository/tests.rs
clemens-msupply: done 731eb54 thanks

src/database/mock/invoice_line.rs
clemens-msupply: done 731eb54 thanks

migrations/sqlite/00000105_create_stock_line_table/up.sql
clemens-msupply: Don't know why it works (diesel or sqlite) but changed it to text ebc1a80

src/database/mock/invoice_line.rs
```diff
@@ -120,7 +120,7 @@ pub fn mock_supplier_invoice_b_invoice_lines() -> Vec<InvoiceLineRow> {
         item_id: String::from("item_a"),
         stock_line_id: Some(String::from("item_a_line_a")),
         batch: Some(String::from("item_a_line_a")),
-        expiry_date: Some(Utc::now().naive_utc()),
+        expiry_date: Some(NaiveDate::from_ymd(2020, 8, 7)),
```
wlthomson, line 123: If we use these in production we should probably add some unit tests to make sure the dates are being parsed and serialised as we expect them to be... for now seems OK though.

wlthomson (approved):

PR #384, opened 2021-09-30
306 name filters

### LINKED ISSUE #306: [API] - Filtering
MW parent job: `[OMSBEA] omSupply Back End - API`

### Description

Mechanism for filtering

### Scope

<ins>Includes</ins>
* For one query only (say names)

<ins>Excludes</ins>
* Filters by related tables

### Tasks

- [ ] MW job added

clemens-msupply:
Closes #306

### DIFF
--- a/src/database/repository/diesel/name_query.rs
+++ b/src/database/repository/diesel/name_query.rs
@@ -4,7 +4,7 @@ use crate::{
         repository::RepositoryError,
         schema::{
             diesel_schema::{
-                name_store_join::dsl::name_store_join, name_table::dsl as name_table_dsl,
+                name_store_join::dsl as name_store_join_dsl, name_table::dsl as name_table_dsl,
             },
             NameRow, NameStoreJoinRow,
         },
@@ -42,6 +42,13 @@ impl From<NameAndNameStoreJoin> for NameQuery {
     }
 }
 
+pub struct NameQueryFilter {
+    pub name: Option<String>,
+    pub code: Option<String>,
+    pub is_customer: Option<bool>,
+    pub is_supplier: Option<bool>,
+}
+
 pub struct NameQueryRepository {
     pool: Pool<ConnectionManager<DBBackendConnection>>,
 }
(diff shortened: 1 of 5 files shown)


src/database/repository/diesel/name_query.rs
```diff
@@ -42,6 +42,18 @@ impl From<NameAndNameStoreJoin> for NameQuery {
     }
 }
 
+pub struct NameQueryStringFilter {
+    pub equal_to: Option<String>,
+    pub like: Option<String>,
+}
```
wlthomson, line 48: Shouldn't we just have a generic type for filtering string types? Is a `NameQueryStringFilter` going to be any different from an `ItemQueryStringFilter`?

src/database/repository/diesel/name_query.rs
```diff
@@ -42,6 +42,18 @@ impl From<NameAndNameStoreJoin> for NameQuery {
...
+
+pub struct NameQueryFilter {
+    pub name: Option<NameQueryStringFilter>,
+    pub code: Option<NameQueryStringFilter>,
+    pub is_customer: Option<bool>,
+    pub is_supplier: Option<bool>,
```
wlthomson, line 54: Unless mistaken, these are mutually exclusive (if they have the same value, the result set will always be empty), so they can be represented with a single bool (or a enum with two variants)?

src/database/repository/diesel/name_query.rs
```diff
@@ -58,18 +70,47 @@ impl NameQueryRepository {
...
+        let connection = get_connection(&self.pool)?;
+
+        let mut query = name_table_dsl::name_table
+            .left_join(name_store_join_dsl::name_store_join)
             .offset(pagination.offset())
             .limit(pagination.first())
+            .into_boxed();
+
+        if let Some(f) = filter {
+            if let Some(code) = &f.code {
```
wlthomson, line 88: I'm guessing the diesel DSL syntax makes it hard to reuse logic here?

src/database/repository/tests.rs
```diff
@@ -300,6 +321,102 @@ mod repository_test {
...
+                        like: None,
+                    }),
+                    is_customer: None,
+                    is_supplier: None,
+                }),
+            )
+            .unwrap();
+        assert_eq!(result.len(), 2);
+
+        /* TODO currently no way to add name_store_join rows for the following tests:
```
wlthomson, line 388: Personally, I think `TODOs` should be made into issues rather than left in the codebase (where possible), but I'm assuming you don't want to lose the work you've done on the code!

src/server/service/graphql/schema/queries/mod.rs
```diff
@@ -24,8 +24,12 @@ impl Queries {
         &self,
         _ctx: &Context<'_>,
         #[graphql(desc = "pagination (first and offset)")] page: Option<Pagination>,
+        #[graphql(desc = "filters")] filter: Option<NameFilter>,
     ) -> NameList {
-        NameList { pagination: page }
+        NameList {
+            pagination: page,
+            filter: filter.map(NameQueryFilter::from),
```
wlthomson, line 31: This is confusing to me, because the gql `NameList` object is taking both gql and application objects as parameters. From our discussion today, I'm assuming this will be refactored into the new approach and the `NameList` object will be goners?

src/server/service/graphql/schema/types/name.rs
```diff
@@ -15,6 +15,42 @@ pub struct NameQuery {
...
 pub struct NameList {
     pub pagination: Option<Pagination>,
+    pub filter: Option<NameQueryFilter>,
+}
+
+#[derive(InputObject)]
+pub struct NameStringFilter {
+    equal_to: Option<String>,
+    like: Option<String>,
+}
```
wlthomson, line 25: Same as above, think we can have one distinct type for all string filtering operations?

wlthomson (approved):

src/database/repository/diesel/name_query.rs
clemens-msupply: Not sure if we have the same options everywhere or if some endpoints need additional options like starts_with... I would say lets keep it for now and see how it looks for more endpoints...

src/database/repository/diesel/name_query.rs
clemens-msupply: will add a todo

src/database/repository/tests.rs
clemens-msupply: Yes I think that will be easy to fix once we have the repository...

src/database/repository/diesel/name_query.rs
clemens-msupply: there is nothing to reuse yet :) but yes I guess Diesel will make it pretty hard...

src/server/service/graphql/schema/queries/mod.rs
clemens-msupply: Will fix the mix

PR #387, opened 2021-09-30
Generic connection example

andreievg:
For generic way to work have to use look ahead, or get total count even if it's not requested


andreievg:
implemented in: https://github.com/openmsupply/remote-server/pull/407

PR #388, opened 2021-09-30
Fix duplicated migration number

clemens-msupply:
Fix duplicated migration order and remove "unused" diesel.toml file which causes `diesel migration` to overwrite the diesel_schema.rs file


wlthomson (approved):
This was SO annoying - great to have it fixed!

PR #392, opened 2021-10-05
307 invoices filter sort

clemens-msupply:
Add filter and sort options for the invoices endpoint.

- Define common filter and sort types/struct that can be reused between endpoints (both on repository layer and graphql layer)
- Use common structs for the name endpoint
- Refactor repo to graph gql enum mapping for the invoices endpoint

### DIFF
--- a/src/database/repository/diesel/invoice_query.rs
+++ b/src/database/repository/diesel/invoice_query.rs
@@ -1,3 +1,7 @@
+use super::{
+    get_connection, DBBackendConnection, DatetimeFilter, EqualFilter, SimpleStringFilter, Sort,
+};
+
 use crate::{
     database::{
         repository::RepositoryError,
@@ -5,18 +9,38 @@ use crate::{
             diesel_schema::{
                 invoice::dsl as invoice_dsl, name_table::dsl as name_dsl, store::dsl as store_dsl,
             },
-            InvoiceRow, NameRow, StoreRow,
+            InvoiceRow, InvoiceRowStatus, InvoiceRowType, NameRow, StoreRow,
         },
     },
     server::service::graphql::schema::queries::pagination::{Pagination, PaginationOption},
 };
 
-use super::{get_connection, DBBackendConnection};
-
 use diesel::{
     prelude::*,
     r2d2::{ConnectionManager, Pool},
 };
+
+pub struct InvoiceFilter {
+    pub name_id: Option<EqualFilter<String>>,
+    pub store_id: Option<EqualFilter<String>>,
+    pub r#type: Option<EqualFilter<InvoiceRowType>>,
+    pub status: Option<EqualFilter<InvoiceRowStatus>>,
+    pub comment: Option<SimpleStringFilter>,
+    pub their_reference: Option<EqualFilter<String>>,
+    pub entry_datetime: Option<DatetimeFilter>,
+    pub confirm_datetime: Option<DatetimeFilter>,
+    pub finalised_datetime: Option<DatetimeFilter>,
+}
+
+pub enum InvoiceSortField {
+    Type,
+    Status,
+    EntryDatetime,
+    ConfirmDatetime,
+    FinalisedDateTime,
+}
+
+pub type InvoiceSort = Sort<InvoiceSortField>;
 
 pub struct InvoiceQueryRepository {
     pool: Pool<ConnectionManager<DBBackendConnection>>,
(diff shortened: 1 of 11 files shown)


src/server/service/graphql/schema/queries/mod.rs
```diff
@@ -57,8 +61,15 @@ impl Queries {
         &self,
         _ctx: &Context<'_>,
         #[graphql(desc = "pagination (first and offset)")] page: Option<Pagination>,
+        #[graphql(desc = "filters option")] filter: Option<InvoiceFilterInput>,
+        #[graphql(desc = "sort options (only first sort input is evaluated for this endpoint)")]
+        sort: Option<Vec<InvoiceSortInput>>,
     ) -> InvoiceList {
-        InvoiceList { pagination: page }
+        InvoiceList {
+            pagination: page,
```
wlthomson, line 69: Imo, if we are going to keep this data struct, may as well rename `pagination` to `page` (or vice-versa). Seems like an unncessary distinction?

src/server/service/graphql/schema/types/invoice_query.rs
```diff
@@ -11,37 +10,20 @@ use async_graphql::{dataloader::DataLoader, ComplexObject, Context, Enum, Object
 use chrono::{DateTime, NaiveDate, Utc};
 
 #[derive(Enum, Copy, Clone, PartialEq, Eq, Debug)]
-pub enum InvoiceType {
+#[graphql(remote = "crate::database::schema::InvoiceRowType")]
```
wlthomson, line 13: Can you import this and just do: `remote = "InvoiceRowType"`?

src/server/service/graphql/schema/types/invoices_query.rs
```diff
@@ -21,8 +69,21 @@ impl InvoiceList {
     async fn nodes(&self, ctx: &Context<'_>) -> Vec<InvoiceNode> {
         let repository = ctx.get_repository::<InvoiceQueryRepository>();
 
+        let filter = self.filter.clone().map(InvoiceFilter::from);
+
+        // Currently only one sort option is supported, use the first from the list.
+        let first_sort = self
+            .sort
+            .as_ref()
```
wlthomson, line 77: Is this being used to unwrap the `Option`? If the field is optional, shouldn't there be some matching logic here to use some default sorting if the argument is not specified?

src/server/service/graphql/schema/types/invoices_query.rs
```diff
@@ -21,8 +69,21 @@ impl InvoiceList {
...
+            .map(|sort_list| sort_list.get(0))
```
wlthomson, line 78: I think `.first()` is more self-documenting here?

src/server/service/graphql/schema/types/invoices_query.rs
```diff
@@ -21,8 +69,21 @@ impl InvoiceList {
...
+            .flatten()
```
wlthomson, line 79: To avoid flattening, you could just do something like:

```
let first_sort = match self.sort {
  Some(sort) => { 
    let first_sort = sort.first().unwrap();
    InvoiceSort { key: InvoiceSortField::from(first_sort.key), desc: first_sort.desc }
  },
  _ => // Default value goes here
}
```

Untested, but something like that should do the job!

wlthomson (approved):
LGTM, only a few minors.

src/server/service/graphql/schema/queries/mod.rs
clemens-msupply: Agree, @andreievg rename it to pagination for the API or keep it as page?

src/server/service/graphql/schema/types/invoice_query.rs
clemens-msupply: Yes it works but I think its clearer what `remote` does when having the full path as a string. (The examples are also like that)

src/server/service/graphql/schema/types/invoices_query.rs
clemens-msupply: The default is chosen in the repo, i.e. it sorts by id

src/server/service/graphql/schema/queries/mod.rs
andreievg: Hmm, in terms of API, i was asked to keep to conventions: https://github.com/openmsupply/remote-server/pull/287#discussion_r700047229, beyond API, not sure pagination made more sense to me, up to you

src/server/service/graphql/schema/types/invoices_query.rs
clemens-msupply: why do you want to avoid flatten()? I also don't want to use unwrap

andreievg: Unwrapping option would panic, afaik.

src/server/service/graphql/schema/types/invoices_query.rs
clemens-msupply: done d6f1b41

src/server/service/graphql/schema/queries/mod.rs
clemens-msupply: Cool I leave it as it is since I think pagination is clearer...

src/database/repository/diesel/invoice_query.rs
```diff
@@ -38,16 +62,123 @@ impl InvoiceQueryRepository {
...
+                } else if let Some(after_or_equal) = &value.after_or_equal_to {
+                    query = query.filter(invoice_dsl::entry_datetime.ge(after_or_equal));
+                }
+            }
+            if let Some(value) = &f.confirm_datetime {
+                if let Some(eq) = &value.equal_to {
+                    query = query.filter(invoice_dsl::confirm_datetime.eq(eq));
+                } else if let Some(before_or_equal) = &value.before_or_equal_to {
+                    query = query.filter(invoice_dsl::confirm_datetime.le(before_or_equal));
+                } else if let Some(after_or_equal) = &value.after_or_equal_to {
```
andreievg, line 124: i think, especially for date time, we should allow multiple filters for the same field (so just if let instead of else if).

clemens-msupply: Makes sense, done e2fdceb

andreievg (approved):

src/server/service/graphql/schema/types/invoices_query.rs
wlthomson: Yeah, just an example which can be made more robust (I think it's probably reasonable to assume that if `Sort` isn't `None` then it will have a first value, but you can certainly add error handling for a default). I don't see the difference between `first` and `get(0)` in terms of safety (they both return `Option`)?
 
I don't mind `flatten` in general, I just think this is a good example of the sort of use-case match is designed for.

src/server/service/graphql/schema/types/invoices_query.rs
wlthomson: I mean, if `sort` is `None`? Is that handled by `as_ref`?

src/server/service/graphql/schema/types/invoice_query.rs
wlthomson: Hmm, I don't think the `crate::database` is needed for clarity (e.g. `schema::InvoiceRow` seems sufficiently clear, no?). It seems a bit incongruous with the style in the rest of the codebase.

src/server/service/graphql/schema/types/invoices_query.rs
wlthomson: Oh, you keep it `None` to trigger a default sort in the repo. Understand now, thanks.

src/server/service/graphql/schema/types/invoice_query.rs
clemens-msupply: Its IMHO the easiest way to do it. Copying/adding this string without autoimport is already annoying enough, splitting it into an importing statement and a string is even worse...

Issue #345, opened 2021-09-17
[TEST] - API tests


andreievg:
MW parent job: `[OMSBET] omSupply Back End - Tests`

### Description

Framework for testing API queries (individually and integration)

### Table

none

### Scope

<ins>Includes</ins>
* Add query test for individual queries
* Add place to have integrated tests for combined integrated queries

<ins>Excludes</ins>
* Any queries not in M1

### Tasks

- [ ] Break down tasks into digestible (by reviewer) units.
- [ ] MW job added

PR #396, opened 2021-10-06
#391 add customer invoice mutation types

### LINKED ISSUE #391: Add CustomerInvoice mutation types
Add `CustomerInvoice` graphql mutation types.

wlthomson:
Closes #391.

Adds outline of basic `CustomerInvoice` mutations (`CustomerInvoiceLines` to follow).

Have attached an Insomia collection for front-end team to experiment with: 
[391-add-customer-invoice-mutations.zip](https://github.com/openmsupply/remote-server/files/7290519/391-add-customer-invoice-mutations.zip)

Disclaimer: had to do a bit of git trickery as based this off #345 initially, but should be clean now!


src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
```diff
@@ -0,0 +1,52 @@
...
+    RecordDoesNotExistError(RecordDoesNotExistError),
+    ForeignKeyError(ForeignKeyError),
+}
+
+pub async fn delete_customer_invoice(
+    _ctx: &Context<'_>,
+    _input: DeleteCustomerInvoiceInput,
+) -> DeleteCustomerInvoiceResultUnion {
+    // TODO: add deletion logic.
+    DeleteCustomerInvoiceResultUnion::Error(DeleteCustomerInvoiceError {
```
wlthomson, line 47: Arbitrary, just wanted to demonstrate an error response.

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -0,0 +1,85 @@
+use async_graphql::Object;
+
+#[derive(Default)]
+pub struct OtherPartyNotASupplierError;
+
+#[Object]
+impl OtherPartyNotASupplierError {
```
wlthomson, line 7: Maybe overkill, but could add a trait for gql errors also (more for readability than anything else, I believe type-safety should already be provided by the `async-graphql` macros).

src/server/service/graphql/schema/mutations/customer_invoice/insert.rs
```diff
@@ -0,0 +1,69 @@
...
+    OtherPartyNotASupplierError(OtherPartyNotASupplierError),
+    DatabaseError(DatabaseError),
+}
+
+pub async fn insert_customer_invoice(
+    _ctx: &Context<'_>,
+    _input: InsertCustomerInvoiceInput,
+) -> InsertCustomerInvoiceResultUnion {
+    // TODO: add insert logic.
+    InsertCustomerInvoiceResultUnion::Ok(InsertCustomerInvoiceOk {
```
wlthomson, line 51: As above, just an arbitrary empty object to demo the shape.

src/server/service/graphql/schema/mutations/customer_invoice/update.rs
```diff
@@ -0,0 +1,77 @@
...
+    InvoiceDoesNotBelongToCurrentStoreError(InvoiceDoesNotBelongToCurrentStoreError),
+    CannotChangeInvoiceBackToDraftError(CannotChangeInvoiceBackToDraftError),
+    DatabasseError(DatabaseError),
+}
+
+pub async fn update_customer_invoice(
+    _ctx: &Context<'_>,
+    _input: UpdateCustomerInvoiceInput,
+) -> UpdateCustomerInvoiceResultUnion {
+    // TODO: add update logic.
```
wlthomson, line 58: As above, arbitary 😃.

src/server/service/graphql/schema/mutations/error.rs
```diff
@@ -0,0 +1,54 @@
+use async_graphql::Object;
+
+#[derive(Default)]
```
wlthomson, line 3: These are just general mutation errors that are likely to be shared. Some of them also generalise to queries... but can refactor later.

src/server/service/graphql/schema/mutations/requisition/insert.rs
```diff
@@ -0,0 +1,74 @@
+use crate::{
```
wlthomson, line 1: This mutation was annoying me, so just moved it out into its own scope.

src/server/service/graphql/schema/mutations/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+};
+
+use requisition::InsertRequisitionInput;
+
+use async_graphql::{Context, Object};
+
+pub struct Mutations;
+
+#[Object]
+impl Mutations {
```
wlthomson, line 19: Might seem a bit overkill, but I liked the ided of keeping the root mutations object nice and clean and readable.

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[Object]
+impl CannotDeleteFinalisedInvoiceError {
+    pub async fn id(&self) -> &str {
+        "CANNOT_DELETE_FINALISED_INVOICE_ERROR"
+    }
+
+    pub async fn description(&self) -> &str {
+        "Once finalised, an invoice cannot be deleted."
+    }
```
clemens-msupply, line 84: All these error definitions are quite verbose but I can't see a good way to make it more concise....

src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+#[derive(Union)]
+pub enum DeleteCustomerInvoiceResultUnion {
+    Ok(DeleteCustomerInvoiceOk),
+    Error(DeleteCustomerInvoiceError),
+}
+
+#[derive(SimpleObject)]
+pub struct DeleteCustomerInvoiceOk {
```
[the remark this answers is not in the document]
andreievg: I had it as a flat enum in queries (PR coming through Today, see below), as per second option by Josh (although I would have thought the combination of two would be used).

```JS
const {__typename: type, error, ...invoice} = data.insertCustomerInvoice
   
if (type === 'InvoiceNode'){
    return invoice
}

return handle(type, error)
```

Error can't be a flat enum though (since we can't use interface inside an enum, and thus interface has to go in `UpdateCustomerInvoiceError` object)

I guess either way is fine, a slight personal preference for flattened way.

Delete can just return string (not sure if we need to specify if it's invoice_id with shape of structure)

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -0,0 +1,85 @@
+use async_graphql::Object;
+
+#[derive(Default)]
+pub struct OtherPartyNotASupplierError;
+
+#[Object]
+impl OtherPartyNotASupplierError {
+    pub async fn id(&self) -> &str {
+        "OTHER_PARTY_NOT_A_SUPPLIER_ERROR"
```
andreievg, line 9: __typename already provide a concrete 'code', so not sure of benefit of ID ?

andreievg: i think `id` was meant to be the entity id ?

andreievg (approved):
Thanks will, looks good to me, a couple of comments only

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
wlthomson: Ah, right. Fair point. I'll remove them.

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub async fn id(&self) -> &str {
+        "NOT_A_CUSTOMER_INVOICE_ERROR"
+    }
+
+    pub async fn description(&self) -> &str {
+        "Invoice must be a customer invoice."
+    }
+}
+
+#[derive(Default)]
```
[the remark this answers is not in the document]
wlthomson: I actually can't remember the motivation now if I'm honest... I think it was just a good-practice rust thing rather than having any specific application here. I'll get rid of them, better to strip out anything that is a potential source of confusion.

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
wlthomson: We should probably add the equivalent of a localisation module for these sorts of things, but I think low priority at the moment.

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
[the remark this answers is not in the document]
wlthomson: f9e37de.

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
wlthomson: 442842b.

PR #397, opened 2021-10-06
395 tx repositories

clemens-msupply:
Initial PR for transaction support for Repositories.
- Add StorageConnectionManager class to get new connections that can be used for transactions as well as for single requests
- Update some basic Repositories to work with the new connections/transactions (all that are used in the SyncRepository
- Remove the SyncRepository and put the logic into the translation module

One problem I ran into: The Loaders need to be Sync but the open raw connection isn't which becomes a problem when using async functions in repositories (Rust can't figure out anymore that the connection is used exactly once). One solution is to wrap the connection into a Mutex. Another solution is to not use async functions in repositories. In this PR I went with the second approach (if we ever need an async backend I am sure we find a solution to move back to async)


src/database/repository/diesel/storage_connection.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::database::repository::RepositoryError;
+
+use diesel::{
+    connection::TransactionManager,
+    r2d2::{ConnectionManager, Pool},
+    Connection,
+};
+use futures_util::Future;
+
+pub struct StorageConnection {
```
andreievg, line 12: might be a good use of Deref trait

clemens-msupply: I will create a new branch because I am not 100% sure if we mean the same

PR #407, opened 2021-10-07
404 align name query with design decisions

andreievg:
Fixed: #404 

Add domain object, service and query with composition (of domain object). Also added generic types for error handling in queries and generic connector result

### Changes

* Move filters and sorts to domain area 
* Created service for name
* Changed repository to use domain object 
* Changed repository to allow for count to be filtered

Not that happy about both total count and nodes being queried (even if just one of them is requested in gql query, can work around it with look ahead, maybe in future PR)

### DIFF
--- a/src/database/repository/diesel/mod.rs
+++ b/src/database/repository/diesel/mod.rs
@@ -38,7 +38,7 @@ pub use master_list::MasterListRepository;
 pub use master_list_line::MasterListLineRepository;
 pub use master_list_name_join::MasterListNameJoinRepository;
 pub use name::NameRepository;
-pub use name_query::{NameQueryFilter, NameQueryRepository, NameQuerySort, NameQuerySortField};
+pub use name_query::NameQueryRepository;
 pub use requisition::RequisitionRepository;
 pub use requisition_line::RequisitionLineRepository;
 pub use sort_filter_types::*;
@@ -58,6 +58,12 @@ pub type DBBackendConnection = SqliteConnection;
 
 #[cfg(not(feature = "sqlite"))]
 pub type DBBackendConnection = PgConnection;
+
+#[cfg(feature = "sqlite")]
+pub type DBType = diesel::sqlite::Sqlite;
+
+#[cfg(not(feature = "sqlite"))]
+pub type DBType = diesel::sqlite::Pg;
 
 pub type DBConnection = PooledConnection<ConnectionManager<DBBackendConnection>>;
 
--- a/src/database/repository/diesel/name_query.rs
+++ b/src/database/repository/diesel/name_query.rs
@@ -1,26 +1,29 @@
-use super::{SimpleStringFilter, Sort, StorageConnection};
-
+use super::{DBType, StorageConnection};
 use crate::{
     database::{
         repository::RepositoryError,
         schema::{
             diesel_schema::{
-                name_store_join::dsl as name_store_join_dsl, name_table::dsl as name_table_dsl,
+                name_store_join, name_store_join::dsl as name_store_join_dsl, name_table,
+                name_table::dsl as name_table_dsl,
             },
             NameRow, NameStoreJoinRow,
         },
     },
-    server::service::graphql::schema::{
-        queries::pagination::{Pagination, PaginationOption},
-        types::NameQuery,
+    domain::{
+        name::{Name, NameFilter, NameSort, NameSortField},
+        Pagination,
     },
 };
 
-use diesel::prelude::*;
+use diesel::{
+    dsl::{IntoBoxed, LeftJoin},
+    prelude::*,
+};
 
 type NameAndNameStoreJoin = (NameRow, Option<NameStoreJoinRow>);
 
-impl From<NameAndNameStoreJoin> for NameQuery {
+impl From<NameAndNameStoreJoin> for Name {
     fn from((name_row, name_store_join_row_option): NameAndNameStoreJoin) -> Self {
         let (is_customer, is_supplier) = match name_store_join_row_option {
             Some(name_store_join_row) => (
@@ -30,7 +33,7 @@ impl From<NameAndNameStoreJoin> for NameQuery {
             None => (false, false),
         };
 
-        NameQuery {
+        Name {
             id: name_row.id,
             name: name_row.name,
             code: name_row.code,
@@ -38,20 +41,6 @@ impl From<NameAndNameStoreJoin> for NameQuery {
             is_supplier,
         }
     }
-}
-
-pub struct NameQueryFilter {
-    pub name: Option<SimpleStringFilter>,
-    pub code: Option<SimpleStringFilter>,
-    pub is_customer: Option<bool>,
-    pub is_supplier: Option<bool>,
-}
-
-pub type NameQuerySort = Sort<NameQuerySortField>;
-
-pub enum NameQuerySortField {
-    Name,
-    Code,
 }
 
 pub struct NameQueryRepository<'a> {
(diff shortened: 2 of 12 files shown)


src/database/repository/diesel/name_query.rs
```diff
@@ -1,26 +1,29 @@
...
 
-use diesel::prelude::*;
+use diesel::{
+    dsl::{IntoBoxed, LeftJoin},
+    prelude::*,
+};
 
 type NameAndNameStoreJoin = (NameRow, Option<NameStoreJoinRow>);
 
-impl From<NameAndNameStoreJoin> for NameQuery {
```
wlthomson, line 23: 👍 .

src/database/repository/diesel/name_query.rs
```diff
@@ -127,27 +89,55 @@ impl<'a> NameQueryRepository<'a> {
...
+            .load::<NameAndNameStoreJoin>(&self.connection.connection)?;
+
+        Ok(result.into_iter().map(Name::from).collect())
+    }
+}
+
+type BoxedNameQuery =
+    IntoBoxed<'static, LeftJoin<name_table::table, name_store_join::table>, DBType>;
+
+pub fn create_filtered_query(filter: Option<NameFilter>) -> BoxedNameQuery {
```
wlthomson, line 104: Shouldn't this be a method on the repository?

src/domain/name.rs
```diff
@@ -0,0 +1,24 @@
+use super::{SimpleStringFilter, Sort};
```
wlthomson, line 1: Really liking this! Very organised and clear data structures.

src/domain/mod.rs
```diff
@@ -0,0 +1,50 @@
+pub mod name;
```
wlthomson, line 1: Nice!

src/server/service/graphql/schema/queries/mod.rs
```diff
@@ -25,17 +28,20 @@ impl Queries {
...
         sort: Option<Vec<NameSortInput>>,
-    ) -> NameList {
-        NameList {
-            pagination: page,
-            filter,
-            sort,
-        }
+    ) -> NamesResponse {
+        let connection_pool = ctx.get_repository::<StorageConnectionManager>();
+        get_names(
```
wlthomson, line 38: Bit confused about the whole connection manager thing, but I thihk I get the idea (I still think I prefer the approach of having the pool directly passed via `ctx` and passing it as an argument to stateless repositories).

src/server/service/graphql/schema/types/mod.rs
```diff
@@ -35,6 +38,131 @@ pub use self::invoices_query::*;
 pub mod sort_filter_types;
 pub use self::sort_filter_types::*;
 
+/// Generic Connector
+#[derive(SimpleObject)]
+#[graphql(concrete(name = "NameConnector", params(NameNode)))]
+pub struct Connector<T: OutputType> {
```
wlthomson, line 44: Is there an advantage of this is over `async-graphql` connections (https://async-graphql.github.io/async-graphql/en/cursor_connections.html)? Since it's already done, we can just stick with this way.

src/server/service/graphql/schema/types/mod.rs
```diff
@@ -35,6 +38,131 @@ pub use self::invoices_query::*;
...
+    }
+}
+
+// Generic Errors
+pub struct DBError(pub RepositoryError);
+
+#[Object]
+impl DBError {
+    pub async fn description(&self) -> &'static str {
+        "Dabase Error"
```
wlthomson, line 121: Typo?

src/server/service/graphql/schema/types/mod.rs
```diff
@@ -35,6 +38,131 @@ pub use self::invoices_query::*;
...
+    }
+
+    pub async fn full_error(&self) -> String {
```
wlthomson, line 124: I would just go with something like `source`?

src/server/service/graphql/schema/types/name.rs
```diff
@@ -35,52 +37,51 @@ impl From<NameFilterInput> for NameQueryFilter {
...
+    pub async fn is_supplier(&self) -> bool {
+        self.name.is_supplier
+    }
+}
 
-        let filter = self.filter.clone().map(NameQueryFilter::from);
+#[derive(Union)]
+pub enum NamesResponse {
+    Error(ErrorWrapper<ConnectorErrorInterface>),
+    Response(Connector<NameNode>),
```
wlthomson, line 71: I will go with this way for the CI shape, too (I prefer having consistent nesting for success and error, but rather than discuss I'm happy to adopt this approach for expediency).

src/service/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+        SingleRecordError::DBError(error)
+    }
+}
+
+pub fn get_default_pagination(
+    pagination_option: Option<PaginationOption>,
+    max_limit: u32,
+    min_limit: u32,
+) -> Result<Pagination, ListError> {
+    let check_limit = |limit: u32| -> Result<u32, ListError> {
```
wlthomson, line 42: Generally, not a big fan of lambdas/anonymous functions unless they are going to be used as callbacks/arguments. Could we just have this as a top-level function?

src/service/name.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub fn get_names(
+    connection_manager: &StorageConnectionManager,
+    pagination: Option<PaginationOption>,
+    filter: Option<NameFilter>,
+    sort: Option<NameSort>,
+) -> Result<ListResult<Name>, ListError> {
+    let pagination = get_default_pagination(pagination, MAX_LIMIT, MIN_LIMIT)?;
+    let connection = connection_manager.connection()?;
+    let repository = NameQueryRepository::new(&connection);
```
wlthomson, line 22: Again, not sure of the benefit of having to instantiate and drop repositories all over the place, versus just keeping them static with methods which take the connection... but this seems to be the way we have chosen, so I'll go with it.

src/service/name.rs
```diff
@@ -0,0 +1,28 @@
...
+    },
+};
+
+use super::{get_default_pagination, i64_to_u32, ListError, ListResult};
+
+pub const MAX_LIMIT: u32 = 1000;
+pub const MIN_LIMIT: u32 = 1;
+
+pub fn get_names(
+    connection_manager: &StorageConnectionManager,
```
wlthomson, line 15: Isn't `connection_manager` being returned from a repository call higher up in the call-chain? If we are getting the `connection_manager` from a repo (as stated, I prefer just having the pool on `ctx`, but no biggie), can it not just be grabbed here instead of passed as a param?

wlthomson (approved):
In general, looking pretty good!

A few thoughts:

- Renaming `NameQuery` to `Name` makes things a lot more consistent/easier to understand 👍 .
- Any reason for retaining `NameQueryRepository` vs. `NameRepository`?
- Any reason for rolling our own connections instead of using the ones provided by `async-graphql`?

src/database/repository/diesel/mod.rs
```diff
@@ -59,6 +59,12 @@ pub type DBBackendConnection = SqliteConnection;
 #[cfg(not(feature = "sqlite"))]
 pub type DBBackendConnection = PgConnection;
 
+#[cfg(feature = "sqlite")]
+pub type DBType = diesel::sqlite::Sqlite;
+
+#[cfg(not(feature = "sqlite"))]
+pub type DBType = diesel::sqlite::Pg;
```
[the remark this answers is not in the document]
andreievg: Yeah, thanks, nice catch. I fixed this upstream somewhere

src/domain/name.rs
andreievg: thanks, but that's just rust analyzer helpers

src/server/service/graphql/schema/queries/mod.rs
andreievg: idk, both ways same same to me. And that's the way Clemens did it so didn't have any reason to push a changes request in his PR.

src/server/service/graphql/schema/types/mod.rs
andreievg: I didn't see how we can use async-graphql connection to do `page` based pagination.

src/service/mod.rs
andreievg: fair point: https://github.com/openmsupply/remote-server/issues/428, and i think i questions myself when doing this. I kind of liked this flexibility in node (keep code in scope), but yeah would be better as top level method

src/server/service/graphql/schema/types/sort_filter_types.rs
```diff
@@ -1,21 +1,53 @@
...
-    schema::{InvoiceRowStatus, InvoiceRowType},
+use crate::{
+    database::{
+        repository::{
+            DatetimeFilter as DatetimeFilterPrev, EqualFilter as EqualFilterPrev,
+            SimpleStringFilter as SimpleStringFilterPrev,
+        },
+        schema::{InvoiceRowStatus, InvoiceRowType},
+    },
+    domain::{DatetimeFilter, EqualFilter, SimpleStringFilter, Sort},
```
clemens-msupply, line 11: Do we need structs like EqualFilter to be defined in repo and domain? Or is that just work in progress?

src/service/name.rs
clemens-msupply: Additionally to what Andrei said, one core part of a Repo is actually that it is defined on an interface and that you can do polymorphism/dyn dispatch. While we are not doing it right now it would be much easier to do in the future.

src/service/name.rs
```diff
@@ -0,0 +1,28 @@
...
+        PaginationOption,
+    },
+};
+
+use super::{get_default_pagination, i64_to_u32, ListError, ListResult};
+
+pub const MAX_LIMIT: u32 = 1000;
+pub const MIN_LIMIT: u32 = 1;
+
+pub fn get_names(
```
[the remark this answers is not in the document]
andreievg: https://github.com/openmsupply/remote-server/issues/428

yes i agree, and thought this will be nice (don't have to pass through connection_manager), in additions to your other comments

src/server/service/graphql/schema/types/sort_filter_types.rs
andreievg: Hmm, not quite sure what you mean. They are currently all in the domain and in api, and repo uses domain versions when filtering (doing conversions when needed)

PR #408, opened 2021-10-07
405 align item query with design decisions

andreievg:
Fixes: #405 
Extending changes in [this pr](https://github.com/openmsupply/remote-server/pull/407), to apply to item query

Branch name is wrong, should save align item

### DIFF
--- a/src/domain/item.rs
+++ b/src/domain/item.rs
@@ -1,0 +1,24 @@
+use super::{EqualFilter, SimpleStringFilter, Sort};
+
+#[derive(PartialEq, Debug)]
+pub struct Item {
+    pub id: String,
+    pub name: String,
+    pub code: String,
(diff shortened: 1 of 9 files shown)


src/domain/item.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub name: String,
+    pub code: String,
+    // Is visible is from master list join
+    pub is_visible: bool,
+}
+#[derive(Clone)]
+pub struct ItemFilter {
+    pub name: Option<SimpleStringFilter>,
+    pub code: Option<SimpleStringFilter>,
+    /// If true it only returns ItemAndMasterList that have a name join row
```
clemens-msupply, line 15: Is the comment correct? "...only returns Items that ..."?

PR #409, opened 2021-10-07
406 align invoice query with design decisions

andreievg:
Fixes: #406
Extending changes in [this pr](https://github.com/openmsupply/remote-server/pull/407), to apply to item query

### Extra changes

* Relocated invoice_line type it a new folder (this should be a domain object I think)
* Removed some queries and mutation (they were relying on each other, and then on old InvoiceInputType)
* Remove deprecated pagination object
* Adding single node error interface and return


src/server/service/graphql/schema/types/invoice_query.rs
```diff
@@ -1,24 +1,78 @@
...
+pub struct InvoiceFilterInput {
+    pub name_id: Option<EqualFilterStringInput>,
+    pub store_id: Option<EqualFilterStringInput>,
+    pub r#type: Option<EqualFilterInput<InvoiceNodeType>>,
+    pub status: Option<EqualFilterInput<InvoiceNodeStatus>>,
+    pub comment: Option<SimpleStringFilterInput>,
+    pub their_reference: Option<EqualFilterStringInput>,
+    pub entry_datetime: Option<DatetimeFilterInput>,
+    pub confirm_datetime: Option<DatetimeFilterInput>,
+    pub finalised_datetime: Option<DatetimeFilterInput>,
```
clemens-msupply, line 46: Just for info, DatetimeFilterInput contains NaiveDateTime which is wrong it should be `DateTime<Utc>`. I fixed it in the tests PR which might cause a merge conflict...

clemens-msupply (approved):

src/server/service/graphql/schema/types/invoice_query.rs
andreievg: I'll make sure it's reflected in end result

src/domain/invoice.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub r#type: InvoiceType,
+    pub invoice_number: i32,
+    pub their_reference: Option<String>,
+    pub comment: Option<String>,
+    pub entry_datetime: NaiveDateTime,
+    pub confirm_datetime: Option<NaiveDateTime>,
+    pub finalised_datetime: Option<NaiveDateTime>,
+}
+#[derive(Clone)]
+pub struct InvoiceFilter {
```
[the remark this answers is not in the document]
andreievg: Hmm, not sure, I know what service would need to add to filter (when we come to do permissions, filter by store_id etc..), and there needs to be a translation between inputobject to them (and if those types are in repo, then we need to mix those layer when from is implemented). Can you comment further in https://github.com/openmsupply/remote-server/issues/428 if you feel like we need to move them to repo

PR #410, opened 2021-10-07
Improve invoices graphql test

clemens-msupply:


tests/graphql/invoices.rs
```diff
@@ -88,10 +102,32 @@ mod graphql {
...
-        assert_eq!(
-            body,
-            "{\"data\":{\"invoices\":{\"nodes\":[{\"id\":\"customer_invoice_a\",\"pricing\":{\"totalAfterTax\":3.0}},{\"id\":\"customer_invoice_b\",\"pricing\":{\"totalAfterTax\":7.0}},{\"id\":\"supplier_invoice_a\",\"pricing\":{\"totalAfterTax\":11.0}},{\"id\":\"supplier_invoice_b\",\"pricing\":{\"totalAfterTax\":15.0}}]}}}"
+        let expected_json_invoice_nodes = mock_invoices
+            .iter()
+            .map(|invoice| {
+                json!({
+                    "id": invoice.id.to_owned(),
+                    "pricing": {
+                      "totalAfterTax": &mock_invoice_lines.iter().fold(0.0, |acc, invoice_line| {
```
andreievg, line 111: maybe filter then fold

andreievg (commented):
I think this approach should work.

I just ran this with some tweaks to both expected data and query, expected data mismatch works very good, and can see issue straight away. Changing query (i.e. adding field that doesn't exist in schema) would require running test manually with debug or with logs, i think assert_json_eq! shows the first field that's missmatched, which is data, not error, I wonder if we can also show error response if test fails ?

PR #425, opened 2021-10-10
345 api tests clemens

clemens-msupply:
- Implement some gql test helpers
- Also adds some (incomplete) sort and filter tests.

Sorry it got a bit bigger when I added some more tests that needed more mock data (so I better stop here)...

One thing I noticed is that for time filters we might want to do ranges? this is currently not possible..


tests/graphql/invoice_query.rs
```diff
@@ -50,60 +49,114 @@ mod graphql {
...
                 .await
                 .unwrap();
         }
 
-        let repository_registry = RepositoryRegistry { repositories };
-        let loader_registry = LoaderRegistry { loaders };
+        let invoice = mock_customer_invoices()[0].clone();
+        let query = format!(
+            r#"{{            
+                invoice(id:\"{}\"){{
```
andreievg, line 68: maybe also variable.

tests/graphql/mod.rs
```diff
@@ -1,3 +1,90 @@
...
+
+use assert_json_diff::assert_json_eq;
+use serde_json::{json, Value};
+
 mod invoice_query;
 mod invoices;
+mod names;
 mod requisition;
+
+async fn run_gql_query(
```
andreievg, line 18: hmm does this need to be behind test feature tag ?

andreievg (approved):
Thanks, looks great

A couple of nit.

Also I think we can do date range by combining greater then and less then filter

src/database/mock/name_store_join.rs
```diff
@@ -0,0 +1,39 @@
+use crate::database::schema::NameStoreJoinRow;
```
wlthomson, line 1: It's super nice to see these mocks coming together so consistently!

src/database/repository/diesel/mod.rs
```diff
@@ -15,6 +15,7 @@ mod master_list_line;
 mod master_list_name_join;
```
wlthomson, line 15: This is probably not the best place to raise it... but wondering if we could just flatten out the `diesel` bit now? i.e. just `src/database/repository/mod.rs`?

src/database/repository/diesel/name_store_join.rs
```diff
@@ -0,0 +1,37 @@
...
+    schema::diesel_schema::name_store_join::dsl as name_store_join_dsl, schema::NameStoreJoinRow,
+};
+
+use diesel::prelude::*;
+
+pub struct NameStoreJoinRepository<'a> {
+    connection: &'a StorageConnection,
+}
+
+impl<'a> NameStoreJoinRepository<'a> {
```
wlthomson, line 14: A bit of a shame about sync needing to directly handle these records! Would have been nice to have repositories take domain objects and not worry about how relational modelling details... ah well.

tests/graphql/invoice_query.rs
```diff
@@ -50,60 +49,114 @@ mod graphql {
...
+                    id,
```
wlthomson, line 69: Are these `,` valid gql?

tests/graphql/invoice_query.rs
```diff
@@ -50,60 +49,114 @@ mod graphql {
...
+                        }}
+                    }}
+                }}            
+            }}"#,
+            invoice.id
+        );
 
-        let repository_registry = actix_web::web::Data::new(repository_registry);
-        let loader_registry = actix_web::web::Data::new(loader_registry);
+        let expected = json!({
```
wlthomson, line 84: Isn't the response shape going to be determined by our gql types? Could we not make them serializable and just use them to build the response (might be more type safe and less prone to typos and other niggly annoyances)?

tests/graphql/invoice_query.rs
clemens-msupply: Yes was before I implemented the variables part will create a new PR (think Will is merging some parts right now and better get this merged)

tests/graphql/invoice_query.rs
clemens-msupply: Seems to work can change though

src/database/repository/diesel/mod.rs
clemens-msupply: I would leave it for now. More than once I thought I don't need the repo pattern (the interface part) and it always paid off...

clemens-msupply:
> Also I think we can do date range by combining greater then and less then filter

At the moment there is only one filter object which makes it impossible... I created a ticket for further discussion: #429

Issue #419, opened 2021-10-10
Add supplier invoice delete mutation


andreievg:
https://github.com/openmsupply/remote-server/issues/349

PR #427, opened 2021-10-11
419 add supplier invoice delete mutation

andreievg:
Fixes: #419

Cary on with the same pattern as previous PR's. Extracted some methods as helpers

### DIFF
--- a/src/server/service/graphql/schema/mutations/supplier_invoice/mod.rs
+++ b/src/server/service/graphql/schema/mutations/supplier_invoice/mod.rs
@@ -1,3 +1,5 @@
+use crate::server::service::graphql::schema::types::Connector;
+use crate::server::service::graphql::schema::types::InvoiceLineNode;
 use crate::server::service::graphql::schema::types::NameNode;
 use async_graphql::*;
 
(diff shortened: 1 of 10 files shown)


src/server/service/graphql/schema/mutations/supplier_invoice/mod.rs
```diff
@@ -25,12 +27,16 @@ impl OtherPartyNotASupplier {
     }
 }
 
-pub struct CannotDeleteInvoiceWithLines;
+pub struct CannotDeleteInvoiceWithLines(Connector<InvoiceLineNode>);
 #[Object]
 impl CannotDeleteInvoiceWithLines {
     pub async fn description(&self) -> &'static str {
```
wlthomson, line 33: No need for `self` here, I don't think?

src/service/invoice/supplier_invoice/delete/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+
+pub fn delete_supplier_invoice(
+    connection_manager: &StorageConnectionManager,
+    id: String,
+) -> Result<String, DeleteSupplierInvoiceError> {
+    let connection = connection_manager.connection()?;
+    // TODO do inside transaction
+    validate(&id, &connection)?;
+
+    InvoiceRepository::new(&connection).delete(&id)?;
```
wlthomson, line 18: This is a bit of an example of why constructing repositories sometimes is a bit unnecessary, I think?
Compare this to something like: `InvoiceRepository.delete(&connection, id)`. Maybe just personal preference 🤷 .

wlthomson (approved):
LGTM.

src/service/invoice/supplier_invoice/delete/mod.rs
andreievg: Agree it's a case where it seems more ergonomic to have connection as parameter (only slightly though). I think we are pretty settled on this, at least until m3.

PR #452, opened 2021-10-13
348 ci insert delete

clemens-msupply:
Here is already the insert and delete part of the ticket.

- I introduced an uuid utils method (which I then don't use but I kept in anyway)
- There is a new transaction_sync() method for the StorageConnection, I plan to refactor that later (think we going fully sync right?)
- There are some validation checks that can be shared with SI. I plan to refactor that as well as soon as SI is settled. To make the reuse easier the common validation methods should return a bool or Option instead of a CommonError. 
- Due to the point above I inlined all delete checks for now. At least for these simple checks its makes things much easier to read.

### DIFF
--- a/src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
+++ b/src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
@@ -2,7 +2,7 @@ use crate::{
     server::service::graphql::schema::{
         mutations::{
             CannotDeleteInvoiceWithLines, CannotEditFinalisedInvoice, DeleteResponse,
-            InvoiceDoesNotBelongToCurrentStore, RecordDoesNotExist,
+            InvoiceDoesNotBelongToCurrentStore, NotACustomerInvoice, RecordDoesNotExist,
         },
         types::{DatabaseError, ErrorWrapper},
     },
@@ -22,6 +22,7 @@ pub enum DeleteCustomerInvoiceResponse {
 pub enum DeleteCustomerInvoiceErrorInterface {
     RecordDoesNotExist(RecordDoesNotExist),
     CannotEditFinalisedInvoice(CannotEditFinalisedInvoice),
+    NotACustomerInvoice(NotACustomerInvoice),
     InvoiceDoesNotBelongToCurrentStore(InvoiceDoesNotBelongToCurrentStore),
     CannotDeleteInvoiceWithLines(CannotDeleteInvoiceWithLines),
     DatabaseError(DatabaseError),
@@ -54,6 +55,9 @@ impl From<DeleteCustomerInvoiceError> for DeleteCustomerInvoiceResponse {
             }
             DeleteCustomerInvoiceError::DatabaseError(error) => {
                 OutError::DatabaseError(DatabaseError(error))
+            }
+            DeleteCustomerInvoiceError::NotACustomerInvoice => {
+                OutError::NotACustomerInvoice(NotACustomerInvoice {})
             }
         };
 
--- a/src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
+++ b/src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
@@ -1,38 +1,62 @@
-use crate::server::service::graphql::schema::mutations::error::DatabaseError;
-
-use super::{
-    CanOnlyEditInvoicesInLoggedInStoreError, FinalisedInvoiceIsNotEditableError,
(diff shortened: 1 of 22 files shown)


src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
```diff
@@ -1,38 +1,62 @@
...
-#[derive(SimpleObject)]
-pub struct DeleteCustomerInvoiceOk {
-    invoice_id: String,
+#[derive(Interface)]
+#[graphql(field(name = "description", type = "&str"))]
+pub enum DeleteCustomerInvoiceErrorInterface {
+    RecordDoesNotExist(RecordDoesNotExist),
+    CannotEditFinalisedInvoice(CannotEditFinalisedInvoice),
+    InvoiceDoesNotBelongToCurrentStore(InvoiceDoesNotBelongToCurrentStore),
+    CannotDeleteInvoiceWithLines(CannotDeleteInvoiceWithLines),
```
andreievg, line 26: I also had not a supplier invoice as a check and error on supplier invoice (can potentially consolidate both deletes and remove that check). Or add the check here (prefer the later, since if we need to auto delete invoice lines we would need to break them up or introduce complexity with branching logic)

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -81,7 +81,7 @@ pub struct OtherPartyNotACustomerOfThisStoreError(pub String);
 impl OtherPartyNotACustomerOfThisStoreError {
```
andreievg, line 81: for later, maybe common other party error

src/server/service/graphql/schema/mutations/customer_invoice/insert.rs
```diff
@@ -1,45 +1,95 @@
...
-
-#[derive(SimpleObject)]
-pub struct InsertCustomerInvoiceOk {
-    invoice: InvoiceNode,
+impl From<InsertCustomerInvoiceInput> for InsertCustomerInvoice {
+    fn from(input: InsertCustomerInvoiceInput) -> Self {
+        InsertCustomerInvoice {
+            id: input.id,
+            other_party_id: input.other_party_id,
+            status: input
```
andreievg, line 35: for late, as per your question, which resulted in a comment from Josh that we can set inserts as drafts

src/service/invoice/customer_invoice/insert/generate.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    Ok(result)
+}
+
+fn new_invoice_number() -> i32 {
+    // TODO Existing mSupply Mechanism for this
+    1
+}
+
+fn confirm_datetime(status: &InvoiceStatus, current_time: &NaiveDateTime) -> Option<NaiveDateTime> {
```
andreievg, line 44: for later, maybe some candidates for shared method

src/service/invoice/customer_invoice/insert/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+pub enum InsertCustomerInvoiceError {
+    OtherPartyCannotBeThisStore,
+    OtherPartyIdNotFound(String),
+    OtherPartyNotACustomerOfThisStore(String),
+    InvoiceAlreadyExists,
+    DatabaseError(RepositoryError),
+}
+
+impl From<RepositoryError> for InsertCustomerInvoiceError {
```
andreievg, line 22: form what i've seen of rust code, general practice is trait implementations are at the bottom, not a biggy, don't need to change

src/service/invoice/customer_invoice/insert/validate.rs
```diff
@@ -0,0 +1,59 @@
...
+    input: &InsertCustomerInvoice,
+    connection: &StorageConnection,
+) -> Result<(), InsertCustomerInvoiceError> {
+    use InsertCustomerInvoiceError::*;
+    let repository = NameQueryRepository::new(&connection);
+
+    let other_party_id = &input.other_party_id;
+    let mut result = repository.query(
+        Pagination::one(),
+        Some(NameFilter::new().match_id(other_party_id)),
```
andreievg, line 50: also check for is customer as per supplier_invoice/mod

src/server/service/graphql/schema/mutations/customer_invoice/delete.rs
clemens-msupply: I added the check

PR #457, opened 2021-10-14
348 ci update

clemens-msupply:


src/service/invoice/customer_invoice/update/generate.rs
```diff
@@ -0,0 +1,105 @@
...
+            id: stock_line.id.to_owned(),
+            item_id: stock_line.item_id.to_owned(),
+            store_id: stock_line.store_id.to_owned(),
+            batch: stock_line.batch.clone(),
+            pack_size: stock_line.pack_size,
+            cost_price_per_pack: stock_line.cost_price_per_pack,
+            sell_price_per_pack: stock_line.sell_price_per_pack,
+            available_number_of_packs: stock_line.available_number_of_packs
+                - invoice_line.number_of_packs,
+            total_number_of_packs: stock_line.total_number_of_packs - invoice_line.number_of_packs,
```
clemens-msupply, line 100: Do I actually need to check here that enough stock is available?

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -68,12 +68,12 @@ impl OtherPartyIdMissingError {
...
 }
 
-pub struct OtherPartyIdNotFoundError(pub String);
+pub struct OtherPartyIdNotFoundError;
 
 #[Object]
 impl OtherPartyIdNotFoundError {
     pub async fn description(&self) -> String {
-        format!("Other party with id '{}' not found.", self.0)
+        "Other party not found.".to_string()
```
andreievg, line 76: maybe return &'static str

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -89,3 +89,25 @@ impl OtherPartyNotACustomerError {
         &self.0
     }
 }
+
+pub struct NotACustomerInvoiceError;
+
+#[Object]
+impl NotACustomerInvoiceError {
+    pub async fn description(&self) -> String {
+        "Not a customer invoice.".to_string()
```
andreievg, line 98: maybe return &'static str

src/server/service/graphql/schema/mutations/customer_invoice/update.rs
```diff
@@ -1,51 +1,116 @@
...
+                OutError::OtherPartyDoesNotExists(OtherPartyIdNotFoundError {})
+            }
+            UpdateCustomerInvoiceError::OtherPartyNotACustomer(name) => {
+                OutError::OtherPartyNotACustomer(OtherPartyNotACustomerError(NameNode { name }))
+            }
+            UpdateCustomerInvoiceError::OtherPartyCannotBeThisStore => {
+                OutError::OtherPartyCannotBeThisStore(OtherPartyCannotBeThisStoreError {})
+            }
+            UpdateCustomerInvoiceError::InternalError(msg) => {
+                OutError::InternalError(InternalError(msg))
```
andreievg, line 107: Maybe static type ?

src/server/service/graphql/schema/types/invoice_query.rs
```diff
@@ -70,8 +70,13 @@ pub enum InvoiceNodeType {
 #[graphql(remote = "crate::domain::invoice::InvoiceStatus")]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")] // only needed to be comparable in tests
 pub enum InvoiceNodeStatus {
+    /// In DRAFT mode only the available_number_of_packs in a stock line gets updated when items are
+    /// added to the invoice.
     Draft,
+    /// When an invoice is CONFIRMED available_number_of_packs and total_number_of_packs gets
+    /// updated when items are added to the invoice.
```
andreievg, line 77: Nice description, can you say it's for customer invoice, i'll add supplier invoice description (which is slightly different)

src/service/invoice/customer_invoice/update/generate.rs
```diff
@@ -0,0 +1,105 @@
...
+};
+
+use super::UpdateCustomerInvoiceError;
+
+pub fn generate(
+    existing_invoice: InvoiceRow,
+    patch: UpdateCustomerInvoice,
+    connection: &StorageConnection,
+) -> Result<(Option<Vec<StockLineRow>>, InvoiceRow), UpdateCustomerInvoiceError> {
+    let should_create_batches = should_create_batches(&existing_invoice, &patch);
```
andreievg, line 18: should_update_batches

src/service/invoice/customer_invoice/update/generate.rs
```diff
@@ -0,0 +1,105 @@
...
+        .map(|line| line.item_id.to_owned())
+        .collect::<Vec<String>>();
+    let stock_lines = StockLineRepository::new(connection).find_many_by_item_ids(&item_ids)?;
+
+    let mut result = Vec::new();
+    for invoice_line in invoice_lines {
+        let stock_line = stock_lines
+            .iter()
+            .find(|stock_line| invoice_line.item_id == stock_line.item_id)
+            .ok_or(UpdateCustomerInvoiceError::InternalError(format!(
```
andreievg, line 86: do you think inner join of line with stock lines return would be useful ?

src/service/invoice/customer_invoice/update/generate.rs
andreievg: technically speaking available_number_of_packs should never exceed total_number_of_packs, so shouldn't need to check

src/service/invoice/customer_invoice/update/validate.rs
```diff
@@ -0,0 +1,96 @@
...
+    connection: &StorageConnection,
+) -> Result<InvoiceRow, UpdateCustomerInvoiceError> {
+    let invoice = check_invoice_exists(&patch.id, connection)?;
+
+    // check_store(invoice, connection)?; InvoiceDoesNotBelongToCurrentStore
+    check_invoice_type(&invoice)?;
+    check_invoice_finalised(&invoice)?;
+    check_invoice_status(patch, &invoice)?;
+    check_other_party(&patch.other_party_id, connection)?;
+
```
andreievg, line 25: i think it would be good to validate stock lines reduction below zero here. and pass along Vec<(InvoiceLineRow, StockLineRow)> upstream, to then be passed to generate

andreievg (approved):
Thanks

A few request, approving, would assume you still want to do the remainders in another PR.

One other request that's not in the comments, is that status may jump from draft to finalised, confirmed to finalised and draft to confirmed.

src/server/service/graphql/schema/mutations/customer_invoice/update.rs
andreievg: An alternative is inner join, mentioned below

src/server/service/graphql/schema/types/invoice_query.rs
andreievg: Or can relocate this to actual input type

src/service/invoice/customer_invoice/update/validate.rs
andreievg: Something that I overlooked, the reduction below zero on invoice should return something similar to delete error with lines I think.

andreievg: Ohh you are right sorry, i am getting slightly confused while thinking about lines update

src/service/invoice/customer_invoice/update/generate.rs
andreievg: yep, good point, just the total_number_of_packs

src/service/invoice/customer_invoice/update/generate.rs
clemens-msupply: Right, and in the case that there is a missing stock_line we just ignore the invoice_line without throwing an error?

clemens-msupply: I would leave it like that for now (query could be optimized though but not sure if its necessary).

Could rename InternalError to a general ResponseError (or similar name) which is a bucket for all errors that don't fit anywhere else and are not handled by the frontend anyway other than for debugging.

andreievg: Ok, API wise we can tweak to whatever front end needs. But why not use enums in service errors ? You are basically creating a constants with the string anyways

clemens-msupply: Ok will add an error for it to be consistent.

My thinking was that if we don't use the error we can merge them into a general error and don't need to have extra match branches if we have multiple errors like that...

src/server/service/graphql/schema/mutations/customer_invoice/error.rs
```diff
@@ -41,12 +41,12 @@ impl InvoiceDoesNotBelongToCurrentStoreError {
...
 }
 
-pub struct InvoiceNotFoundError(pub String);
+pub struct InvoiceNotFoundError();
 
 #[Object]
 impl InvoiceNotFoundError {
     pub async fn description(&self) -> String {
-        format!("Invoice with id '{}' not found.", self.0)
+        "Invoice not found.".to_string()
```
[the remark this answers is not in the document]
andreievg: you do (it's a comment you've made in my pr a few times, to which i replied that it's needed for async_graphql Object)

Issue #456, opened 2021-10-14
Add stock check and adjustments in customer invoice line mutation


andreievg:
https://github.com/openmsupply/remote-server/issues/350

PR #462, opened 2021-10-14
456 add adjustment and stock validation logic to customer invoice line mutation

andreievg:
Closes: #456 

Not 100% happy with the BatchPair implementation inside main mod, maybe it should be inside a helper module ?

### DIFF
--- a/src/server/service/graphql/schema/mutations/customer_invoice/line/mod.rs
+++ b/src/server/service/graphql/schema/mutations/customer_invoice/line/mod.rs
@@ -35,6 +35,20 @@ impl StockLineAlreadyExistsInInvoice {
     }
 }
 
+pub struct NotEnoughStockForReduction(pub String);
+#[Object]
+impl NotEnoughStockForReduction {
+    pub async fn description(&self) -> &'static str {
+        "Not enought stock for reduction"
+    }
+
+    pub async fn line(&self, ctx: &Context<'_>) -> InvoiceLineResponse {
+        let connection_manager = ctx.get_repository::<StorageConnectionManager>();
+
+        get_invoice_line(connection_manager, self.0.clone()).into()
+    }
+}
+
 pub struct LineDoesntReferenceStockLine;
 #[Object]
 impl LineDoesntReferenceStockLine {
--- a/src/service/invoice_line/customer_invoice_line/insert/generate.rs
+++ b/src/service/invoice_line/customer_invoice_line/insert/generate.rs
@@ -1,5 +1,5 @@
 use crate::{
-    database::schema::{InvoiceLineRow, InvoiceRow, ItemRow, StockLineRow},
+    database::schema::{InvoiceLineRow, InvoiceRow, InvoiceRowStatus, ItemRow, StockLineRow},
     domain::customer_invoice::InsertCustomerInvoiceLine,
     service::u32_to_i32,
 };
@@ -10,13 +10,31 @@ pub fn generate(
     input: InsertCustomerInvoiceLine,
     item_row: ItemRow,
     batch: StockLineRow,
-    _: InvoiceRow,
-) -> Result<InvoiceLineRow, InsertCustomerInvoiceLineError> {
+    invoice: InvoiceRow,
+) -> Result<(InvoiceLineRow, StockLineRow), InsertCustomerInvoiceLineError> {
+    let adjust_total_number_of_packs = invoice.status == InvoiceRowStatus::Confirmed;
+
+    let update_batch = generate_batch_update(&input, batch.clone(), adjust_total_number_of_packs);
     let new_line = generate_line(input, item_row, batch);
 
-    // Adjust stock line
+    Ok((new_line, update_batch))
+}
 
-    Ok(new_line)
+fn generate_batch_update(
+    input: &InsertCustomerInvoiceLine,
+    batch: StockLineRow,
+    adjust_total_number_of_packs: bool,
+) -> StockLineRow {
+    let mut update_batch = batch;
+
+    let reduction = u32_to_i32(input.number_of_packs);
+
+    update_batch.available_number_of_packs = update_batch.available_number_of_packs - reduction;
+    if adjust_total_number_of_packs {
+        update_batch.total_number_of_packs = update_batch.total_number_of_packs - reduction;
+    }
+
+    update_batch
 }
 
 fn generate_line(
--- a/src/service/invoice_line/customer_invoice_line/update/generate.rs
+++ b/src/service/invoice_line/customer_invoice_line/update/generate.rs
@@ -1,23 +1,76 @@
 use crate::{
-    database::schema::{InvoiceLineRow, InvoiceRow, ItemRow, StockLineRow},
+    database::schema::{InvoiceLineRow, InvoiceRow, InvoiceRowStatus, ItemRow, StockLineRow},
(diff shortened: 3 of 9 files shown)


andreievg:
> My main point was that I think validate and generate is not the right split here but works nicely for other cases. Or at least we shouldn't be dogmatic about a clean cut between validate and generate because otherwise you have to do things twice. Doing the range check in generate is IMHO easier and doesn't require the helper method to calculate the changes twice.

Hmm, are we worried about performance, i.e. it seems like `doing this twice` is the issue ?

Also it seems easier to locate the range check in code if it's calling a method that is named by the functionality (i don't think methods is a concept of keeping things dry, but as abstraction of functionality/complexity).

> a super method stays at least kind of readable... 

I guess I have a phobia of super methods: https://github.com/sussol/msupply/blob/master/Project/Sources/Methods/sync_process_sync_record.4dm.

Btw what do you think of the SLA concept ? I think getting clarity on your opinion about it would help our discussion on that front going forward.

Last note, to me, when reading/navigating code, I don't need to know every underlying condition in the context, only the conditions that I am interested in (and I can find those easier if i am presented with abstracted view, i drill down to methods that I need to modify), like in this example. If validations all pass correctly, but data does not match what is expected, I know that I need to just look in generate method, and within generate method it's quite clear what methods are responsible for which record/s)

src/server/service/graphql/schema/mutations/customer_invoice/line/mod.rs
```diff
@@ -35,6 +35,20 @@ impl StockLineAlreadyExistsInInvoice {
     }
 }
 
+pub struct NotEnoughStockForReduction(pub String);
+#[Object]
+impl NotEnoughStockForReduction {
+    pub async fn description(&self) -> &'static str {
```
wlthomson, line 41:
```suggestion
    pub async fn description() -> &'static str {
```

😉.

src/server/service/graphql/schema/mutations/customer_invoice/line/mod.rs
```diff
@@ -35,6 +35,20 @@ impl StockLineAlreadyExistsInInvoice {
...
+        "Not enought stock for reduction"
```
wlthomson, line 42:
```suggestion
        "Not enough stock for reduction"
```

Minor nit: I prefer "Insufficient stock".

src/service/invoice_line/customer_invoice_line/insert/generate.rs
```diff
@@ -10,13 +10,31 @@ pub fn generate(
     input: InsertCustomerInvoiceLine,
     item_row: ItemRow,
     batch: StockLineRow,
-    _: InvoiceRow,
-) -> Result<InvoiceLineRow, InsertCustomerInvoiceLineError> {
+    invoice: InvoiceRow,
+) -> Result<(InvoiceLineRow, StockLineRow), InsertCustomerInvoiceLineError> {
+    let adjust_total_number_of_packs = invoice.status == InvoiceRowStatus::Confirmed;
+
+    let update_batch = generate_batch_update(&input, batch.clone(), adjust_total_number_of_packs);
```
wlthomson, line 17: Personally, rather than passing a boolean, I would just pass the whole `invoice`. This way `generate_batch_update` is more decoupled from the caller and doesn't "leak" implementation details.

By passing the whole `invoice`, we don't reveal anything about the relationship between the status and the pack adjustment logic, separating the concerns more cleanly and giving `generate_batch_update` full responsibility and flexibility over how pack adjustments are calculated (see the Open-Closed and SRP principles).

wlthomson: If both functions need ownership of `batch`, I would explicity make clones in the caller for both of them. I think it's just a bit more explicit as to what's going on.

src/service/invoice_line/customer_invoice_line/insert/generate.rs
```diff
@@ -10,13 +10,31 @@ pub fn generate(
...
     let new_line = generate_line(input, item_row, batch);
 
-    // Adjust stock line
+    Ok((new_line, update_batch))
+}
+
+fn generate_batch_update(
+    input: &InsertCustomerInvoiceLine,
+    batch: StockLineRow,
```
wlthomson, line 25: Could you maybe avoid transferring ownership here and instead pass a reference here and do something like:

```
    let reduction = u32_to_i32(input.number_of_packs);

    let available_number_of_packs = batch.available_number_of_packs - reduction;

    let total_number_of_packs = if adjust_total_number_of_packs {
	update_batch.total_number_of_packs - reduction;
    } else {
        update_batch.total_number_of_packs;
    }

    let updated_batch = StockLineRow {
       ... // copy from batch
       available_number_of_packs,
       total_number_of_packs
    }

    updated_batch
```

src/service/invoice_line/customer_invoice_line/update/generate.rs
```diff
@@ -1,23 +1,76 @@
...
+) -> Result<(InvoiceLineRow, BatchPair), UpdateCustomerInvoiceLineError> {
+    let adjust_total_number_of_packs = invoice.status == InvoiceRowStatus::Confirmed;
 
-    // Adjust stock line
+    let batch_pair = BatchPair {
+        main_batch: generate_batch_update(
+            &input,
+            &existing_line,
+            &batch_pair,
+            adjust_total_number_of_packs,
```
wlthomson, line 23: Same as above, suggest just passing a reference to the `invoice` here.

wlthomson (approved):

src/service/invoice_line/customer_invoice_line/insert/generate.rs
andreievg: > Personally, rather than passing a boolean, I would just pass the whole invoice. This way generate_batch_update is more decoupled from the caller and doesn't "leak" implementation details.

yep, makes sense, added, https://github.com/openmsupply/remote-server/issues/484

> If both functions need ownership of batch, I would explicity make clones in the caller for both of them. I think it's just a bit more explicit as to what's going on.

 clone batch pair and pass that on ? Could do yeah: https://github.com/openmsupply/remote-server/issues/484

src/service/invoice_line/customer_invoice_line/insert/generate.rs
wlthomson: I was just thinking that you could avoid cloning by doing it this way. Full example:

```
fn generate_batch_update(
    input: &InsertCustomerInvoiceLine,
    batch: &StockLineRow,
    invoice: &InvoiceRow,
) -> StockLineRow {
    let reduction = u32_to_i32(input.number_of_packs);

    // Adjust available_number_of_packs
    let available_number_of_packs = batch.available_number_of_packs - reduction;

    // If invoice is confirmed, adjust total_number_of_packs
    let adjust_total_number_of_packs = invoice.status == InvoiceRowStatus::Confirmed;
    let total_number_of_packs = if adjust_total_number_of_packs {
	update_batch.total_number_of_packs - reduction;
    } else {
        update_batch.total_number_of_packs;
    }

    let updated_batch = StockLineRow {
       ... // copy from batch
       available_number_of_packs,
       total_number_of_packs
    }

    updated_batch
}
```

Note that now you don't need to clone batch at all, which I think is helpful to the reader as the signature shows that the existing batch is not touched by this function. If you think it's more complicated, then no worries :).

PR #470, opened 2021-10-15
464 ci insert test

clemens-msupply:
Closes #464

### DIFF
--- a/src/server/service/graphql/schema/mutations/customer_invoice/insert.rs
+++ b/src/server/service/graphql/schema/mutations/customer_invoice/insert.rs
@@ -4,7 +4,10 @@ use crate::{
         invoice::{Invoice, InvoiceStatus},
     },
     server::service::graphql::schema::{
-        mutations::{ForeignKey, ForeignKeyError, RecordAlreadyExist},
+        mutations::{
+            customer_invoice::OtherPartyIdNotFoundError, ForeignKey, ForeignKeyError,
+            RecordAlreadyExist,
+        },
         types::{DatabaseError, ErrorWrapper, InvoiceNodeStatus, InvoiceResponse, NameNode},
     },
     service::{invoice::InsertCustomerInvoiceError, SingleRecordError},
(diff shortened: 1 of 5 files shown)


src/server/service/graphql/schema/mutations/customer_invoice/insert.rs
```diff
@@ -75,7 +79,7 @@ impl From<InsertCustomerInvoiceError> for InsertCustomerInvoiceResponse {
                 OutError::OtherPartyCannotBeThisStore(OtherPartyCannotBeThisStoreError {})
             }
             InsertCustomerInvoiceError::OtherPartyIdNotFound(_) => {
-                OutError::ForeignKeyError(ForeignKeyError(ForeignKey::OtherPartyId))
+                OutError::OtherPartyIdNotFound(OtherPartyIdNotFoundError {})
```
andreievg, line 82: hmm why not use generic foreign key error ?

tests/graphql/mod.rs
```diff
@@ -85,6 +86,7 @@ async fn assert_gql_query(
     let expected_with_data = json!({
         "data": expected,
     });
+    println!("{}", actual);
```
andreievg, line 89: is this log left over, or do you want to keep it ?

andreievg: is it something that should be auto logged (be part of assert error message?)

andreievg (approved):
looks good thanks.

Would be great for errors in helper methods to show what they refer to in the actual test. I managed to break some test a few times, and it wasn't straight away apparent where (although it was pretty quick to find)

tests/graphql/mod.rs
clemens-msupply: thanks done 
8c32729 

think we would have to write our own macro to print the actual value when there is an error...

src/server/service/graphql/schema/mutations/customer_invoice/insert.rs
clemens-msupply: thanks missed that, thought ForeignKeyError is for something else...
3291ae6

clemens-msupply: I update the ForeignKeyError for the update endpoint as well, the error names might be a bit confusing/inconsitent though since there are the other OtherParty* errors...

PR #472, opened 2021-10-17
440 add test for update supplier invoice line mutation

andreievg:
Closes #440 

* Didn't like errors being thrown inside helper methods  so extracted them to macros
* Added mock data (stock lines being linked to supplier invoice lines only)
* Add to query and schema  .graphql files
* Add exhaustive check for update supplier invoice line mutation
* Fixed logic error

Realised that our optional nullable input is not quite correct (for updates of batch, expiry_date, comment and their_reference), will add an issue to fix


tests/graphql/schema.graphql
```diff
@@ -96,34 +107,59 @@ input EqualFilterInvoiceTypeInput {
...
 	status: InvoiceNodeStatus
 	comment: String
 	theirReference: String
 }
-type InsertCustomerInvoiceOk {
-	invoice: InvoiceNode!
+"""
+Generic Error Wrapper
+"""
+type InsertCustomerInvoiceLineError {
```
clemens-msupply, line 148: Is that the correct doc comment?

tests/graphql/common.rs
```diff
@@ -0,0 +1,60 @@
+use serde::{de::DeserializeOwned, Serialize};
+
+macro_rules! get_invoice_inline {
+    ($filter:expr, $connection:expr) => {{
+        remote_server::database::repository::InvoiceQueryRepository::new($connection)
+            .query(Pagination::one(), Some($filter), None)
+            .unwrap()
+            .pop()
+            .unwrap()
+    }};
```
clemens-msupply, line 10: Think the reason you are doing it is to get the correctly line when it fails?

Think macros are already a bit overused in Rust and I wonder if it can be done with a normal helper. For example, can we return `anyhow` results (anyhow crate) from the helper and then just unwrap them in the test?

andreievg: > Think the reason you are doing it is to get the correctly line when it fails?

Yep, and to remove clutter from tests.

> Think macros are already a bit overused in Rust

I am aware that some macros are too big and seem like magic ( proc_macros, that dynamically generate syntax tree/code), but I am not sure i agree that declarative macros are overused in rust (they typically just substitute parameters in code and are easy to follow). I think small macros that are easy to read and are fine (they don't seem like magic)

> For example, can we return anyhow results (anyhow crate) from the helper and then just unwrap them in the test?

Can you give an example ? You would still need to unwrap().pop().unwrap(), which is what i was trying to avoid with this helper.

tests/graphql/schema.graphql
andreievg: Not sure what it should be? I can't find any way to comment on concrete type in simple object that derive concrete type, so the naming should be generic, so idk, happy to change though

tests/graphql/common.rs
```diff
@@ -0,0 +1,60 @@
...
+    ($invoice_id:expr, $connection:expr) => {{
+        remote_server::database::repository::InvoiceLineRepository::new($connection)
+            .find_many_by_invoice_id($invoice_id)
+            .unwrap()
+    }};
+}
+
+macro_rules! assert_unwrap_enum {
+    ($enum:ident, $variant:path) => {{
+        let debug = format!("{:#?}", $enum);
```
wlthomson, line 23: Personally, I think we should avoid pretty-printing in general as it seems to often lead to lots of whitespace and formatting characters. If the user wants to pretty-print a particular log, it's pretty easy to do so using a unix command or log analyser or just the ol' copy-and-paste into an editor.

tests/graphql/common.rs
```diff
@@ -0,0 +1,60 @@
...
+
+        if let $variant(result) = $enum {
+            result
+        } else {
+            panic!("\ncannot unwrap {} from\n {}", stringify!($variant), debug);
```
wlthomson, line 28: Are (both) these newlines needed (assuming no pretty-printing)?

tests/graphql/common.rs
```diff
@@ -0,0 +1,60 @@
...
+            panic!(
+                "\ncannot unwrap Some(value) of key '{}' in\n {}",
+                stringify!($key),
+                debug
+            );
+        }
+    }};
+}
+
+pub fn convert_graphql_client_type<FROM, TO>(f: FROM) -> TO
```
wlthomson, line 49: I've never seen multi-character generic parameters, I think sticking with convention where possible is good (maybe just `F` and `T`)? I think the `where` clause makes it pretty clear what each one represents.

tests/graphql/common.rs
andreievg: it does make it look better, imo

tests/graphql/common.rs
andreievg: added to: https://github.com/openmsupply/remote-server/issues/484, although if generic do get ambiguous i would insist to use more explanatory generic, in favour of readability over loyalty to predefined conventions

PR #475, opened 2021-10-17
474 finalise supplier invoice mutations

### LINKED ISSUE #474: Finalise supplier invoice mutations
https://github.com/openmsupply/remote-server/issues/435

Apply existing pattern to mutations.

Test stock lines are created when status is changes in update supplier invoice.

CannotChangeInvoiceBackToDraft ?

andreievg:
Closes #474 

* Refactored supplier invoice insert/update/delete to use helpers from line mutations tests
* Added more specific check in delete, to confirm correct lines are returned in `CannotDeleteInvoiceWithLines`
* Added check to confirm stock line is generated correctly and linked to invoice lines when supplier invoice status is updated (from draft).
* Fix logic error for the last bulletpoint

### DIFF
--- a/tests/graphql/common.rs
+++ b/tests/graphql/common.rs
@@ -18,6 +18,16 @@ macro_rules! get_invoice_lines_inline {
     }};
 }
 
+macro_rules! get_name_inline {
+    ($filter:expr, $connection:expr) => {{
+        remote_server::database::repository::NameQueryRepository::new($connection)
+            .query(Pagination::one(), Some($filter), None)
+            .unwrap()
+            .pop()
+            .unwrap()
+    }};
+}
+
 macro_rules! assert_unwrap_enum {
     ($enum:ident, $variant:path) => {{
         let debug = format!("{:#?}", $enum);
@@ -71,8 +81,20 @@ where
     serde_json::from_str(&serde_json::to_string(&f).unwrap()).unwrap()
 }
 
+pub fn compare_option<A, B>(a: &Option<A>, b: &B) -> bool
+where
+    B: PartialEq<A>,
+{
+    if let Some(a) = a {
+        b == a
+    } else {
+        true
+    }
+}
+
 pub(crate) use assert_matches;
 pub(crate) use assert_unwrap_enum;
 pub(crate) use assert_unwrap_optional_key;
 pub(crate) use get_invoice_inline;
 pub(crate) use get_invoice_lines_inline;
+pub(crate) use get_name_inline;
--- a/tests/graphql/delete_supplier_invoice.rs
+++ b/tests/graphql/delete_supplier_invoice.rs
@@ -1,17 +1,57 @@
 mod graphql {
+    use crate::graphql::common::{assert_matches, get_invoice_lines_inline};
+    use crate::graphql::common::{
+        assert_unwrap_enum, assert_unwrap_optional_key, get_invoice_inline,
+    };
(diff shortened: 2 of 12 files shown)


tests/graphql/common.rs
```diff
@@ -71,8 +81,20 @@ where
...
 }
 
+pub fn compare_option<A, B>(a: &Option<A>, b: &B) -> bool
+where
+    B: PartialEq<A>,
+{
+    if let Some(a) = a {
+        b == a
+    } else {
+        true
```
wlthomson, line 91: Should this be `false` (if `a` is `None`, then it is not equal to `b`)?

tests/graphql/common.rs
```diff
@@ -71,8 +81,20 @@ where
     serde_json::from_str(&serde_json::to_string(&f).unwrap()).unwrap()
 }
 
+pub fn compare_option<A, B>(a: &Option<A>, b: &B) -> bool
```
wlthomson, line 84: I think this might be a bit over-cooked? Isn't this already pretty tidy to do with a one-liner? Something like `a.map_or(false, |a| a == b)`?

tests/graphql/delete_supplier_invoice.rs
```diff
@@ -1,168 +1,166 @@
...
-                        ),
-                    },
-                ),
-        };
-        assert_eq!(response.data.unwrap(), expected);
+
+        assert_error!(
+            response,
+            NotASupplierInvoice(delete::NotASupplierInvoice {
+                description: "Invoice is not Supplier Invoice".to_string(),
```
wlthomson, line 106:
```suggestion
                description: "Invoice is not a Supplier Invoice".to_string(),
```

wlthomson (approved):

tests/graphql/common.rs
andreievg: if you do a global search of where it's used:

https://github.com/openmsupply/remote-server/blob/a6c7574ef0d8b1c868c4dc03f6d6e2634ac0f66e/tests/graphql/update_supplier_invoice_line.rs#L411-L420

https://github.com/openmsupply/remote-server/blob/a6c7574ef0d8b1c868c4dc03f6d6e2634ac0f66e/tests/graphql/update_supplier_invoice_line.rs#L437-L448

And also in update_supplier_invoice

I think it adds a good utility when used like this repetitively

Also how would it look for references (you showed owned example)

tests/graphql/delete_supplier_invoice.rs
andreievg: added: https://github.com/openmsupply/remote-server/issues/484

tests/graphql/common.rs
wlthomson: I can see how it reduces the density of this code, but I think it also is a bit over-cooked in that it is a function dedicated to performing an operation that Rust already provides sufficient primitives to do pretty cleanly. 

It could probably be done better, but taking `item_id_option` from `update_supplier_invoice` as an arbitrary example:

```
&& item_id_option.as_ref().map_or(false, |item_id| item_id == &self.item_id)
```

Tried and seems to work the same as your helper, except returns falsey if the option is `None` (this behaviour can be reversed if desired by changing `false` to `true`). It's slightly more verbose, but I think it's more understandable and probably good to use existing primitives if they exist (even if it means a few more lines of code).

PR #488, opened 2021-10-19
Ignore all item with foreign key errors durring sync (for now)

clemens-msupply:


src/util/sync/translation/mod.rs
```diff
@@ -144,6 +146,19 @@ pub async fn import_sync_records(
     Ok(())
 }
 
+fn ignore_foreign_key_error<T: Debug>(
+    result: Result<(), RepositoryError>,
+    item: T,
+) -> Result<(), RepositoryError> {
+    match result {
+        Err(RepositoryError::ForeignKeyViolation) => {
+            println!("Failed to import (ignore): {:?}", item);
```
wlthomson, line 155: Suggest using `log::warning` for this.

src/util/sync/translation/mod.rs
```diff
@@ -153,23 +168,31 @@ async fn store_integration_records(
...
-                    IntegrationUpsertRecord::Store(record) => {
-                        StoreRepository::new(con).upsert_one(record)?
-                    }
-                    IntegrationUpsertRecord::MasterList(record) => {
-                        MasterListRepository::new(con).upsert_one(record)?
-                    }
-                    IntegrationUpsertRecord::MasterListLine(record) => {
-                        MasterListLineRepository::new(con).upsert_one(record)?
-                    }
+                    IntegrationUpsertRecord::Name(record) => ignore_foreign_key_error(
```
wlthomson, line 171: I find this pattern (not just here, have seen it around the codebase) a bit detrimental for code clarity. If a function doesn't have utility outside of another function, I think adding helper functions outside of that functions scope tends to make the code more complex rather than less (I understand the DRY impulse, but I think sometimes it should be resisted). Maybe a compromise would be to define functions like this as lambdas within the parent function scope?

src/util/sync/translation/mod.rs
```diff
@@ -6,6 +6,8 @@ mod name;
 mod store;
 pub mod test_data;
 
+use std::fmt::Debug;
```
wlthomson, line 9: Is this used?

wlthomson (changes requested):

src/util/sync/translation/mod.rs
wlthomson: My question would be: what is this function providing us?

I personally can't really see what it provides except for reducing a bit of repetition (and not really that much, as shown by the need for a generic parameter)? 

From personal experience, I know it can be easy to get caught up in DRY-mania and end up over-engineering a simple piece of logic and distributing it throughout a file or even different files. I think this is a good example of a good principle (trying to reduce repetition) being applied a bit over-zealously. The outcome (IMO), is less concise/maintainable code. 

I would argue it almost doesn't matter whether you or I understand it, but whether someone else at Sussol who wanted to contribute would find this easier or harder to maintain or work with.

Personally, I think functions should be used for encapsulating responsibilities - if the user has to read the function to understand the parent function calling it, then what we've done is not encapsulation or simplification, but rather we have effectively spread logic across two places. It might reduce some lines of code (which is good, don't get me wrong), but IMO isn't worth the cost of needing to look in two places to understand what's going on. 

This is only a small example that I'm picking on, but I think reflects a bigger pattern that I've started to see emerging in the codebase.

All this is just my opinion, if others disagree, that's fine :).

src/util/sync/translation/mod.rs
andreievg: good point, will update it in your branch after resolving conflicts

PR #502, opened 2021-10-20
363 sync postgres subtransactions

clemens-msupply:


src/util/sync/translation/mod.rs
```diff
@@ -194,13 +194,18 @@ async fn store_integration_records(
...
     con.transaction(|con| async move {
         for record in &integration_records.upserts {
-            match integrate_record(record, con) {
-                Ok(_) => {}
-                Err(RepositoryError::ForeignKeyViolation(extra)) => {
-                    warn!("Failed to import (ignore): {:?}, error {}", record, extra);
+            let sub_result = con.transaction_sync(|sub_tx| integrate_record(record, sub_tx));
+            match sub_result {
+                Ok(_) => Ok(()),
+                Err(TransactionError::Inner(err @ RepositoryError::ForeignKeyViolation(_))) => {
```
andreievg, line 200: i think it's useful to log inner part of ForeginKeyViolation (as it actually tells us what fields are experiencing problems), would be good to add it to warn. In general I think we should be logging more.

andreievg:
> Postgres always aborts the whole transaction if there is a DB error. For this reason I wrapped every update in a sub transaction. Could be a performance issue in production but probably the easiest way for now to see where the problems are.

I think it should be fine, I think we should be aware of performance for API, sync itself doesn't need to be optimised

src/util/sync/translation/mod.rs
clemens-msupply: I tried to print more details but Diesel unfortunately doesn't provides more for ForeignKeyViolation so I just print our error message (which basically says the same)...

Or did you try it? maybe its different for Postgres and sqlite...

src/database/repository/diesel/mod.rs
```diff
@@ -104,7 +104,11 @@ impl From<DieselError> for RepositoryError {
             DieselError::AlreadyInTransaction => {
                 Error::as_db_error("DIESEL_ALREADY_IN_TRANSACTION", "")
             }
-            DieselError::__Nonexhaustive => Error::as_db_error("DIESEL_UNKNOWN", ""),
+            DieselError::__Nonexhaustive => {
+                // get a more detailed diesel msg:
+                let diesel_msg = format!("{}", err);
```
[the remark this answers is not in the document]
clemens-msupply: Ok thanks, I think I got it wrong how the code changed. I removed the _Nonexhaustive clause. ad119bd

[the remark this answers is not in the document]
andreievg: ok, although i don't think it does anything, and can't see any indication of it in diesel code

src/util/sync/translation/mod.rs
andreievg: btw i am just suggesting something like this:
```rust
  Err(TransactionError::Inner(err @ RepositoryError::ForeignKeyViolation(extra))) => {
       warn!("Failed to import ({}): {:?}, extra {}", err, record, extra);
                    Ok(())
                }
  }
```

also to me one `return Err(SyncImportError::as_integration_error(error, record))` seems easier to follow then Ok(()) return then map_error and the question mark

andreievg (dismissed):

src/util/sync/translation/mod.rs
clemens-msupply: Thanks, doing all error handling in the match now (think that was what you meant, right?)

PR #503, opened 2021-10-20
Add minor doc updates for how_to_m1

joshxg:
Couple of minor mistakes I made while integrating my mSupply instance with the remote-backend. Temporary issues that shouldn't be a problem in the future. I am assuming this document will also be removed in the short term also so haven't done any other changes.

Sorry about the diff 🙄  formatting when I save in vscode 🤷  - I'll get it sorted for the future. Just added:

```
- Ensure you're on the mSupply branch: `feature/sync-v5`.
- Ensure your data file has run the migration code in the mSupply method `centralChangeLogGenerateRecords()`. It is possible the `feature/sync-v5` method is behind `main` which may result in your data file being upgraded without the sync-v5 code having being run.
```

### DIFF
--- a/docs/content/docs/M1/how_to_m1.md
+++ b/docs/content/docs/M1/how_to_m1.md
@@ -14,13 +14,16 @@ toc = true
 
 ## Setup
 
-[readme](https://github.com/openmsupply/remote-server#remote-server) 
+[readme](https://github.com/openmsupply/remote-server#remote-server)
 
 ##### Pull Main
+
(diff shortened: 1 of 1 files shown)


andreievg:
> I am assuming this document will also be removed in the short term also so haven't done any other changes.

yep, it will be removed.

joshxg:
> Thanks Josh, I completely forgot about the V5 bit of mSupply

I did know about it .. at some point..! I had forgotten also :)

PR #522, opened 2021-10-28
#preference docs

joshxg:
Do the zola docs get auto-built from develop or main? I just branched from main without thinking, sorry.

This is just a draft docs for [preferences]. I've added a new folder for preferences specifically. Thought this was enough to PR and get feedback before I did anything else.


craigdrown (approved):
I mean, why not?

Issue #509, opened 2021-10-22
All mutations in one


andreievg:
- [x] add Inbound/Outbound batch mutations
- [ ] add tests

I did a quick check, just to see type things for batchedRequest (graphql-request):

* All types have to be specified in the call to that method (so can build dynamicly but only for one particular record type), i.e. `batchedRequest<InsertSupplierInvoiceLine[]>(...)`
* Async_graphql doesn't support batching (i thought that batching means listing multiple queries in one 'batch', with aliases ), but it's actually putting them in seperate arrays i.e.

What i though (actual post payload), `...` mean anything (lazy to type exact type out), 
```
{
 "query":  "mutations { 
      firstMutation: insertSupplierInvoiceLine ( ... ) { ... }, 
      secondMutation: insertSupplierInvoiceLine ( ... ) { ... }
       }",
 "variables": { ... }
}
```

What it is 

```
[
    {
    "query":  "mutations { 
        fnsertSupplierInvoiceLine ( ... ) { ... }
        }",
    "variables": { ... }
    },
    {
    "query":  "mutations { 
        fnsertSupplierInvoiceLine ( ... ) { ... }
        }",
    "variables": { ... }
    },
    {
    "query":  "mutations { 
        fnsertSupplierInvoiceLine ( ... ) { ... }
        }",
    "variables": { ... }
    }
]
```

### Proposed solution

This might sound a bit weird, but we can have one mutation, which is a combination of all mutations, optional array inputs. I think it's quite a good solution (very little extra work on back end). Check out `all-mutations-example` branch for types

```graphql
type AllMutations {
  insertCustomerInvoice: [InsertCustomerInvoiceResponseWithId!]
  insertSupplierInvoice: [InsertSupplierInvoiceResponseWithId!]
  insertCustomerInvoiceLine: [InsertCustomerInvoiceLineResponseWithId!]
  insertSupplierInvoiceLine: [InsertSupplierInvoiceLineResponseWithId!]
  updateCustomerInvoiceLine: [UpdateCustomerInvoiceLineResponseWithId!]
  updateSupplierInvoiceLine: [UpdateSupplierInvoiceLineResponseWithId!]
  deleteCustomerInvoiceLine: [DeleteCustomerInvoiceLineResponseWithId!]
  deleteSupplierInvoiceLine: [DeleteSupplierInvoiceLineResponseWithId!]
  updateCustomerInvoice: [UpdateCustomerInvoiceResponseWithId!]
[...]

PR #524, opened 2021-10-28
Remove all From<Result> traits

### LINKED ISSUE #523: Remove From<Result>
From<Result> complicates navigating to implementations in API queries, remove From<Result> (replace with helper methods or inline match where appropriate)

Part of: https://github.com/openmsupply/remote-server/issues/428

andreievg:
Closes #523 

As per commit description.

Actually ended up being less lines and easier to navigate to implementation

Wanted to do this before jumping into: https://github.com/openmsupply/remote-server/issues/509


clemens-msupply (approved):
Cool looks good!

Issue #526, opened 2021-10-28
Add on hold for invoice


andreievg:
As per request issue, on hold field on invoice and on mutations, when invoice is on hold, don't allow status change

https://github.com/openmsupply/remote-server/issues/512

PR #527, opened 2021-10-29
526 add invoice on hold

andreievg:
Closes: #526 

As per commits. Had to add a few more test condition, since on_hold logic is a little bit tricky, also relocates status check in service to one common method.

I still used the same from pattern for service, haven't had the time yet to do a draft PR with new suggested error mapping.


clemens-msupply (approved):
Looks all good

PR #529, opened 2021-10-29
Add support for user logout

clemens-msupply:
Currently all user tokens are invalidated on logout. Tokens are stored in memory but there is a trait to have a different backend.

Not sure if I am going to far here, e.g. some open questions:
- should we really invalidate all tokens if a user logs out or just the token for a certain client
- after a token refresh, should the expiry time for old tokens be reduced, e.g. down to 5min

### DIFF
--- a/src/service/token_bucket.rs
+++ b/src/service/token_bucket.rs
@@ -1,0 +1,92 @@
+use std::collections::{hash_map::Entry, HashMap};
+
+use chrono::Utc;
+
+use crate::util::auth::sha256;
+
+/// Tracks if a token is valid or if a user has been logged out
+pub trait TokenBucket {
+    /// Checks if the token is known for the given user
+    fn contains(&self, user_id: &str, token: &str) -> bool;
+    /// Adds a token for a given user.
+    /// If token is already known the expiry_date is updated.
+    /// This can be used to reduce the expiry date of a token on the server
+    fn put(&mut self, user_id: &str, token: &str, expiry_date: usize);
+    /// Removes all known tokens for a given user
+    fn clear(&mut self, user_id: &str);
+}
+
+struct TokenInfo {
+    token_hash: String,
+    expiry_date: usize,
+}
+
+pub struct InMemoryTokenBucket {
+    users: HashMap<String, Vec<TokenInfo>>,
+}
+
+impl InMemoryTokenBucket {
+    pub fn new() -> Self {
+        InMemoryTokenBucket {
+            users: HashMap::new(),
+        }
+    }
+}
+
+fn token_hash(token: &str) -> String {
+    sha256(token)
+}
+
+impl TokenBucket for InMemoryTokenBucket {
+    fn contains(&self, user_id: &str, token: &str) -> bool {
+        let user_tokens = match self.users.get(user_id) {
+            Some(value) => value,
+            None => return false,
+        };
+
+        let token_hash = token_hash(token);
+        let existing_token = match user_tokens
+            .iter()
+            .find(|item| item.token_hash == token_hash)
+        {
+            Some(value) => value,
+            None => return false,
+        };
+
+        let now = Utc::now().timestamp() as usize;
+        existing_token.expiry_date >= now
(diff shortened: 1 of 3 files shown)


src/service/token_bucket.rs
```diff
@@ -0,0 +1,90 @@
...
+        let existing_token = match user_tokens
+            .iter()
+            .find(|item| item.token_hash == token_hash)
+        {
+            Some(value) => value,
+            None => return false,
+        };
+
+        let now = Utc::now().timestamp() as usize;
+        existing_token.expiry_date >= now
```
andreievg, line 55: Just wondering, how would this every be the case, i thought that tokens are validated (expiry is checked at that point)

src/service/token_bucket.rs
```diff
@@ -0,0 +1,90 @@
...
+            Entry::Occupied(o) => o.into_mut(),
+            Entry::Vacant(v) => v.insert(Vec::new()),
+        };
+
+        // clean up expired tokens
+        user_tokens.retain(|item| item.expiry_date > now);
+
+        // update existing
+        let token_hash = token_hash(token);
+        if let Some(value) = user_tokens
```
andreievg, line 73: i kind of like the match syntax here (i guess personal style, but i think it's easier to see then the return statement)

```rust 
match found_token {
  Some(value) => value.expiry_date = expiry_date,
  None => { 
   user_tokens.push(TokenInfo {
            token_hash,
            expiry_date,
        });
   }
}
```

src/service/token_bucket.rs
```diff
@@ -0,0 +1,90 @@
...
+            .iter_mut()
+            .find(|item| item.token_hash == token_hash)
```
andreievg, line 75: I thought that tokens expiry date inside the token always changes, so you'll never get the same token/hash ?

andreievg (commented):
I am not quite sure about the use of TokenBucket in this case, I might be missing something though.

* On server restart everyone needs to re-login
* We should be able to trust the JWT mechanism, so don't really need to re-check expiry date, or maintain list of tokens (since JWT mechanism guarantees that they are valid)
* We do need a way to flag user as needing to re-login, or invalidate their token, but we can just do that by having an array of user_ids that need an immediate token refresh (and if we store both of the tokens in cookies, we can auto refresh their tokens while an API query is happening)

andreievg:
Also, for logging out, front end can overwrite tokens, even http only tokens

andreievg (approved):
Thanks looks good, although would be great to depict this in a diagram of some sort, (we can do that towards the end i guess).

I have some suggestions for future:

* each token to have session_id (which is what can be used to identify it in the hash map, rather then by full contents of token)

And I am still unsure about the use cases (even after our convo), to me the only use case that required maintaining in memory map of sessions/tokens, is when user permissions have changed, and we need to immediately ask front end to refresh the token, which means:

* every user token (across all clients) to be refreshed
* so we need to `maintain a list of active tokens` for user, and flag them all for immediate refresh (which would mean we send some error message on next operation, so that front end can immediately do the refresh token thing, or auto refresh it ourselves, but that's only possible if API token is also in cookies)
* `maintain a list of active tokens`, would mean we have session_id in both tokens, and when refresh token is used, we just set time to live on the tokens with previous id to like a couple of seconds, and put new tokens with new session_id in the list. 

So TokenInfo would be something like:

```rust
struct TokenInfo {
  session_id: String,
  time_to_live: Option<i32>,
  requires_immidate_refresh: bool,
}
```

clemens-msupply:
thanks for the feedback. Where should the diagram go? in the repo? you mean for the overall auth documentation, right? but yes maybe wait a bit more till we finalized the auth flow.

Yes a session id would also be good to simplify a session logout. 

For permission changes, maybe we just flag old tokens that their permissions are outdated and if there is a request from such a token we could do the authorization check using the permissions from our DB rather than the permissions from the token. This would avoid some complicated error handling for the client, e.g. when the permissions change durring some complex client operations.

PR #533, opened 2021-11-03
Add some docs and code refactoring

clemens-msupply:
This should have been part of the previous PR but got interrupted and missed to commit it...

### DIFF
--- a/src/service/token_bucket.rs
+++ b/src/service/token_bucket.rs
@@ -13,7 +13,12 @@ fn token_hash(token: &str) -> String {
     sha256(token)
 }
 
-/// Tracks if a token is valid or if a user has been logged out
+/// Tracks if a token is still valid
+///
+/// There are two ways a token can expire prematurely:
+/// 1) User logs out and token is removed from the bucket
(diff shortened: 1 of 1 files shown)


src/service/token_bucket.rs
```diff
@@ -13,7 +13,12 @@ fn token_hash(token: &str) -> String {
...
 }
 
-/// Tracks if a token is valid or if a user has been logged out
+/// Tracks if a token is still valid
+///
+/// There are two ways a token can expire prematurely:
+/// 1) User logs out and token is removed from the bucket
+/// 2) Token expiry time is reduce (server side), e.g. when an token has been renewed and the old
+/// token should expiry sooner.
 pub struct TokenBucket {
```
andreievg, line 22: I just realised, this would have to be wrapped in Arc<Mutex<>> right ?

andreievg (approved):

src/service/token_bucket.rs
clemens-msupply: Yes doing this right now using a RwLock...

Issue #230, opened 2021-07-23
Schema Documentation with SchemaSpy


andreievg:
As per https://github.com/openmsupply/application-manager-server/issues/467, https://openmsupply.github.io/application-manager-server

* We need a place for current full schema (the migration patches provided by sqlx, are meant for migration imo), similar to: https://github.com/openmsupply/application-manager-server/tree/master/database/buildSchema
* Each column and tables should have a description (pg comment) in the schema itself (btw, can use markdown)
* Every time schema changes we would need to manually run creation of documentation, and upload to gh-pages branch (in schema folder)

Question, does our action re-create the branch (i only see one commit in it), want to make sure that schema docs persist.

wlthomson:
> Question, does our action re-create the branch (i only see one commit in it), want to make sure that schema docs persist.

After recent experience with `CNAME` records, I don't recommend outting anything directly in the `gh-pages` branch. Haven't looked into it, but (if possible) I'd suggest using another github action to generate the docs after the zola action has run. If you need to do it manually, another option could be to include the manually generated docs as static assets (this use-case is more complex, but this is how I prevented the `CNAME` record getting wiped on every build).

andreievg:
As part of this issue, add comments to all db schema field and tables

PR #536, opened 2021-11-03
534 add unit table

### LINKED ISSUE #534: Add unit table
* Add unit table to migrations schema
* Repository and diesel schema
* Central translations and test

andreievg:
Closes #534 

Apart from issue specific additions also added extra info to sync error (to actually see record being deserialised)


src/database/schema/unit.rs
```diff
@@ -0,0 +1,10 @@
+use super::diesel_schema::unit;
+
+#[derive(Clone, Insertable, Queryable, Debug, PartialEq, Eq, AsChangeset)]
+#[table_name = "unit"]
+pub struct UnitRow {
+    pub id: String,
+    pub name: String,
+    pub description: Option<String>,
+    pub index: i32
```
clemens-msupply, line 9: Maybe add doc comments here?

What is stored in the unit, is it something like package with 100 tablets?
What is index?

src/util/sync/translation/mod.rs
```diff
@@ -35,6 +37,7 @@ use thiserror::Error;
 pub struct SyncTranslationError {
     pub table_name: &'static str,
     pub source: serde_json::Error,
+    pub record: String,
```
clemens-msupply, line 40: Cool think this is useful!

clemens-msupply (approved):

src/database/schema/unit.rs
andreievg: Good point, i'll add it in: https://github.com/openmsupply/remote-server/issues/230

This is what is in mSupply datafile i have
Screen Shot 2021-11-05 at 1 41 35 PM

PR #552, opened 2021-11-04
550 add more sorts field to invoices

### LINKED ISSUE #550: Add otherPartyName, comment and invoiceNumberSort to invoices query

andreievg:
Closes #550 

Added macro in tests. But had issues with tests for Option (comment) and for date time (since default serialization of NaiveDateTime is different to the one in graphql type (didn't think sort test are high enough priority to pursue further)

### DIFF
--- a/src/database/repository/diesel/invoice_query.rs
+++ b/src/database/repository/diesel/invoice_query.rs
@@ -149,6 +149,27 @@ impl<'a> InvoiceQueryRepository<'a> {
                         query = query.order(invoice_dsl::finalised_datetime.asc());
                     }
                 }
+                InvoiceSortField::OtherPartyName => {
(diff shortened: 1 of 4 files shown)


clemens-msupply (approved):
Maybe we could transform the NaiveDateTime to an UTC time in our domain model and only use NaiveDateTime to store the time value in the database. (Not sure if really needed)

Not a change request: I think in this case you could avoid a macros by passing in a lambda into sort_test? e.g. (untested)

``` rust
type FieldExtractor<F: Ord> = fn(invoice: InvoiceRow) -> F;

fn sort_test(..., extractor: FieldExtractor) {...}

// and then use it like:
sort_test(..., |invoice| invoice.field,...);

```

PR #553, opened 2021-11-04
506 gql user account endpoints

clemens-msupply:
Add gql endpoints:
- registerUser
- authToken (rename to login?)
- refreshToken
- logout

I only manually tested the full flow in the playground so far...

@andreievg you where right asking why token methods where in the UserAccountService. I now moved them into a TokenService and think this is cleaner and more flexible now.

I extract the Bearer token and refresh cookie from the http request and then make the them accessible (if present) in the gql context. I had a look how to do this in a middleware but actix-web seems to make that overly complicated so I didn't follow this approach for now.

Only the logout method validates the bearer token so far. I think the errors in the logout methods are a bit to detailed right now and should be merged into a single generic "Authentication/Permission" error...

### DIFF
--- a/src/server/service/graphql/mod.rs
+++ b/src/server/service/graphql/mod.rs
@@ -1,9 +1,12 @@
 pub mod schema;
 
+use actix_web::cookie::Cookie;
+use actix_web::HttpRequest;
 use actix_web::{guard::fn_guard, web::Data, HttpResponse, Result};
 use async_graphql::http::{playground_source, GraphQLPlaygroundConfig};
 use async_graphql::{Context, EmptySubscription, SchemaBuilder};
 use async_graphql_actix_web::{Request, Response};
+use reqwest::header::COOKIE;
 
 use self::schema::{Mutations, Queries, Schema};
 use crate::server::data::auth::AuthData;
@@ -26,7 +29,7 @@ impl<'a> ContextExt for Context<'a> {
     }
 
     fn get_auth_data(&self) -> &AuthData {
-        self.data_unchecked()
+        self.data_unchecked::<Data<AuthData>>()
     }
 }
 
@@ -39,11 +42,13 @@ pub fn build_schema() -> Builder {
 pub fn config(
     repository_registry: Data<RepositoryRegistry>,
     loader_registry: Data<LoaderRegistry>,
+    auth_data: Data<AuthData>,
 ) -> impl FnOnce(&mut actix_web::web::ServiceConfig) {
     |cfg| {
         let schema = build_schema()
             .data(repository_registry)
             .data(loader_registry)
+            .data(auth_data)
             .finish();
         cfg.service(
             actix_web::web::scope("/graphql")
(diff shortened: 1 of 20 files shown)


src/server/service/graphql/schema/queries/logout.rs
```diff
@@ -0,0 +1,143 @@
...
+                return LogoutResponse::Error(ErrorWrapper {
+                    error: LogoutErrorInterface::InternalError(InternalError(
+                        "Lock error".to_string(),
+                    )),
+                });
+            }
+        },
+    };
+    // invalid the refresh token cookie
+    set_refresh_token_cookie(ctx, "logged out", 0, auth_data.debug_no_ssl);
```
andreievg, line 140: should this be at the top ?

src/server/service/graphql/schema/queries/auth_token.rs
```diff
@@ -0,0 +1,157 @@
...
+        Err(err) => {
+            return AuthTokenResponse::Error(ErrorWrapper {
+                error: match err {
+                    JWTIssuingError::CanNotCreateToken(_) => {
+                        AuthTokenErrorInterface::CanNotCreateToken(CanNotCreateToken)
+                    }
+                    JWTIssuingError::ConcurrencyLockError(e) => {
+                        error!("{}", e);
+                        AuthTokenErrorInterface::InternalError(InternalError(
+                            "Lock error".to_string(),
```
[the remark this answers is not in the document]
andreievg: ohh, i've just read rust docks for read and write, it only errors when RwLock is poisoned (other write holders panicked), so error sounds correct

src/service/token.rs
```diff
@@ -0,0 +1,336 @@
...
+            valid_for,
+            refresh_token_valid_for,
+        )
+        .map_err(|err| JWTIssuingError::CanNotCreateToken(err))?;
+
+        // add tokens to bucket
+        let mut token_bucket = self
+            .token_bucket
+            .write()
+            .map_err(|e| JWTIssuingError::ConcurrencyLockError(anyhow!("jwt_token: {}", e)))?;
```
andreievg, line 118: I think anyhow! is used when you would like to return error explicitly, format! is sufficient here. (i.e. when method returns anyhow::Result<T>)

src/service/token.rs
```diff
@@ -0,0 +1,336 @@
...
+        Ok(decoded.claims)
+    }
+
+    /// Log a user out of all sessions
+    pub fn logout(&mut self, user_id: &str) -> Result<(), JWTLogoutError> {
+        let mut token_bucket = self
+            .token_bucket
+            .write()
+            .map_err(|e| JWTLogoutError::ConcurrencyLockError(anyhow!("logout: {}", e)))?;
+        token_bucket.clear(user_id);
```
andreievg, line 216: So logging out means i log out of all of my devices ?

andreievg (approved):
Look pretty good, sorry quite a few comments and suggestions though. Can discuss in our meetings.

Most of the comments are stylistic. I am still fan of SLA, i think it's super useful in big projects (even if it means logic is spread out)

I think tokenBucket in context should actually be inside another struct that has things like expiry settings, secret etc..

src/server/service/graphql/mod.rs
```diff
@@ -57,12 +68,50 @@ pub fn config(
...
+    auth_token: Option<String>,
+    refresh_token: Option<String>,
+}
+
+fn auth_data_from_request(http_req: &HttpRequest) -> RequestUserData {
+    let headers = http_req.headers();
+    // retrieve auth token
+    let auth_token = headers.get("Authorization").and_then(|header_value| {
+        header_value.to_str().ok().and_then(|header| {
+            if header.starts_with("Bearer ") {
```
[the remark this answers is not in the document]
clemens-msupply: only had a brief look and it looks like replace doesn't return where the replace happens, e.g. if it is in the middle of the string or if multiple matches have been replaced. So keep this versions as it is more correct.

clemens-msupply:
I addressed the changes as discussed in the meeting

PR #556, opened 2021-11-05
555 ingore case in alphabetic sort

### LINKED ISSUE #555: Ignore case in alphabetic sort
For sortable text fields, make sure case is ignored in alphabetic sort

andreievg:
Closes #555 

I tried a few different ways, first just added nocase collation to actual field definitions, but can't do 'like' on non deterministic collection fields. Decided to just add dsl extensions, this if we have performance issues can revert to nocase collation if field definitions and make an extension for 'like' (or for contains).

Sqlite and postgres sorts nulls differently by default (noticed in comment sort) so added no nulls first/last. Would have liked to test in docs but it seems to only really work correctly for libs, will update once we re-arrange to docs.


src/database/diesel_extensions.rs
```diff
@@ -0,0 +1,128 @@
+use diesel::Expression;
+
+diesel_postfix_operator!(AscNoCase, " COLLATE NOCASE ASC NULLS FIRST", ());
+diesel_postfix_operator!(DescNoCase, " COLLATE NOCASE DESC NULLS LAST", ());
+diesel_postfix_operator!(AscNullsFirst, " ASC NULLS FIRST", ());
+diesel_postfix_operator!(DescNullsLast, " DESC NULLS LAST", ());
+
+// Expression extensions for order by
+pub trait OrderByExtensions: Sized {
```
clemens-msupply, line 9: OrderByExtensions -> OrderByExtension ?

migrations/postgres/00000098_no_case_collation/up.sql
```diff
@@ -0,0 +1 @@
+CREATE COLLATION NOCASE (provider = icu, locale = pg_catalog."default", deterministic = false);
```
clemens-msupply, line 1: I guess using ICU turns something like Ä into ä if server language would be German? sqlite doesn't do that does it?

clemens-msupply (approved):
Looks good

PR #559, opened 2021-11-07
Use an url to specify the sync server

clemens-msupply:
Add option to use https to connect to the central server for sync .

Not sure about local and production config. Is that ok as I did it?


andreievg (approved):
I was thinking it would just be https flag in settings, but i like this better

PR #562, opened 2021-11-09
558 workspaces

### LINKED ISSUE #558: Use Rust workspaces
# Motivation:
- useful for plugin development, e.g. shared C types, plugin sdk, plugin examples
- might improve build time

# To get plugins going:
- setup workspaces
- move all current code into a remote_server crate

# Next steps:
- crate for shared domain types (e.g. shared between repo and service layer)
- crate for repos
- crate for services
- crate for gql

clemens-msupply:
The workspace has the following crates:
- domain
- repository
- service
- graphql
- server

I tried to minimize the changes unrelated to the workspace migration and there are a couple of things that could be refactored in the future, (e.g. another utils crate?) However, some changes where necessary to keep the server working:

- Update readme, e.g. say that you should be in the server directory for the migration commands and to run the server
- update graphql_export so that it can run from the project root and from the graphql crate (it tries to auto detect from where it is running and where to generate the output, thought this could avoids some confusion...)
- Removed RepositoryRegistry and only store the StorageConnectionManager in the app data (this simplified the workspace transition and RepositoryRegistry was only holding StorageConnectionManager anyway...)
- Inline a handful of `from`s (3 or 4 if I remember correctly) They caused errors like below and it was easier to inline them than to workaround the tuple problem:
```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
  --> server/src/database/repository/./diesel/item_query.rs:25:1
   |
25 | impl From<ItemAndMasterList> for Item {
   | ^^^^^-----------------------^^^^^----
   | |    |                           |
   | |    |                           `domain::item::Item` is not defined in the current crate
   | |    this is not defined in the current crate because tuples are always foreign
   | impl doesn't use only types from inside the current crate
   |
   = note: define and implement a trait or new type instead
```

EDIT: I also removed the "mock" feature, should I put it back in? Was late last night and I spontaneously decided we are not using it and it can be removed. 

Closes #558


repository/src/repository/diesel/item_query.rs
```diff
@@ -85,7 +71,18 @@ impl<'a> ItemQueryRepository<'a> {
             .limit(pagination.limit as i64)
             .load::<ItemAndMasterList>(&self.connection.connection)?;
 
-        Ok(result.into_iter().map(Item::from).collect())
+        Ok(result
+            .into_iter()
+            .map(
```
andreievg, line 76: I really think this should be a method.

fn toDomain( .. )

.map(toDomain)

clemens-msupply: Actually like it because the whole method is not to big and having it unlined leaves all code in one place.

Will give the to_domain method a try to see how it looks. Or would comment help? e.g.

``` rust 
       // transform DB results to domain objects
       Ok(result
            .into_iter()
            .map(
```

README.md
```diff
@@ -68,17 +67,27 @@ cargo install diesel_cli --no-default-features --features "sqlite-bundled postgr
 - Migrate database tables:
 
 ```bash
+# The following commands assume you are in the server directory
+cd server
```
andreievg, line 71: why is this needed, i ran `diesel migration run --database-url ./omsupply-database.sqlite --migration-dir ../repository/migrations/sqlite/` from root repository and it created `omsupply-database.sqlite`

graphql/graphql_schema/export.rs
```diff
@@ -0,0 +1,21 @@
+use graphql::build_schema;
+use std::fs;
+//  Entry point for export_graphql target, see cargo.toml and reademe.md test section
+fn main() {
+    // Try to detect if we are in the graphql crate or in the root crate and adjust the output
+    // path accordingly.
+    let has_schema_dir = fs::read_dir(".").unwrap().find_map(|entry| {
```
andreievg, line 7: hmm, can we assume that we always run from root ? And that's the only way to run the project crates ?

README.md
clemens-msupply: You mean you called it from repository? yes that works as well, it just doesn't work if you are in the workspace root.

server/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "remote_server"
```
andreievg, line 2: i think package name should be the same as workspace package name ?

Also should configurations be in root, rather in this crate ?

andreievg (commented):
Thanks ! That's huge effort. 

I want to test before approving. 

Main comment for me is, i think we should alway run everything from `root`, rather then from individual crates.

README.md
andreievg: I ran it from workspace root

andreievg: oops, i didn't paste the correct command
`diesel migration run --database-url ./omsupply-database.sqlite --migration-dir ./repository/migrations/sqlite/`

repository/src/repository/diesel/item_query.rs
andreievg: > Or would comment help?  

I think if we can do it with a method name, there is no need for comment. (don't need to maintain the comment or validate it)

> Actually like it because the whole method is not to big

I am wondering, what would be considered big ? To me invoice query is big enough to warrant extraction of that method. And in that case it's better to use a consistent pattern across all domain repos

I really like the practice of extraction logic into method that explain what the logic is with the method name, and letting the dev have an option to dig deeper into the logic if they need to, or just see the overview, rather then forcing them to parse all of the code in their head to deduce the meaning.

andreievg:
> EDIT: I also removed the "mock" feature, should I put it back in? Was late last night and I spontaneously decided we are not using it and it can be removed.

I think mock features is not needed

andreievg:
> update graphql_export so that it can run from the project root and from the graphql crate (it tries to auto detect from where it is running and where to generate the output, thought this could avoids some confusion...)

As per earlier comment, why would we want to run things from individual crates ?

andreievg:
> Removed RepositoryRegistry and only store the StorageConnectionManager in the app data (this simplified the workspace transition and RepositoryRegistry was only holding StorageConnectionManager anyway...)

great

andreievg:
> Inline a handful of froms (3 or 4 if I remember correctly) They caused errors like below and it was easier to inline them than to workaround the tuple problem:

The more i see inline code, the less i question the notion that it's more readable. As per earlier comments if the method name describes correctly what the logic does, there should be no need to look into the actual underlying code. I think in general it's easier to digest code in 'layers', rather than large chunks with all of the layers.

graphql/src/lib.rs
```diff
@@ -1,3 +1,12 @@
+// where_clauses_object_safety is a future compatibility lint over a soundness issue which is
```
andreievg, line 1: Was this comment added in this PR ? Wondering where does the annoying notification appear ? I tried removing `#![allow(where_clauses_object_safety)]` globally, and had no warnings in compilation

repository/src/repository/diesel/item_query.rs
clemens-msupply: done in 9910caa

Just bike-shedding: big depends a bit on how much logic is in the method. In this case it just a simple DB read for which reason I would say its small (just verbose) and have left it inline. If to_domain has some more transform logic in it and is used multiple times I would put it in a method.

graphql/src/lib.rs
clemens-msupply: This was already in the old code. Maybe its not valid anymore, removed in daa9382

server/Cargo.toml
clemens-msupply: done in b477daf

graphql/graphql_schema/export.rs
clemens-msupply: Agree only problem I see is. How do we distribute the final binary and how do we expect the dir structure to be. As you said below, we can move configurations dir into the root and everything should work fine :) Will try it this afternoon.

andreievg (approved):
Thanks for changes looks good!

andreievg:
Oops i didn't do a full check after approving. Pending `everything should be run from root` comment. 

If you run with `-p` it would look inside configuration folder in root of the project, which i think makes it work correctly in producation

clemens-msupply:
All commands assume the root directory now: ef9727c

andreievg:
> All commands assume the root directory now:

Thanks, just one more thing to pass tests, 

https://github.com/openmsupply/remote-server/compare/558-workspaces...558-tests

PR #568, opened 2021-11-12
Simplify server directory structure

clemens-msupply:
Some initial clean up (sync is still in server)

- Remove REST
- Merge content from utils into the server root dir
- Merge content from server into the server root dir


andreievg (approved):
Thanks, looks good

PR #572, opened 2021-11-15
563 repository dir

### LINKED ISSUE #563: Improve directory structure (follow up from workspace refactoring)
graphql:
- Remove schema directory and merge content into parent directory
server:
- remove utils directory and server directory and merge both 
- put translation it into a TranslationService?
- remove service directory
- remove rest or just move it into parent directory

clemens-msupply:
1) repository::repository::* -> repository::*
2) Introduce a util (or better utils?) crate. Currently only contains hash and uuid
3) Make sqlite the default feature and update readme

Closes #563


andreievg (approved):
Awesome, thanks!

clemens-msupply (commented):
@andreievg I just had another look and I actually don't understand why this works:
`#[cfg(all(feature = "postgres", not(feature = "sqlite")))]`
It compiles for postgres but sqlite is still set on default, isn't it? I tried to remove the `not` part and the compilation fails for postgres. Do you have any idea?

I will merge now anyway to keep things going.

PR #573, opened 2021-11-15
Fix tests when run from VSCode

clemens-msupply:
Two problems:
- sqlite feature was not set for service tests
- In the IDE tests a run from the project root dir -> fix search for
migration files

### DIFF
--- a/repository/src/test_db.rs
+++ b/repository/src/test_db.rs
@@ -7,7 +7,17 @@ use crate::{
 use diesel::r2d2::{ConnectionManager, Pool};
 use diesel_migrations::{mark_migrations_in_directory, search_for_migrations_directory};
 
-use std::path::Path;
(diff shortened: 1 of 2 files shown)


andreievg:
> In the IDE tests a run from the project root dir -> fix search for
migration files

Which of the test need this ? I've been running test via vs code rust analyzer play buttons (with the features changes in toml too), but never had issues with any of the migrations not being found

clemens-msupply:
The service tests are not working, but I just checked they are working when pressing run but debug doesn't work...

andreievg:
> The service tests are not working, but I just checked they are working when pressing run but debug doesn't work...

Oki, we could also have fixed it by putting migrations in root (as suggested earlier), as `search_for_migrations_directory` already does something similar to `find_test_migration_directory` (except it expects migrations dir to be in root)

PR #574, opened 2021-11-15
Versioned document

clemens-msupply:
Includes:
- Document repository
- CanonicalJsonValue to serialize a json object in a deterministic way, i.e. to get the same object hash even if input object keys are sorted differently => `hash(obj) == hash(obj')` if `obj` and `obj'` are the same but have differently sorted keys.
- Common ancestor finder: finds the latest common ancestor of two document versions (needed for a three way merge)
- Two and three way merges
- Document Service to add new document versions and proposes a merged document version if there was a concurrent edit

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1322,6 +1322,9 @@ name = "domain"
 version = "0.1.0"
 dependencies = [
  "chrono",
+ "serde 1.0.130",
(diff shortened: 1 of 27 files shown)


clemens-msupply:
New branch: document-versions-2

PR #575, opened 2021-11-16
506 auth/permission validation helper and me endpoint

clemens-msupply:
- Add helper method to validate that a user is authenticated and has the correct permissions.
- Add `me` endpoint that uses the helper method
- Refactor the `logout` endpoint to use the helper
- Add debug flag to disable the authentication check and make the user an Admin user

The permission system is work in progress, happy to discuss. For the validation I use some higher order functions to have something like a permission DSL. For example, a complex permission check could look like:
``` rust
validate(..., vec![
   // ANY required roles to access this endpoint
   any(vec![ 
        has_api_role(ApiRole::Support),
        has_api_role(ApiRole::Plugin)
   ]),
   // AND:
   has_store_access((store_id, StoreRole::StoreWorker)),
   // AND all of the following:
   all(vec![
       some(),
       other(),
       role(),
       checker()
   ])
])
```

### DIFF
--- a/graphql/src/schema/queries/me.rs
+++ b/graphql/src/schema/queries/me.rs
@@ -1,10 +1,9 @@
 use async_graphql::*;
-use service::permission_validation::{
-    has_api_role, validate, validation_denied_kind_to_string, ValidationError,
-};
+use service::permission_validation::{has_api_role, validate, ValidationError};
 use service::user_account::{UserAccount, UserAccountService};
 
 use crate::schema::types::{AccessDenied, DatabaseError, InternalError};
+use crate::schema::validation_denied_kind_to_string;
 use crate::ContextExt;
 
 use super::ErrorWrapper;
--- a/graphql/src/schema/queries/me.rs
+++ b/graphql/src/schema/queries/me.rs
@@ -15,11 +15,12 @@ pub struct Me {
 
 #[Object]
 impl Me {
-    /// User id of the logged out user
+    /// Internal user id
     pub async fn user_id(&self) -> &str {
         &self.user.id
     }
 
+    /// The user's email address
     pub async fn email(&self) -> &Option<String> {
         &self.user.email
     }
--- a/graphql/src/schema/queries/me.rs
+++ b/graphql/src/schema/queries/me.rs
@@ -1,0 +1,83 @@
+use async_graphql::*;
+use service::permission_validation::{
+    has_api_role, validate, validation_denied_kind_to_string, ValidationError,
+};
+use service::user_account::{UserAccount, UserAccountService};
+
+use crate::schema::types::{AccessDenied, DatabaseError, InternalError};
+use crate::ContextExt;
+
+use super::ErrorWrapper;
+
+pub struct Me {
+    pub user: UserAccount,
+}
+
+#[Object]
+impl Me {
+    /// User id of the logged out user
+    pub async fn user_id(&self) -> &str {
+        &self.user.id
+    }
+
+    pub async fn email(&self) -> &Option<String> {
+        &self.user.email
+    }
+}
+
+#[derive(Interface)]
+#[graphql(field(name = "description", type = "&str"))]
+pub enum MeErrorInterface {
+    AccessDenied(AccessDenied),
+    DatabaseError(DatabaseError),
+    InternalError(InternalError),
+}
+
+pub type MeError = ErrorWrapper<MeErrorInterface>;
+
+#[derive(Union)]
+pub enum MeResponse {
+    Error(MeError),
+    Response(Me),
+}
+
+pub fn me(ctx: &Context<'_>) -> MeResponse {
+    let user = match validate(
+        ctx.get_connection_manager(),
+        ctx.get_auth_data(),
+        &ctx.get_auth_token(),
+        vec![has_api_role(service::permissions::ApiRole::User)],
+    ) {
+        Ok(value) => value,
+        Err(err) => {
+            let error = match err {
+                ValidationError::Denied(denied) => MeErrorInterface::AccessDenied(AccessDenied(
+                    validation_denied_kind_to_string(denied),
+                )),
+                ValidationError::InternalError(err) => {
+                    MeErrorInterface::InternalError(InternalError(err))
+                }
+            };
+            return MeResponse::Error(ErrorWrapper { error });
+        }
+    };
+
+    let user_service = UserAccountService::new(&user.connection);
+    let user = match user_service.find_user(&user.user_id) {
+        Ok(Some(user)) => user,
+        Ok(None) => {
+            return MeResponse::Error(ErrorWrapper {
+                error: MeErrorInterface::InternalError(InternalError(
+                    "Can't find user account data".to_string(),
+                )),
+            })
+        }
+        Err(err) => {
+            return MeResponse::Error(ErrorWrapper {
+                error: MeErrorInterface::DatabaseError(DatabaseError(err)),
+            })
+        }
+    };
+
+    MeResponse::Response(Me { user })
+}
--- a/graphql/src/schema/queries/me.rs
+++ b/graphql/src/schema/queries/me.rs
@@ -9,12 +9,12 @@ use crate::ContextExt;
 
 use super::ErrorWrapper;
 
-pub struct Me {
+pub struct User {
     pub user: UserAccount,
 }
 
 #[Object]
-impl Me {
+impl User {
     /// Internal user id
     pub async fn user_id(&self) -> &str {
         &self.user.id
@@ -28,21 +28,21 @@ impl Me {
 
 #[derive(Interface)]
 #[graphql(field(name = "description", type = "&str"))]
-pub enum MeErrorInterface {
+pub enum UserErrorInterface {
     AccessDenied(AccessDenied),
     DatabaseError(DatabaseError),
     InternalError(InternalError),
 }
 
-pub type MeError = ErrorWrapper<MeErrorInterface>;
+pub type UserError = ErrorWrapper<UserErrorInterface>;
 
 #[derive(Union)]
-pub enum MeResponse {
-    Error(MeError),
-    Response(Me),
+pub enum UserResponse {
+    Error(UserError),
+    Response(User),
 }
 
-pub fn me(ctx: &Context<'_>) -> MeResponse {
+pub fn me(ctx: &Context<'_>) -> UserResponse {
     let user = match validate(
         ctx.get_connection_manager(),
         ctx.get_auth_data(),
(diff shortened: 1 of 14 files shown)


graphql/src/schema/types/mod.rs
```diff
@@ -233,6 +236,20 @@ impl From<RepositoryError> for NodeError {
...
+
+pub struct AccessDenied(pub String);
+
+#[Object]
+impl AccessDenied {
+    pub async fn description(&self) -> &'static str {
+        "Access Denied"
+    }
+
+    pub async fn full_error(&self) -> String {
```
andreievg, line 248: maybe -> &str
&self.0

graphql/src/schema/queries/me.rs
```diff
@@ -0,0 +1,84 @@
...
+pub enum MeErrorInterface {
+    AccessDenied(AccessDenied),
+    DatabaseError(DatabaseError),
+    InternalError(InternalError),
+}
+
+pub type MeError = ErrorWrapper<MeErrorInterface>;
+
+#[derive(Union)]
+pub enum MeResponse {
```
andreievg, line 40: maybe UserResponse ? (route can still stay as me, but it seems clearer to api consumer)

graphql/src/schema/queries/me.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    };
+
+    let user_service = UserAccountService::new(&user.connection);
+    let user = match user_service.find_user(&user.user_id) {
+        Ok(Some(user)) => user,
+        Ok(None) => {
+            return MeResponse::Error(ErrorWrapper {
+                error: MeErrorInterface::InternalError(InternalError(
+                    "Can't find user account data".to_string(),
```
andreievg, line 72: H, i wonder if we can use existing single record response patter (or extend it generically to include validation errors), generic single record response pattern returns node error (which in turn returns NotFound, which seems the same as "Can't find user account data")

graphql/src/schema/queries/me.rs
```diff
@@ -0,0 +1,84 @@
...
+        ctx.get_connection_manager(),
+        ctx.get_auth_data(),
+        &ctx.get_auth_token(),
+        vec![has_api_role(service::permissions::ApiRole::User)],
+    ) {
+        Ok(value) => value,
+        Err(err) => {
+            let error = match err {
+                ValidationError::Denied(denied) => MeErrorInterface::AccessDenied(AccessDenied(
+                    validation_denied_kind_to_string(denied),
```
andreievg, line 56: Should this mapping be defined in API rather then service ?

service/src/permission_validation.rs
```diff
@@ -0,0 +1,233 @@
...
+            checked_msg.push_str(&name);
+        }
+        (checked_msg.to_owned(), Some(format!("Not {}", checked_msg)))
+    });
+    checker
+}
+
+pub fn has_api_role(role: ApiRole) -> PermissionChecker {
+    Box::new(move |data: &ValidationUserData| {
+        let name = "api role".to_string();
```
andreievg, line 35: Really not a fan of hard coded strings, can this be an enum ?

service/src/permission_validation.rs
```diff
@@ -0,0 +1,233 @@
...
+    checkers: Vec<PermissionChecker>,
+) -> Result<ValidatedUser, ValidationError> {
+    let validated_auth = validate_auth(auth_data, auth_token)?;
+
+    // TODO not used yet here but will be needed later for the PermissionService
+    let connection = match connection_manager.connection() {
+        Ok(con) => con,
+        Err(err) => {
+            error!("validate: Failed to get connection: {}", err);
+            return Err(ValidationError::InternalError(
```
andreievg, line 159: Why not use existing pattern with database error ?

service/src/permission_validation.rs
```diff
@@ -0,0 +1,233 @@
...
+    return Ok(ValidatedUserAuth { user_id, claims });
+}
+
+pub struct ValidatedUser {
+    pub user_id: String,
+    pub claims: OmSupplyClaim,
+    pub permissions: UserPermissions,
+    /// For convenience the connection that is used to validate the user is passed on for further
+    /// use.
+    pub connection: StorageConnection,
```
andreievg, line 142: Can i ask what the convenience is ?

graphql/src/schema/queries/me.rs
```diff
@@ -0,0 +1,84 @@
...
+    Error(MeError),
+    Response(Me),
+}
+
+pub fn me(ctx: &Context<'_>) -> MeResponse {
+    let user = match validate(
+        ctx.get_connection_manager(),
+        ctx.get_auth_data(),
+        &ctx.get_auth_token(),
+        vec![has_api_role(service::permissions::ApiRole::User)],
```
andreievg, line 50: I really like the idea of centrally defining and listing permissions, and I think it's service responsibility ? I am still digesting this, so can't quite figure out the benefit of closures here.  As far as i can see there are two subsets of permissions so far, api and store.

I can't see quite yet what a global static permission list can't do that closure DSL can.

i.e.
```rust
enum Route {
   Me,
   Invoice,
   Invoices,
   InsertOutboudShipmnet,
   UpdateInboundShipment,
   ...
}

struct RoutePermission {
   api: Option<Vec<ApiRole>>,
   store_restrictions: Option<Vec<StoreRole>>
}

// Global permission definition

let permissions: HashMap<Route, RoutePermission> = HashMap::new();

permissions.insert(Route::Me, RoutePermissions {
     api: Some(vec![ApiRole::Admin, ApiRole::User]),
     store_restrictions: None,
 });

permissions.insert(Route::InsertOutboundShipment, RoutePermissions {
     api: Some(vec![ApiRole::Admin, ApiRole::User]),
     store_restrictions: Some(vec![StoreRole::EditOutboundShipments]),
 });

```

andreievg (commented):
Still getting my head around it.

Main point, i am trying to understand the need for DSL and closure use, see comments below

graphql/src/schema/types/mod.rs
clemens-msupply: It says:

error[E0515]: cannot return value referencing temporary value
   --> graphql/src/schema/types/mod.rs:242:1
    |
242 | #[Object]
    | ^^^^^^^^^ returns a value referencing data owned by the current function
...
249 |         &format!("{:#}", self.0)
    |          ----------------------- temporary value created here
    |
    = note: this error originates in the attribute macro `Object` (in Nightly builds, run with -Z macro-backtrace for more info)

graphql/src/schema/queries/me.rs
clemens-msupply: done in 98afae0

graphql/src/schema/queries/me.rs
clemens-msupply: Think its nice to have a central place to have an overview on all permissions. On the other hand its a bit of a pain to develop because you have another file to touch when adding a new endpoint. Are there any other advantages to have it centrally? seems to be simpler to me to just have it inlined.

This was the auth lib I ran into when looking up ffi for the plugins:
https://docs.osohq.com/rust/getting-started/quickstart.html
Not sure if its a good idea seem to be pretty complex...

service/src/permission_validation.rs
clemens-msupply: The `validate` method needs to read the permissions from the DB. So either you open a connection before doing the auth check and pass it to `validate` or you safe some boilerplate and do it in `validate` and return the connection for further use.

service/src/permission_validation.rs
clemens-msupply: Yes right, I tried to minimize the errors that needs to be handled though, i.e. when using the validation method only two errors need to be handled, AccessDenied and InternalError.

DB connection error is an internal errors really, it shouldn't really happen in normal operation.

graphql/src/schema/queries/me.rs
clemens-msupply: Is "user account not found" a really error though? For example, if you delete a user you should have invalidated the existing tokens and this code shouldn't be reached. Its nothing the frontend should care about.

graphql/src/schema/queries/me.rs
clemens-msupply: Thanks done in 970b34c

service/src/permission_validation.rs
clemens-msupply: I think I have to have a good look at the PermissionChecker return value anyway. The returned name was just to generate more meaningful error messages in the `any` checker. This can probably be done in a nicer way.

graphql/src/schema/types/mod.rs
andreievg: I was meaning
```rust
 pub async fn full_error(&self) -> &str {
        &self.0
    }
```

graphql/src/schema/queries/me.rs
andreievg: There is a bunch of places (for single record node responses) where the code is technically unreachable (database error like this) but we still have to handle a case and return not found error. I guess it's fine to return InternalError, but was hoping we could keep consistency.  Anyways this can be done further up the track when we do authorisation errors on existing queries/mutations and see maybe we can see if we can consolidate.

service/src/permission_validation.rs
andreievg: oki

service/src/permission_validation.rs
andreievg: Sorry, I still don't understand, what is the boilerplate you are saving ? You are still getting connection inside validate so what is saved by extracting it further up ? Also this is not being done inside transaction so why wouldn't both validate and other methods that use it further get another connection ?

graphql/src/schema/queries/me.rs
andreievg: > On the other hand its a bit of a pain to develop because you have another file to touch when adding a new endpoint. Are there any other advantages to have it centrally? seems to be simpler to me to just have it inlined.

To me, having a central definitions of permissions is a better structure (it's easy to see full overview of permissions for all queries, and separates functional area by context). 

Also it's much easier to do the testing at service end rather then graphql (you wouldn't be able to split the mapping and service test by having this inline in query)

Lastly, personally, I really think it's trivial to edit another file (we've already split the object up in a way where you have to edit multiple files for adding new queries  -> repository data type is different from query data type, there is also the domain type, there is a service for a query). I think API should be `just` for mapping.

andreievg (approved):

andreievg:
I've been thinking about the requirement to use closures, we can group validations by condition quite easy without closures:

```rust
pub enum PermissionDSL {
   HasApiRole(ApiRole),
   HasStoreAccess(StoreRole),
   And(Vec<PermissionDLS>),
   Or(Vec<PermissionDSL>),
   Any(Vec<PermissionDSL>)
}

// ... Central definition

use PermissionDSL::*;
// route Me
permissions.insert(Route::Me, HasApiRole(ApiRole::User));

// route Complex DSL

permissions.insert(
    Route::Complex,
    And(vec![
        Any(vec![
            HasApiRole(ApiRole::Support),
            HasApiRole(ApiRole::Plugin),
        ]),
        HasStoreAccess(StoreRole::StoreWorker),
        All(vec![Some(), Other(), Role(), Checker()]),
    ]),
);
```

This leaves one more scenario where closures inline in query can be beneficial, and that is if they have dynamic parameters that need to be evaluated by permission validation that come from say input. I really can't think of a use case where these dynamic parameters need to exist, and if they do need to be present we can do an another enum varian of in data for validation with extra parameters being passed through

service/src/permission_validation.rs
clemens-msupply: Its just to avoid doing the error handling for the connection error twice, i.e. the possible db connection error when fetching user permissions and the connection error when getting the connection for the actual request... Not strong about it.

clemens-msupply:
Cool lets do the PermissionDSL! probably easier if we don't need anything more complex anyway

graphql/src/schema/types/mod.rs
clemens-msupply: thanks dee29f7

Issue #543, opened 2021-11-04
Add location to existing queries and mutations


andreievg:
* Stock line node and invoice line node to have location
* Able to specify location in inbound shipment line
* Transfer location in inbound shipment line to stock line
* Populate location correctly in outbound shipment line (from stock line)

PR #576, opened 2021-11-17
543 add location query

andreievg:
Closes: #543 

Quite a big diff, but a lot of changes are from tests.

I really like the service as trait idea, and testing just the mapping in graphql tests, great idea @clemens-msupply.


repository/src/schema/location.rs
```diff
@@ -4,7 +4,8 @@ use super::diesel_schema::location;
 #[table_name = "location"]
 pub struct LocationRow {
     pub id: String,
-    pub code: String,
```
andreievg, line 7: this is really weird, maybe i missed something, but it seems fields must be in the same order as in dsl definitions ? Otherwise code and name were being mixed up in diesel results.

repository/src/db_diesel/location.rs
```diff
@@ -0,0 +1,143 @@
...
+///     if let Some(value) = equal_filter.equal_to {
+///         query = query.filterd(location_dsl::id.eq(value));
+///     }
+///
+///     if let Some(value) = equal_filter.equal_any {
+///         query = query.filter(location_dsl::id.eq_any(value));
+///     }
+/// }
+/// ```
+macro_rules! apply_equal_filter {
```
andreievg, line 30: I think this is quite ease to read and makes the create_filtered_query easy to read too, can i do this for all of the filtering/sorting ?

clemens-msupply: Can't warm up with macros but lets give it a try, this seem quite repetitive and used a lots.

You can't pass `location_dsl::id` or even `location_dsl::id.eq` as a macro parameter, can you?

server/tests/graphql/locations.rs
```diff
@@ -0,0 +1,351 @@
...
+                Ok(ListResult {
+                    rows: vec![Location {
+                        id: "test_id".to_owned(),
+                        name: "test_name".to_owned(),
+                        code: "test_code".to_owned(),
+                        on_hold: true,
+                    }],
+                    count: 1,
+                })
+            })));
```
clemens-msupply, line 95: Personally I would move the service setup before setting up the query but that might be a matter of taste.

server/tests/graphql/locations.rs
```diff
@@ -0,0 +1,351 @@
...
+
+        assert_gql_query(
+            &settings,
+            query,
+            &None,
+            &expected,
+            Some(ServicesOverride::new().location_service(service)),
+        )
+        .await;
+    }
```
clemens-msupply, line 216: I guess we don't need to test the pagination for every endpoint and testing it for one is enough?

server/tests/graphql/locations.rs
```diff
@@ -0,0 +1,351 @@
...
+        let service: Option<Box<dyn LocationServiceQuery>> =
+            Some(Box::new(TestService(|_, _, sort| {
+                assert_eq!(
+                    sort,
+                    Some(Sort {
+                        key: LocationSortField::Name,
+                        desc: None
+                    })
+                );
+                Err(ListError::LimitAboveMax(1000))
```
clemens-msupply, line 259: I got confused by the returned error, maybe say earlier that the test tests that the variables are passed correctly to the service and that the error is not important...

service/src/location/query/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+        let mut result = repository.query(
+            Pagination::one(),
+            Some(LocationFilter::new().match_id(&id)),
+            None,
+        )?;
+
+        if let Some(record) = result.pop() {
+            Ok(record)
+        } else {
+            Err(SingleRecordError::NotFound(id))
```
clemens-msupply, line 44: Did we agree to to return Option instead of NotFound? (if yes, no need to change right now)

clemens-msupply (approved):
Very nice. Tests are very easy to read now! Almost looking forward writing some tests now ;-)

service/src/service_registry.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub type ServiceMap = Map<AnyService>;
+pub type AnyService = dyn Any + Send + Sync;
+
+pub struct ServiceRegistry {
+    pub services: ServiceMap,
+}
+
+impl ServiceRegistry {
+    pub fn get<T: anymap::any::Any + Send + Sync>(&self) -> &T {
```
clemens-msupply, line 14: Regarding my comment earlier for how to pass in parameters. Maybe make the registry the "service factory" and have methods like `get_location_service(...)`?

Or maybe something like `get_abc_service_new_tx(...) -> Result<(AbcService, StorageConnection), RepositoryError>` ? to also start a new tx. This could help to keep the gql layer cleaner while also keeping the tx management out of the services and making the services more reusable in other tx..

This would also remove the possible unreachable panic when calling anymap.get(), even though this should be caught by tests. However, don't see why we need to give up type safety here if we can write it so that it is guaranteed that the service exists.

repository/src/db_diesel/location.rs
andreievg: > pass location_dsl::id 

Do you mean use
```rust
apply_equal_filter!(query, filter, location_dsl::id, id);
apply_equal_filter!(query, filter, location_dsl::name, name);
apply_equal_filter!(query, filter, location_dsl::code, code);
```
vs 
```rust
apply_equal_filter!(query, filter, location_dsl, id);
apply_equal_filter!(query, filter, location_dsl, name);
apply_equal_filter!(query, filter, location_dsl, code);
```
i thought the later is cleaner, but happy to change to the former

>  even location_dsl::id.eq

What would be the call of the macro for this one ? (`apply_equal_filter!()` has a match to do either eq or eq_any in this PR)


```

File: service/src/invoice_line/inbound_shipment_line/update/validate.rs [hunk 8]
```diff
@@ -31,13 +31,16 @@ pub fn validate(
...
     check_line_belongs_to_invoice(&line, &invoice)?;
     check_invoice_type(&invoice, InvoiceType::InboundShipment)?;
     check_invoice_finalised(&invoice)?;
 
     check_batch(&line, connection)?;
 
+    // InvoiceDoesNotBelongToCurrentStore
+    // StockLineDoesNotBelongToCurrentStore
+    // LocationDoesNotBelongToCurrentStore
+
```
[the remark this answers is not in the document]
andreievg: Yeah for permission system, added TODO:
e588b3d334d54e52970725c1a7132eae41d40bf9

service/src/invoice_line/inbound_shipment_line/insert/validate.rs
```diff
@@ -30,10 +30,13 @@ pub fn validate(
...
 
     let invoice = check_invoice_exists(&input.invoice_id, connection)?;
-    // check_store(invoice, connection)?; InvoiceDoesNotBelongToCurrentStore
     check_invoice_type(&invoice, InvoiceType::InboundShipment)?;
     check_invoice_finalised(&invoice)?;
 
+    // InvoiceDoesNotBelongToCurrentStore
+    // StockLineDoesNotBelongToCurrentStore
+    // LocationDoesNotBelongToCurrentStore
+
```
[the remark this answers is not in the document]
andreievg: e588b3d334d54e52970725c1a7132eae41d40bf9

server/tests/graphql/locations.rs
andreievg: do you mean before the `expected` ? 
Since `query` is common here and service changes

graphql/src/lib.rs
```diff
@@ -32,6 +34,10 @@ impl<'a> ContextExt for Context<'a> {
         self.data_unchecked::<Data<LoaderRegistry>>().get::<T>()
     }
 
+    fn get_service<T: anymap::any::Any + Send + Sync>(&self) -> &T {
+        self.data_unchecked::<Data<ServiceRegistry>>().get::<T>()
+    }
```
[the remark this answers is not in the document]
andreievg: Replying below

server/tests/graphql/locations.rs
andreievg: yeah, i guess just the one, i moved it to pagination test to be more specific: b91a5ad1360d245614574587dd84b0a05000f1a3

server/tests/graphql/locations.rs
andreievg: Changed to return empty list: b91a5ad1360d245614574587dd84b0a05000f1a3

service/src/location/query/mod.rs
andreievg: oops, thanks for reminder, a7ce64f73c0108cbdd45affa94e9f79fc7ac2483

service/src/service_registry.rs
andreievg: > get_abc_service_new_tx(...) -> Result<(AbcService, StorageConnection), RepositoryError>

Hmm, I kind of like the factory idea. 

For transaction though, I think we need to have a deeper think, i.e. it needs to be inside a closure, so `get_abc_service_new_tx` would be returning AbcService and StorageConnection inside closure inside it's parameters, which is fine also, but if we want to return a group of services to all be run within a transaction, we can do something like: 

```rust
match service_factory.start_tx(|service_factory_tx| {
  service_factory_tx.abc_service(abc_service_parameters).map_err(..)?

  service_factory_tx.cba_service(cba_service_parameters).map_err(..)?

}) {
...
};
```

repository/src/db_diesel/location.rs
clemens-msupply: Personally I find `apply_equal_filter!(query, filter.field, location_dsl::id)` easier to read...

I thought about something like `apply_filter!(query, filter_value, dsl::column, eq)` so that you can use the macro for all kind of filters not only for eq... I may miss something, ignore me in this case.

server/tests/graphql/locations.rs
clemens-msupply: yes before expected

service/src/service_registry.rs
clemens-msupply: nice

repository/src/db_diesel/location.rs
andreievg: > Personally I find apply_equal_filter!(query, filter.field, location_dsl::id) easier to read...

Agree

f940390b21742f34096c234f702d011bdaa5d289

> I thought about something like apply_filter!(query, filter_value, dsl::column, eq) so that you can use the macro for all kind of filters not only for eq... I may miss something, ignore me in this case.

I see, so just one macro for all use cases, hmm, I think individual macros might be better then super macro. Let's see how it looks when we do it for other filter types

server/tests/graphql/locations.rs
andreievg: 3d9d48aafe1e08bd5ebbffb8e22184f8343f324b

service/src/service_registry.rs
andreievg: I'll do this in seperate PR: https://github.com/openmsupply/remote-server/issues/602

PR #578, opened 2021-11-17
543 add location query part 2

andreievg:
Closes: #543 

* Added location to stock line and invoice line
* Added stock to location
* Added test for stock in location

Was going to add tests for location and stock/invoice line, but thought to hold of, it seems easier to test this when services as exposed as traits. (did manual graphql test for those)


clemens-msupply (approved):
Looks good to me

PR #597, opened 2021-11-22
596 postgres feature to work

andreievg:
Closes: #596 

Sqlite default features was still enabled when using `--features postgres` flag. So though to do a check on postgres flag rather than sqlite flag.


clemens-msupply (approved):
Thanks!

repository/src/database_settings.rs
```diff
@@ -29,7 +29,7 @@ impl DatabaseSettings {
     }
 }
 
-#[cfg(feature = "sqlite")]
+#[cfg(not(feature = "postgres"))]
```
mark-prins, line 32: commenting as a non-rust developer! so I don't know the implications of what I'm saying and you can happily ignore.

having this logic in so many places in the code feels brittle - and requires a lot of file changes. Is it not possible to have this check once, and set a global flag for `database_engine` or similar?

Would also help to isolate the implicit configuration here, of `not(postgres)` = `sqlite`

clemens-msupply: Yes it's brittle, and everything to save a dynamic dispatch... Not aware of an easier solution, probably you can write a macro for it ;-)

mark-prins:
tested and working on windows!

PR #599, opened 2021-11-22
update M1 doc with multi-platform quotes

### LINKED ISSUE #561: Update docs for windows users
Another item raised earlier when I had trouble with windows didn't make it into the docs, and I've now had the same problem because I forgot about the issue.

When running the diesel migration, you cannot use single quotes in the parameters, when running on windows.

It would also be good to warn users not to try and use the `DATABASE_URL` env var... as this will crash the server

mark-prins:
Fixes #561 

Scratches an itch.. I've referred to the M1 docs a few times when working on windows and it would be useful to copy and paste. Have taken the change from the main readme across.

Also detailed the parts of the command a little, as it took me a moment to realise what was going on when I first used it, thought that it would be clearer for new users if I broke the command down.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,14 @@ cargo install diesel_cli --no-default-features --features "sqlite-bundled postgr
 
 ```bash
 # postgres
-diesel migration run --database-url="postgres://postgres:password@localhost:5432/omsupply-database" --migration-dir ./repository/migrations/postgres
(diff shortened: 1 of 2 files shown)


clemens-msupply (approved):
Thanks looks good. Do we have to duplicate the readme in the docs though?

mark-prins:
@clemens-msupply - no, I don't think so. Thought that in this case it was just a temporary entry for M1, so yes, this is a fairly redundant PR! Should be linking to the readme instead of duplicating

PR #601, opened 2021-11-23
Use a PermissionDSL struct instead of some checker functions

clemens-msupply:
- Global permission setup
- Introduce a ValidationService (needed for testing to swap out the
global permissions)
- Remove StorageConnection setup since this would be replaced when using
a service factory.

Still a bit verbose, may another validation helper method once we have
the service factory (?)...

### DIFF
--- a/service/src/permission_validation.rs
+++ b/service/src/permission_validation.rs
@@ -1,46 +1,32 @@
-use log::error;
-use repository::{StorageConnection, StorageConnectionManager};
+use std::collections::HashMap;
+
+use repository::StorageConnection;
 
 use crate::{
     auth_data::AuthData,
-    permissions::{ApiRole, PermissionService, UserPermissions},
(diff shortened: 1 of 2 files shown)


service/src/permission_validation.rs
```diff
@@ -1,46 +1,32 @@
...
-        let name = "api role".to_string();
-        if let Some(_) = data.permissions.api.iter().find(|r| r == &&ApiRole::Admin) {
-            return (name, None);
-        }
-        if let Some(_) = data.permissions.api.iter().find(|r| r == &&role) {
-            return (name, None);
-        }
-        (name, Some("Missing api role".to_string()))
-    })
+/// TODO use lasy_static! ?
```
andreievg, line 25: Would assume we can add this to authentication service instantiation (either when it's instantiated and put into registry or service factory/provider), thus wouldn't need to instantiate it on every call to the service.

andreievg (approved):
Thanks!

PR #603, opened 2021-11-23
Remove #[graphql(flatten)]

clemens-msupply:
Surprising easy to remove the macro!

For the NodeError when getting the invoice/line after mutating the invoice/line. I think if we put the whole mutation into a transaction we can remove the error and merge it into the invoice/line error? i.e. if we fail to retrieve the invoice/line than it is an internal error that shouldn't really happen.

Related, do we actually need to get the invoice/line after mutating it? isn't it enough to return the mutated invoice/line.

### DIFF
--- a/graphql/src/schema/mutations/inbound_shipment/insert.rs
+++ b/graphql/src/schema/mutations/inbound_shipment/insert.rs
@@ -3,8 +3,8 @@ use async_graphql::*;
 use crate::schema::{
     mutations::{ForeignKey, ForeignKeyError, RecordAlreadyExist},
     types::{
-        get_invoice_response, DatabaseError, ErrorWrapper, InvoiceNodeStatus, InvoiceResponse,
-        NameNode,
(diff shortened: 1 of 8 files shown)


andreievg:
> Surprising easy to remove the macro!

Cool, I am glad that the structure makes some things easy

> For the NodeError when getting the invoice/line after mutating the invoice/line. I think if we put the whole mutation into a transaction we can remove the error and merge it into the invoice/line error? i.e. if we fail to retrieve the invoice/line than it is an internal error that shouldn't really happen.

That sounds's better, for services to return InvoiceLine/Invoice. But yeah there is still the potential of DB error, but that could be internal error that will never really occurs, but still needs to be handled in rust (unless we do as suggested and return the InvoiceLine in mem rather then DB) 

> Related, do we actually need to get the invoice/line after mutating it? isn't it enough to return the mutated invoice/line.

I was thinking it's more accurate to return the line based on the new state of it in DB.

andreievg (approved):

clemens-msupply:
Thanks, I created a ticket for the error issue #605

PR #607, opened 2021-11-24
Add an item type

clemens-msupply:
Did I get the translation right?

```
        LegacyItemType::non_stock => ItemType::NonStock,
        LegacyItemType::service => ItemType::Service,
        LegacyItemType::general => ItemType::Stock,
```


andreievg:
> Did I get the translation right?

I think so. Although I am not sure what non stock would be in omSupplyl

andreievg (approved):
Thanks!

PR #608, opened 2021-11-24
Fix invoice filter like query

clemens-msupply:


mark-prins (approved):
👍  perfect! thanks - updated demo server and tested against that, works nicely

PR #609, opened 2021-11-25
514 numbering table

clemens-msupply:
Implements just a basic key value store (counter id -> value) and a NumberService that allows to specify store id an table.

What is the reason you wanted to store store id and table in the the number row? do we need to? I left it out for now because I though it can all be handled in the service (just my assumption, will add it if it's important).

I use the primary table key to store the counter name so that the db can use the index to look up the table by name (without creating a new index)...

For translation: how does sync works? how to make sure that local counter is not modified while new data is pulled from the central server?

EDIT: working on the transaction issue in a separate PR (the problem that the increment needs be done in an tx)


repository/src/db_diesel/number_row.rs
```diff
@@ -0,0 +1,52 @@
...
+                diesel::insert_into(number_dsl::number)
+                    .values(&row)
+                    .execute(&self.connection.connection)?;
+                Ok(row)
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    pub fn find_one_by_id(&self, counter_id: &str) -> Result<NumberRow, RepositoryError> {
```
andreievg, line 46: Should this be Result<Option, ?

andreievg (approved):
> What is the reason you wanted to store store id and table in the the number row? do we need to? I left it out for now because I though it can all be handled in the service (just my assumption, will add it if it's important).

Just though it's rude to have those identifiers as a concatenated strings, this caused some complicated code in mSupply when it ties to figure out which records need to sync to the site. And since we are going to re-write central some time, it would be good if both remote and central data structure matches (would change to store_id and table, etc.. as columns for to simplify sync in central).

> I use the primary table key to store the counter name so that the db can use the index to look up the table by name (without creating a new index)...

I have a feeling this is the reason it was done in mSupply. I think we don't actually need to look up by ID, and creating uuid should be simple. I do think we need to split up into store_id and table.

> For translation: how does sync works? how to make sure that local counter is not modified while new data is pulled from the central server?

this table should only have remote data (data editable only by the site where particular store_id is active), so should never update from central (apart form initial sync when initialising the site)

repository/src/db_diesel/number_row.rs
clemens-msupply: yes, don't know why I didn't do it... e3e5264
The commit also ensures the increment counter method rans in a transaction.

PR #610, opened 2021-11-25
Reuse existing tx on default when doing nested con.transaction() calls

clemens-msupply:
Add transaction_etc method which has the option to enforce a nested
transaction.

This avoids creating unnecessary nested transaction while ensuring that operations are performed in a transaction.

### DIFF
--- a/repository/src/db_diesel/storage_connection.rs
+++ b/repository/src/db_diesel/storage_connection.rs
@@ -65,10 +65,8 @@ impl StorageConnection {
     {
         let current_level = self.transaction_level.get();
         if current_level > 0 && reuse_tx {
-            return match f(self).await {
-                Ok(ok) => Ok(ok),
-                Err(err) => Err(TransactionError::Inner(err)),
-            };
+            let result = f(self).await.map_err(|err| TransactionError::Inner(err))?;
+            return Ok(result);
         }
 
         let con = &self.connection;
--- a/repository/src/db_diesel/storage_connection.rs
+++ b/repository/src/db_diesel/storage_connection.rs
@@ -1,3 +1,5 @@
+use std::cell::Cell;
+
 use super::{get_connection, DBBackendConnection, DBConnection};
 
 use crate::repository_error::RepositoryError;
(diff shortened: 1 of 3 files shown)


repository/src/db_diesel/storage_connection.rs
```diff
@@ -34,11 +44,33 @@ impl From<TransactionError<RepositoryError>> for RepositoryError {
...
+        f: F,
+        reuse_tx: bool,
+    ) -> Result<T, TransactionError<E>>
+    where
+        F: FnOnce(&'a StorageConnection) -> Fut,
+        Fut: Future<Output = Result<T, E>>,
+    {
+        let current_level = self.transaction_level.get();
+        if current_level > 0 && reuse_tx {
+            return match f(self).await {
```
andreievg, line 68: map_err ?

repository/src/db_diesel/storage_connection.rs
```diff
@@ -11,8 +13,16 @@ use futures_util::Future;
 
 pub struct StorageConnection {
     pub connection: DBConnection,
+    /// Current level of nested transaction.
+    /// For example:
+    /// 0 => no transaction
+    /// 1 => in transaction
+    /// 2 => 1st nested transaction
+    /// 3 => 2nd nested transaction
+    pub transaction_level: Cell<i32>,
```
andreievg, line 22: i guess this way you don't need to `mut` ?

andreievg (approved):
I think it's good to add to tests to make sure level is decreased correctly. Also since current_level helps with logging, maybe good to log level in transaction errors

andreievg:
Forgot to ask, do we need async transaction ?

repository/src/db_diesel/storage_connection.rs
clemens-msupply: Yes having it mut would have required quite a lot of changes and not sure if it would have worked, e.g. if we pass the storage connection to multiple services... I also tried an AtomicInt but then went with Cell.

repository/src/db_diesel/storage_connection.rs
clemens-msupply: thanks done 16ab7e4, I manually split it into two lines because it was a bit cryptic...

clemens-msupply:
I think we can remove async transaction, i.e. make the existing repos all sync

PR #612, opened 2021-11-25
Increment the counter value in a transaction

clemens-msupply:
With #610 this doesn't create a nested transaction


clemens-msupply:
Closed because latest change in 514-numbering-table already includes it

PR #616, opened 2021-11-28
602 service provider 3

andreievg:
This is my last attempt.

I first tried to implement something like `ServiceHolder`, which would hold StorageConnection and instance of a service, or would hold StorageConnection and say a reference to service factory (ServiceHolder would be created by ServiceProvider before creating a service from ServiceHolder), this turned out to be quite difficult, and I got stuck with lifetimes. 

Ended up which I think is the simplest option, to have ServiceConnection (which is an instance or reference of StorageConnection).

ServiceOverride is somewhat difficult, since Box<dyn SomeServiceTrait> cannot be made thread safe since it has to hold an instance of StorageConnection, thus instead of storing Box<dyn SomeServiceTrait> in structure that implement ServiceFactoryTrait, had to store Box<dyn Fn() -> Box<dyn SomeServiceTrait>>, and since we override SomeServiceTrait with a closure, and box it, it ended up being quite messy (triple box), thus used a macro. 

Without macro

```rust
        let service_override = ServiceOverride::new().set_location_query_service(Box::new(|| {
            Box::new(TestService(Box::new(|_, _, _| {
                Ok(ListResult {
                    rows: Vec::new(),
                    count: 0,
                })
            })))
        }));
```

With macro

https://github.com/openmsupply/remote-server/blob/ab0c3d2e7b11daba2d9ed7a0b2d00b7eb4eafb1b/server/tests/graphql/locations.rs#L115-L120


clemens-msupply:
Just played around a little bit myself. I though about having the ServiceFactory just holding a list of closures that could be swapped out like in the tests (to reduce the trait boilerplate...) But then had lifetime problems, i.e. somehow you can't specify the closure argument lifetimes without giving the whole ServiceFactory struct a lifetime :(

I m wondering if we could use a Rc to hold the StorageConnection instead of passing a reference around? and if that could simplify the ServiceConnection e.g.

```
type CreateLocation = Box<dyn Fn(Rc<StorageConnection>) -> Box<dyn LocationQueryServiceTrait>>;

pub struct ServiceFactory {
    location_query: CreateLocation,
}
```

Sorry need a break from the review, could chat later/tomorrow if that doesn't make sense

clemens-msupply:
I am a bit torn between "this is so cool we can swap out the service for the tests" and "we are creating a complicated  ServiceFactoryFactory monster" ;-)

clemens-msupply:
As far as I see, it lifetime parameter of the reference is the problem to create a simpler service provider. This is because you can't specify the lifetime of some closure parameters for a closure that is stored in a struct without specifying the lifetime in this struct as well. And this breaks Send+Sync... I managed to remove lifetimes by using  a Rc. Below is my test code which stores the "factory" closures in the ServiceProvider struct . The nice thing is you can use the same ServiceProvider struct for prod and tests.

``` rust
#[derive(Clone)]
struct NewStorageConnection {
    connection: Rc<DBConnection>,
}

impl NewStorageConnection {
    pub fn tx<T, F>(&self, f: F) -> T
    where
        F: FnOnce(NewStorageConnection) -> T,
    {
        f(self.clone())
    }
}

trait TestServiceTrait {
    fn hello(&self);
}

struct TestService {
    con: NewStorageConnection,
}

impl TestServiceTrait for TestService {
    fn hello(&self) {
        println!("Hello from TestService");
    }
}

struct TestService2 {
    con: NewStorageConnection,
}

impl TestServiceTrait for TestService2 {
    fn hello(&self) {
        println!("Hello from TestService2");
    }
}

type CreateService1 = Box<dyn Fn(&NewStorageConnection) -> Box<dyn TestServiceTrait> + Send + Sync>;

struct ServiceProvider {
    service1: CreateService1,
}

// fake gql data method
pub fn data<D: Any + Send + Sync>(data: D) {}

#[cfg(test)]
mod test {
    use crate::{get_storage_connection_manager, test_db};

    use super::*;

    #[test]
    fn test() {
        let settings = test_db::get_test_db_settings("omsupply-...");
        let connection_manager = get_storage_connection_manager(&settings);
        let connection = connection_manager.connection().unwrap();
        let con = NewStorageConnection {
            connection: Rc::new(connection.connection),
        };

        // setup a provider
        let mut provider = ServiceProvider {
            service1: Box::new(|con| Box::new(TestService { con: con.clone() })),
        };

        // nested tx
        let _result = con.tx(|tx| {
            let _service1 = TestService { con: tx.clone() };
            tx.tx(|tx1| {
                let service = (&provider.service1)(&tx1);
                service.hello();

                true
            })
        });

        // Replace service implementation:
        provider.service1 = Box::new(|con| Box::new(TestService2 { con: con.clone() }));
        let service = (&provider.service1)(&con);
        service.hello();

        // Test that provider can be stored in a gql context:
        data(provider);
    }
}
```

andreievg:
> As far as I see, it lifetime parameter of the reference is the problem to create a simpler service provider. This is because you can't specify the lifetime of some closure parameters for a closure that is stored in a struct without specifying the lifetime in this struct as well. And this breaks Send+Sync... I managed to remove lifetimes by using a Rc. Below is my test code which stores the "factory" closures in the ServiceProvider struct . The nice thing is you can use the same ServiceProvider struct for prod and tests.

Hmm looks a bit simpler yeah. Rc would work nicely for StorageConnection. 

If we going for simplicity, I would argue that there is no need for service to store connection in it's state, thus can combine both original service_registry and service_provider

```rust
// ServiceProvider goes in ctx, exposed as ctx.services()
ServiceProvider {
   pub location_service: Box<dyn LocationServiceQuery>,
}

// Init (easy to override in test)
Service Provider {
    location_service: Box::new(LocationService::new())
}

// in Graphql
let connection_manager = match ctx.get_connection_manager().connection() {
   Ok(connection) => connection,
   Err(error) => LocationsResponse::Error(error.into())
}

   match  ctx.services().location_service.get_locations(
            page.map(PaginationOption::from),
            filter.map(LocationFilter::from),
            convert_sort(sort),
            &connection,
        ) {
            Ok(locations) => LocationsResponse::Response(locations.into()),
            Err(error) => LocationsResponse::Error(error.into()),
        }

```

clemens-msupply:
The StorageConnection is kind of an implementation detail so it would be a bit odd having it in the ServiceTrait. We might also want to have something like a request logger in each Service which would be another parameter...

andreievg:
Hmm, i think it might be easier to chat about this in our meeting

service/src/location/query/mod.rs
```diff
@@ -6,21 +6,20 @@ use repository::LocationRepository;
 
 use crate::{get_default_pagination, i64_to_u32, ListError, ListResult, SingleRecordError};
 
-use super::{LocationService, LocationServiceQuery};
+use super::{LocationQueryService, LocationQueryServiceTrait};
```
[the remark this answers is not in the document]
andreievg: i'll change downstream: https://github.com/openmsupply/remote-server/issues/633

server/tests/graphql/locations.rs
```diff
@@ -1,52 +1,62 @@
...
+
+    struct TestService(pub Box<GetLocations>);
+
+    impl LocationQueryServiceTrait for TestService {
         fn get_locations(
             &self,
             pagination: Option<PaginationOption>,
             filter: Option<LocationFilter>,
             sort: Option<LocationSort>,
+            _: &ServiceContext,
```
clemens-msupply, line 32: ServiceContext as the first or last parameter? Would say its more common to put it as the first parameter (like self), e.g. in c you do OOP like that by passing the state as first parameter...

server/tests/graphql/locations.rs
```diff
@@ -1,52 +1,62 @@
...
             &self,
             _: String,
+            _: &ServiceContext,
         ) -> Result<domain::location::Location, service::SingleRecordError> {
             todo!()
         }
     }
 
+    impl TestService {
+        pub fn service_provider(
```
clemens-msupply, line 47: Found it confusing that the service creates a service provider. Maybe `service_provider(service: TestService,...)`?

clemens-msupply (approved):
Looks good! just some small comments

server/tests/graphql/locations.rs
andreievg: I thought, it's best to put variable parameters first, since they are read first and they vary, to me it's kind of easy to ignore or pay less attention to something at the end rather than the start. Can change though if you feel strong about it.

server/tests/graphql/locations.rs
andreievg: I've propagated the changes through the branches now, will do in this one: https://github.com/openmsupply/remote-server/issues/633

server/tests/graphql/locations.rs
clemens-msupply: don't mind much, but think its more common to put it first, e.g. C apis/ drivers, Android View classes also take it as a first parameter...

andreievg: and async-graphql, added: https://github.com/openmsupply/remote-server/issues/633

PR #628, opened 2021-11-29
569 outbound service item

clemens-msupply:
Just the service part.

Is location a thing for service lines?
Shouldn't we have a separate table for service lines? the existing row doesn't seem a good fit since many fields are unused for service lines...

### DIFF
--- a/service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
+++ b/service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
@@ -21,21 +21,6 @@ pub fn delete_outbound_shipment_service_line(
             let stock_line_id_option = line.stock_line_id.clone();
 
             InvoiceLineRowRepository::new(&connection).delete(&line.id)?;
-
-            if let Some(stock_line_id) = stock_line_id_option {
-                let invoice_repository = InvoiceRepository::new(&connection);
-                let stock_line_repository = StockLineRowRepository::new(&connection);
-
-                let mut stock_line = stock_line_repository.find_one_by_id(&stock_line_id)?;
-                stock_line.available_number_of_packs += line.number_of_packs;
-
-                let invoice = invoice_repository.find_one_by_id(&line.invoice_id)?;
-                if invoice.status == InvoiceRowStatus::Confirmed {
-                    stock_line.total_number_of_packs += line.number_of_packs;
-                }
-
-                stock_line_repository.upsert_one(&stock_line)?;
-            }
             Ok(line)
         })
         .map_err(
--- a/service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
+++ b/service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
@@ -1,0 +1,80 @@
+use crate::WithDBError;
+use domain::outbound_shipment::DeleteOutboundShipmentLine;
+use repository::{
+    schema::InvoiceRowStatus, InvoiceLineRowRepository, InvoiceRepository, RepositoryError,
+    StockLineRowRepository, StorageConnectionManager, TransactionError,
+};
+
+mod validate;
+
+use validate::validate;
+
+pub fn delete_outbound_shipment_service_line(
+    connection_manager: &StorageConnectionManager,
+    input: DeleteOutboundShipmentLine,
+) -> Result<String, DeleteOutboundShipmentServiceLineError> {
+    let connection = connection_manager.connection()?;
+
+    let line = connection
+        .transaction_sync(|connection| {
+            let line = validate(&input, &connection)?;
+            let stock_line_id_option = line.stock_line_id.clone();
+
+            InvoiceLineRowRepository::new(&connection).delete(&line.id)?;
+
+            if let Some(stock_line_id) = stock_line_id_option {
+                let invoice_repository = InvoiceRepository::new(&connection);
+                let stock_line_repository = StockLineRowRepository::new(&connection);
+
+                let mut stock_line = stock_line_repository.find_one_by_id(&stock_line_id)?;
+                stock_line.available_number_of_packs += line.number_of_packs;
+
+                let invoice = invoice_repository.find_one_by_id(&line.invoice_id)?;
+                if invoice.status == InvoiceRowStatus::Confirmed {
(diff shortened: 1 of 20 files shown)


service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    let connection = connection_manager.connection()?;
+
+    let line = connection
+        .transaction_sync(|connection| {
+            let line = validate(&input, &connection)?;
+            let stock_line_id_option = line.stock_line_id.clone();
+
+            InvoiceLineRowRepository::new(&connection).delete(&line.id)?;
+
+            if let Some(stock_line_id) = stock_line_id_option {
```
andreievg, line 25: should never really exist, i wonder if this logic is worth having ?

service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    Ok(line.id)
+}
+
+pub enum DeleteOutboundShipmentServiceLineError {
+    LineDoesNotExist,
+    DatabaseError(RepositoryError),
+    InvoiceDoesNotExist,
+    NotAnOutboundShipment,
+    NotThisStoreInvoice,
+    ItemNotFound,
```
andreievg, line 58: This one is because of check_item helper ?

service/src/invoice_line/outbound_shipment_service_line/insert/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+        )?;
+    Ok(new_line.id)
+}
+
+pub enum InsertOutboundShipmentServiceLineError {
+    LineAlreadyExists,
+    DatabaseError(RepositoryError),
+    InvoiceDoesNotExist,
+    NotAnOutboundShipment,
+    //NotThisStoreInvoice,
```
andreievg, line 49: maybe add todo

service/src/invoice_line/outbound_shipment_service_line/insert/validate.rs
```diff
@@ -0,0 +1,64 @@
...
+    input: &InsertOutboundShipmentServiceLine,
+    connection: &StorageConnection,
+) -> Result<(ItemRow, InvoiceRow), InsertOutboundShipmentServiceLineError> {
+    check_line_does_not_exists(&input.id, connection)?;
+    let item = check_item(&input.item_id, connection)?;
+    if item.r#type != ItemType::Service {
+        return Err(InsertOutboundShipmentServiceLineError::NotAServiceItem);
+    }
+    let invoice = check_invoice_exists(&input.invoice_id, connection)?;
+    // check_store(invoice, connection)?; InvoiceDoesNotBelongToCurrentStore
```
andreievg, line 29: would be good as todo

service/src/invoice_line/outbound_shipment_service_line/update/generate.rs
```diff
@@ -0,0 +1,128 @@
+use repository::schema::{InvoiceLineRow, ItemRow};
+
+use super::{UpdateOutboundShipmentServiceLine, UpdateOutboundShipmentServiceLineError};
+
+pub fn generate(
+    input: UpdateOutboundShipmentServiceLine,
+    existing_line: InvoiceLineRow,
+    item: ItemRow,
+) -> Result<InvoiceLineRow, UpdateOutboundShipmentServiceLineError> {
+    // 1) Use name from input (if specified)
```
andreievg, line 10: As per our convo, i think we can do similar to insert, unwrap name or use item.name

service/src/invoice_line/outbound_shipment_service_line/update/generate.rs
```diff
@@ -0,0 +1,128 @@
...
+    // 3) else: use existing line name
+    let item_name = if let Some(input_name) = input.name {
+        input_name
+    } else if input.item_id.is_some() && input.item_id != Some(existing_line.item_id.to_owned()) {
+        item.name
+    } else {
+        existing_line.item_name
+    };
+
+    let update_line = InvoiceLineRow {
```
andreievg, line 21: i think it's simpler to use existing_line

andreievg (approved):
Looks good, didn't really look at the test. Will check it out when all tests are done.

I think we would need to add item_type check normal line mutations

andreievg:
> Is location a thing for service lines?

nope

> Shouldn't we have a separate table for service lines? the existing row doesn't seem a good fit since many fields are unused for service lines...

As per our convo, too much translation work (but yeah, a better way)

service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
clemens-msupply: DeleteOutboundShipmentServiceLineError

service/src/invoice_line/outbound_shipment_service_line/insert/validate.rs
clemens-msupply: done

service/src/invoice_line/outbound_shipment_service_line/update/generate.rs
clemens-msupply: You mean editing the input? I just tried it but it feels a bit unsafe, for example, if new fields are added we might miss to update it here. I left it as it is for now.

service/src/invoice_line/outbound_shipment_service_line/update/generate.rs
clemens-msupply: I think then name would get updated even though neither name nor item has changed

service/src/invoice_line/outbound_shipment_service_line/delete/mod.rs
clemens-msupply: thanks, removed in 38492bb

service/src/invoice_line/outbound_shipment_service_line/update/generate.rs
andreievg: Oki.

I mean destructure the input and mutate existing_line

Issue #620, opened 2021-11-29
Update Location Mutation


andreievg:
https://github.com/openmsupply/remote-server/issues/544

PR #629, opened 2021-11-29
620 update location mutation

andreievg:
Closes: #620 

Pretty straight forward, copy paste really

### DIFF
--- a/repository/src/db_diesel/storage_connection.rs
+++ b/repository/src/db_diesel/storage_connection.rs
@@ -22,7 +22,7 @@ pub enum TransactionError<E> {
 }
 
 impl<E> TransactionError<E> {
-    pub fn convert(self) -> E
+    pub fn to_inner_error(self) -> E
     where
         E: From<RepositoryError>,
     {
--- a/server/tests/graphql/location_update.rs
+++ b/server/tests/graphql/location_update.rs
@@ -1,10 +1,13 @@
 mod graphql {
-    use crate::graphql::{assert_gql_query, ServiceOverride};
+    use crate::graphql::assert_gql_query;
     use domain::location::{Location, UpdateLocation};
-    use repository::mock::MockDataInserts;
+    use repository::{mock::MockDataInserts, StorageConnectionManager};
     use serde_json::json;
     use server::test_utils::setup_all;
-    use service::location::update::{UpdateLocationError, UpdateLocationServiceTrait};
+    use service::{
+        location::update::{UpdateLocationError, UpdateLocationServiceTrait},
+        service_provider::{ServiceContext, ServiceProvider},
+    };
 
     type UpdateLocationMethod =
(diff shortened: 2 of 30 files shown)


server/tests/graphql/location_update.rs
```diff
@@ -0,0 +1,233 @@
...
+        "#;
+
+        let variables = Some(json!({
+          "input": {
+            "id": "n/a",
+
+          }
+        }));
+
+        // Record Already Exists
```
clemens-msupply, line 205: Is the comment correct?

service/src/location/update/validate.rs
```diff
@@ -0,0 +1,47 @@
...
+    connection: &ServiceConnection,
+) -> Result<LocationRow, UpdateLocationError> {
+    let location_option = LocationRowRepository::new(connection).find_one_by_id(id)?;
+
+    if let Some(location) = location_option {
+        Ok(location)
+    } else {
+        Err(UpdateLocationError::LocationDoesNotExist)
+    }
+}
```
clemens-msupply, line 35: Keep it simple: just inline it, the method doesn't do anything

andreievg: It makes the calling method simpler. I think what's been done in the draft pr you've made and in Location Mutation tidy up PR works quite well. I wouldn't like  jumping down to repository level abstraction in higher level validation method, so I don't agree with `just line it` (I like how the draft PR was done, even if it means one line `check` method). Also `keep it simple` wasn't the argument for changing to inline matches (btw there are no froms here), the argument was visibility of what check methods do by exposing what types are returned, i am not sure if you can get any simpler.

andreievg: This was refactored in: https://github.com/openmsupply/remote-server/pull/635

server/tests/graphql/location_update.rs
andreievg: thanks for noticing, removed in downstream branch: af98d57871dc9c6a7b1771e8d01385d0d0745d46

repository/src/db_diesel/storage_connection.rs
```diff
@@ -22,7 +22,7 @@ pub enum TransactionError<E> {
 }
 
 impl<E> TransactionError<E> {
-    pub fn convert(self) -> E
+    pub fn to_inner_error(self) -> E
```
clemens-msupply, line 25: There is a RepositoryError::TransactionError which could be used instead of RepositoryError::DBError in the body of to_inner_error (if already merged)...

clemens-msupply (approved):

repository/src/db_diesel/storage_connection.rs
andreievg: I remembered that but was behind master in upstream branches, seemed too much effort, added to:https://github.com/openmsupply/remote-server/issues/633

thanks for reminder

PR #630, opened 2021-11-30
569 gql outbound service line

clemens-msupply:


graphql/src/schema/mutations/outbound_shipment/batch.rs
```diff
@@ -27,6 +33,12 @@ pub struct BatchOutboundShipmentResponse {
     insert_outbound_shipment_lines: Option<Vec<MutationWithId<InsertOutboundShipmentLineResponse>>>,
     update_outbound_shipment_lines: Option<Vec<MutationWithId<UpdateOutboundShipmentLineResponse>>>,
     delete_outbound_shipment_lines: Option<Vec<MutationWithId<DeleteOutboundShipmentLineResponse>>>,
+    insert_outbound_shipment_service_lines:
```
andreievg, line 36: I was thinking about our transaction thinking for this, i think it should be done within a service (basically moving most of this logic to service). Or service provider can give service_context within closure (and here it could be the case where transaction is beneficial in graphql, although I still think a service can do this, and call other services)

graphql/src/schema/mutations/outbound_shipment/service_line/insert.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+pub fn get_insert_outbound_shipment_service_line_response(
+    connection_manager: &StorageConnectionManager,
+    input: InsertOutboundShipmentServiceLineInput,
+) -> InsertOutboundShipmentServiceLineResponse {
+    use InsertOutboundShipmentServiceLineResponse::*;
+    let id = match insert_outbound_shipment_service_line(
+        connection_manager,
+        InsertOutboundShipmentServiceLine {
```
andreievg, line 44: can potentially have a phantom input field (going back to why i was destructuring input)

andreievg (approved):
looks good thanks

graphql/src/schema/mutations/outbound_shipment/service_line/insert.rs
clemens-msupply: Ok done, in this case I thought it will be unlikely but probably safer to do it anyway

graphql/src/schema/mutations/outbound_shipment/batch.rs
clemens-msupply: Maybe we explicitly wrap everything that should be in a transaction in a service into a transaction. With the flattening of nested transaction this shouldn't create any overhead in case we need to use multiple services in parent transaction. e.g.

```rust
fn gql_endpoint() {
   // fine because do_work takes care that logic runs in a tx
   service1.do_work();
}

fn gql_endpoint_2() {
   con.tx(|con| {
       // service1 doesn't create a nested transaction because already in a tx
       service1.do_work();
       service2.do_work();
   });
}
```

The UserAccountService::create_user actually already works like that...

Issue #621, opened 2021-11-29
Delete Location Mutation


andreievg:
https://github.com/openmsupply/remote-server/issues/544

PR #631, opened 2021-11-30
621 delete location mutation

andreievg:
Closes: #621 

Pretty straight forward. You'll note that LocationInUse error return stock lines and invoice lines that are using location.

Also renamed `LocationDoesNotExists` to `LocationDoesNotExist` globally

### DIFF
--- a/graphql/src/schema/mutations/mod.rs
+++ b/graphql/src/schema/mutations/mod.rs
@@ -5,7 +5,7 @@ pub mod location;
 pub mod outbound_shipment;
 pub mod user_register;
 
-use self::location::{InsertLocationInput, InsertLocationResponse};
+use self::location::{InsertLocationInput, InsertLocationResponse, UpdateLocationInput, UpdateLocationResponse};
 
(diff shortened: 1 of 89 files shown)


graphql/src/schema/mutations/mod.rs
```diff
@@ -68,6 +69,26 @@ impl Mutations {
...
+            Err(error) => return DeleteLocationResponse::Error(error.into()),
+        };
+
+        match service_provider
+            .delete_location_service
+            .delete_location(input.into(), &service_context)
+        {
+            Ok(location_id) => DeleteLocationResponse::Response(DeleteResponse(location_id)),
+            Err(error) => DeleteLocationResponse::Error(error.into()),
+        }
```
clemens-msupply, line 89: Code could be moved into location/deletion.rs ...

clemens-msupply (approved):
LGTM

graphql/src/schema/mutations/mod.rs
andreievg: indeed, i am glad i moved it in downstream branch

PR #632, opened 2021-12-01
Use alternative validate error handling

clemens-msupply:
- For this example all needed check functions are inlined and updated if
needed
- All From, except of `impl From<RepositoryError> for
UpdateInboundShipmentLineError` are removed

### DIFF
--- a/service/src/invoice_line/inbound_shipment_line/update/mod.rs
+++ b/service/src/invoice_line/inbound_shipment_line/update/mod.rs
@@ -1,4 +1,3 @@
-use crate::WithDBError;
 use domain::inbound_shipment::UpdateInboundShipmentLine;
 use repository::{
     InvoiceLineRowRepository, RepositoryError, StockLineRowRepository, StorageConnectionManager,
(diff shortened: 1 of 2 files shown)


andreievg (commented):
This looks good,

I like how you've kept methods with one lines (even though they could be inlined), since it keeps the main validation method consistent and easy to read.  

Although number_of_packs_below_one and pack_size_below one takes an option but check_item does not. Any reason to do it differently ? (would prefer check methods taking option tbh). 

Also some returns are truthy for error and some are falsy for error, would prefer consistency if possible.

clemens-msupply:
Yes good point the methods should all be either like
`if !check_... return error` or something like `if not_ok return error`

The check_item_option method was a bit tricky because it it returned `ItemNotFound | Some | None` (with None not an error case) and I decided to avoid it by inlining it. In the end I though doing it explicitly makes it clearer what check_item_option does since its quite hard to encode it in the method name or in a comment.  I didn't put too much thought into it though...

andreievg:
I think we agreed to go with this way, it's one of the tasks in https://github.com/openmsupply/remote-server/issues/749

PR #640, opened 2021-12-03
569 invoice lines taxes

clemens-msupply:
This makes it possible to insert and update total_before_tax, total_after_tax and tax for all invoice lines.

I removed the existing automatic calculation for total_after_tax since frontend is responsible for it.


domain/src/outbound_shipment.rs
```diff
@@ -34,6 +39,9 @@ pub struct UpdateOutboundShipmentLine {
     pub item_id: Option<String>,
     pub stock_line_id: Option<String>,
     pub number_of_packs: Option<u32>,
+    pub total_before_tax: Option<f64>,
```
andreievg, line 42: wouldn't this be calculated from lines ?

domain/src/inbound_shipment.rs
```diff
@@ -20,6 +22,9 @@ pub struct UpdateInboundShipment {
     pub comment: Option<String>,
     pub their_reference: Option<String>,
     pub color: Option<String>,
+    pub total_before_tax: Option<f64>,
+    pub total_after_tax: Option<f64>,
+    pub tax: Option<ShipmentTaxUpdate>,
```
andreievg, line 27: hmm, should this just be `taxPercentage: Option<f64>` ?

andreievg (approved):
Looks good, but why have totals and tax in shipment inputs ? And alos why ShipentTaxUpdate vs just f64 ?

domain/src/inbound_shipment.rs
clemens-msupply: Sorry should have put a comment in. I need to:
1) Able to unset the tax if it is currently set -> `Some(ShipmentTaxUpdate { precentage: None})`
2) Update a tax -> `Some(ShipmentTaxUpdate { precentage: value})`
3) Keep existing tax -> `None`

A single Option can't do that...

domain/src/outbound_shipment.rs
clemens-msupply: Removed invalid fields from non-line shipment mutations in 62c14f4

domain/src/inbound_shipment.rs
andreievg: I kind of think we need a generic nullable api type for this kind of input, i've encountered this issue in a couple of places (for expiry date, batch etc.. which are optional fields in database, but in update mutation we only just have them as Optional<value>, so there is no way to set them to null). So we kind of need Option<Option<T>>

PR #641, opened 2021-12-05
569 invoice lines tax stats

clemens-msupply:
The lines stats could probably be moved completely into the View but I found it easier to have a relatively simple View and do some post processing afterwards. What do you think?


andreievg (approved):
Looks good, but I would have used `filter` of aggregates (works both in sqlite and postgres)

```SQL

SELECT
	invoice_line.invoice_id,
    item.type as item_type,
    SUM(invoice_line.total_before_tax) AS total_before_tax,
    SUM(invoice_line.total_after_tax) AS total_after_tax,
	SUM(invoice_line.total_after_tax) FILTER(WHERE item.type = 'SERVICE') AS total_service_after_tax ,
	SUM(invoice_line.total_before_tax) FILTER(WHERE item.type = 'SERVICE') AS total_service_before_tax ,
	SUM(invoice_line.total_after_tax) FILTER(WHERE item.type = 'STOCK')  AS total_stock_after_tax  ,
	SUM(invoice_line.total_before_tax) FILTER(WHERE item.type = 'STOCK') AS total_stock_before_tax 
FROM
	invoice_line 
	LEFT JOIN item ON (invoice_line.item_id = item.id)
GROUP BY
	invoice_line.invoice_id;
```

clemens-msupply:
Thanks, the sql works. I made the following changes:
- I updated the invoice stats view
- Renamed invoice line stats to invoice stats
- Removed the InvoicePricing domain object and use the raw data as a domain object (as discussed)

graphql/src/schema/types/invoice_query.rs
```diff
@@ -221,24 +227,44 @@ impl From<Invoice> for InvoiceNode {
 
 // INVOICE LINE PRICING
 pub struct InvoicePricingNode {
-    invoice_pricing: InvoicePricing,
+    invoice_pricing: InvoiceStatsRow,
```
andreievg, line 230: Thanks, I think this pattern will reduce number of changes, although would still be good to have an intermediary object with composition (not for this simple node, but for more complex ones)

andreievg (approved):

graphql/src/schema/types/invoice_query.rs
clemens-msupply: Agree!

PR #643, opened 2021-12-06
Enable ilike (case insensitive like) for Postgres

### LINKED ISSUE #604: Ignore case in text filters
I thought we'd been here before, but turns out that was #555!

I find it super annoying when searching for items that I have to match the case of items in the database; searching for 'acetylsalicylic acid' when adding items to a shipment, if I type `ace` into the autocomplete, I get `Paracetamol` only. I have to type `Ace` to get what I want!

Could you please ignore case when filtering, I think this is a more natural usage.

clemens-msupply:
Use ilike for postgres (works for ASCII and UTF8 chars, e.g. tested with Ä <--> ä ).

Sqlite is case insensitive on default. However, it only works for ASCII chars. Using something like ICU or providing a custom UPPER method is a bit more complicated (e.g. https://forum.lazarus.freepascal.org/index.php?topic=34259.0 has some solution for Pascal).

Do we need to filter for non ASCII char?

Closes #604

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ openssl req -x509 -newkey rsa:4096 -nodes -keyout certs/key.pem -out certs/cert.
 
 ## Dependencies
 
+When using Postgres, Postgres 12 or higher is required.
+
 ### Windows
 
 - Install [rustup.exe](https://www.rust-lang.org/tools/install), following the instructions for installing Visual Studio C++ Build Tools if prompted.
--- a/repository/src/diesel_macros.rs
+++ b/repository/src/diesel_macros.rs
@@ -50,6 +50,7 @@ macro_rules! apply_equal_filter {
 ///     }
 /// }
 /// ```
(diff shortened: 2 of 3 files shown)


mark-prins (approved):
Sorry for the delay in reviewing! I've had to install postgres 12 (as well as switch mSupply branch etc..)

Change looks good to me! Thanks 👍 

One thing I'm curious about, is why the file `00000097_no_case_collation/up.sql` doesn't show in the review. Because the collation is using `deterministic=false`, the required postgres version is now 12+ As I understand it, without the collation we can still have case insensitivity, as you've used `ilike`, but not the comparison of accents and true unicode equivalence. That was something I was wondering about, we've only struck it so far with French, likely to appear in other supported languages though, so great to have the support for it!

Anyway - given that migration fails if postgres is < 12, could you also add a note to the docs or readme to let users know of the minimum requirement please?

clemens-msupply:
Thanks I added a comment in the readme.

PR #646, opened 2021-12-08
Add a general graphql ServerError

clemens-msupply:
Some playing around with general gql errors

### DIFF
--- a/graphql/src/schema/queries/invoice_counts.rs
+++ b/graphql/src/schema/queries/invoice_counts.rs
@@ -1,7 +1,10 @@
 use async_graphql::*;
 use chrono::Utc;
-use service::dashboard::invoice_count::{InvoiceCountService, InvoiceCountServiceTrait};
+use service::dashboard::invoice_count::{
+    InvoiceCountError, InvoiceCountService, InvoiceCountServiceTrait,
+};
 
+use crate::errors::ServerError;
 use crate::schema::types::invoice_query::InvoiceNodeType;
 use crate::ContextExt;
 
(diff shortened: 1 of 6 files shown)


graphql/src/schema/queries/invoice_counts.rs
```diff
@@ -25,33 +28,27 @@ pub fn invoice_counts(
...
-                extensions: None,
-            },
-        })?;
+    let result = service.invoice_count_created(
+        &service_ctx,
+        invoice_type.into(),
+        Utc::now(),
+        timezone_offset,
+    );
+    let created = match result {
```
andreievg, line 41: I think it would be quite simple to just stringify debug output of the error (I thought that's what you actually suggested from the start, something like this: https://github.com/openmsupply/remote-server/compare/579-improved-general-errors...579-Example-of-using-debug-stringify-for-standard-error, I had to change ServerError to StandardError in that change, was having conflicts with async_graphql::* import, which includes build in ServerError)

https://github.com/openmsupply/remote-server/blob/192a39edab8f850f49d214119d5e11d68b6f5e3b/graphql/src/schema/mutations/location/delete.rs#L17-L56

clemens-msupply: Your branch looks good! I would leave StructuredError as Error in the returned Union though.

We might need to be a bit careful with returning the stringified errors, e.g. if the error contains some sensible data or leaks some other internal data that shouldn't be public... I don't think that is the case right now but probably good to keep in mind...

andreievg: > error contains some sensible data

Good point, especially if we storing some other error inside the error (like repository error), will keep that in mind

> I would leave StructuredError as Error in the returned Union though.

Sure

I think GraphqlError might even be a better name then StandardError

andreievg:
@clemens-msupply what would be the next step for this ? I want to use this pattern going forward and working on a PR where i want to use it

andreievg:
I wanted to use the errors in this [pr](https://github.com/openmsupply/remote-server/pull/653), so added it there

clemens-msupply:
I approved your PR so I close this one

PR #648, opened 2021-12-09
Mock data collection

clemens-msupply:
I tried to use a simple HashMap for the MockDataCollection but I need to write mock data in a fixed order to the DB. For this reason MockDataCollection looks a little bit complicated since I made it behave like a HashMap... Otherwise it should be quite straight forward. Currently there are three MockData entries in the collection, the base MockData and mock data for two tests.

This approach could actually also make it easier to find test specific mock data in a test (e.g. mock_data["my-test"].names[0])

### DIFF
--- a/repository/src/mock/invoice.rs
+++ b/repository/src/mock/invoice.rs
@@ -258,48 +258,6 @@ pub fn mock_outbound_shipment_invalid_stock_line() -> InvoiceRow {
     }
 }
 
-// Added for invoice_count_service_test
-pub fn mock_inbound_shipment_invoice_count_service_a() -> InvoiceRow {
-    InvoiceRow {
-        id: String::from("inbound_shipment_invoice_count_a"),
-        name_id: String::from("name_store_b"),
-        store_id: String::from("store_a"),
-        invoice_number: 4,
-        r#type: InvoiceRowType::InboundShipment,
-        status: InvoiceRowStatus::New,
-        on_hold: false,
-        comment: Some("Sort comment test Ac".to_owned()),
-        their_reference: Some(String::from("")),
-        created_datetime: NaiveDate::from_ymd(2021, 12, 7).and_hms_milli(13, 30, 0, 0),
-        allocated_datetime: None,
-        picked_datetime: None,
-        shipped_datetime: None,
-        delivered_datetime: None,
-        verified_datetime: None,
-        color: None,
-    }
-}
-pub fn mock_inbound_shipment_invoice_count_service_b() -> InvoiceRow {
-    InvoiceRow {
-        id: String::from("inbound_shipment_invoice_count_b"),
-        name_id: String::from("name_store_b"),
-        store_id: String::from("store_a"),
-        invoice_number: 4,
-        r#type: InvoiceRowType::InboundShipment,
-        status: InvoiceRowStatus::New,
-        on_hold: false,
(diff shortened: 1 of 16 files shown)


andreievg:
This looks good, although I can't quite figure out the benefit, let's try to break it down;

With the addition of full_invoice and full_master_list, `mock_full_invoices` and `mock_full_master_list` were made to return hash map, so could access test specific entries:

https://github.com/openmsupply/remote-server/blob/8f294b1c5bf977ac3eb97e475c0fd7168897c22a/server/tests/graphql/master_lists.rs#L93-L97

https://github.com/openmsupply/remote-server/blob/8f294b1c5bf977ac3eb97e475c0fd7168897c22a/server/tests/graphql/outbound_shipment_update.rs#L233-L236

So the idea was, you add test data to just those files and easily access them. i.e. if `inbound_shipment_invoice_count_a` was added to full_invoice, it can be accessed with:

```rust
 let full_invoice = mock_data 
     .full_invoices 
     .get("inbound_shipment_invoice_count_a") 
     .unwrap(); 
```

With the way in this PR, it would be accessed with

```rust
let full_invoice = mock_data["test_invoice_count_service_data"].invoices[0];
```

And would need to be added to both the invoices mock and mock mod.rs (I kind of thought we are trying to minimise number of files we need to edit).

> This approach could actually also make it easier to find test specific mock data in a test (e.g. mock_data["my-test"].names[0])

As per above, I think it would be better to make sure that mock_names() returns a hash map, then can access exactly the name that's needer (rather then by index)

```rust
 let name = mock_data 
     .names 
     .get("my-test") 
     .unwrap(); 
```

Although I think that is an overkill, i realise that it's better to just use the method that creates that mock data entity:

```
let name1 = mock_name_name_service_test1();
let name2 = mock_name_name_service_test2();
```
vs
```
let name1 =  mock_data["name_service"].names[0]);
let name2 =  mock_data["name_service"].names[1]);
```

Another possible issue, if we are creating test specific set's of data, we have to making sure relation dependencies are met. i.e. special test invoice and names are added, and another test want's to use those names in the data that's added.

I'll approve this, but I think ideally we want this:

* One method to insert all mock data (that always inserts all mock data)
* Remove the hashmap from full invoice and full master list and just access them via methods that make the data (this would be soo much quicker to view the actual data in the test, just go to definition)

```rust
let mock_data_lines = master_list_master_list_line_filter_test().lines
```
vs
```rust
 let mock_data_lines = &mock_data 
     .full_master_list 
     .get("master_list_master_list_line_filter_test") 
     .unwrap() 
     .lines; 
```

clemens-msupply:
Sorry think I missed to describe the main motivation. I basically just want to put all test data for one test into one place.  For example, test_outbound_shipment_update.rs in the PR contains three different types of mock data that would otherwise be spread over multiple files. I think grouping the data would makes it easier to setup and understand a test since all relevant data is in one place and you don't have to jump around.

Yes agree just accessing the data by id or using the create method directly, is easier than `mock_data["my-data"].names.get(0)` I could either merge multiple MockData into a single big MockData or as you said the tests just create new instances of the required piece of data.

Also acknowledge that it involves some more editing of files. To reduce this we could, for example, put all mock data for invoices into one file and this file produces a list of MockData (instead of having a mock file for each test).

And yes all mock data should end up in one big dataset.

Lets chat on Mon

PR #649, opened 2021-12-09
579 outbound shipment counts

clemens-msupply:
Restructured the endpoint a little bit. The gql errors make it quite easy :) (still using temp gql errors, will update later)

### DIFF
--- a/service/src/dashboard/invoice_count.rs
+++ b/service/src/dashboard/invoice_count.rs
@@ -9,30 +9,32 @@ use repository::{InvoiceQueryRepository, RepositoryError};
 
 use crate::service_provider::ServiceContext;
 
-pub struct InvoiceCountCreated {
-    pub today: i64,
-    pub this_week: i64,
-}
-
 #[derive(Debug)]
 pub enum InvoiceCountError {
     RepositoryError(RepositoryError),
     BadTimezoneOffset,
 }
 
+pub enum CountTimeRange {
+    Today,
+    ThisWeek,
+}
+
 pub trait InvoiceCountServiceTrait {
-    /// Returns number of created invoices
+    /// Returns number of invoices of a certain status for today
     ///
     /// Arguments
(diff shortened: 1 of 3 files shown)


service/src/dashboard/invoice_count.rs
```diff
@@ -70,52 +78,92 @@ fn start_of_week(datetime: NaiveDateTime) -> NaiveDateTime {
...
+        range: &CountTimeRange,
+        now: &DateTime<Utc>,
+        timezone_offset: &Option<i32>,
+    ) -> Result<i64, InvoiceCountError> {
         let repo = InvoiceQueryRepository::new(&ctx.connection);
         let tz = offset_to_timezone(timezone_offset).ok_or(InvoiceCountError::BadTimezoneOffset)?;
         let now = to_local(now, &tz);
+        let oldest = match range {
+            CountTimeRange::Today => {
+                to_utc(&start_of_day(&now), &tz).ok_or(InvoiceCountError::BadTimezoneOffset)?
```
andreievg, line 133: should this be to_naive_datetime ?

andreievg (approved):
I wonder if we need the timezone_offset logic, or if we can just outsource the date time selections to front end (there are great libraries in js like moment which makes it very easy to get start of day end of day, start of week and end of week). I'll try to make a comment on the issue with suggestion.

I think it would be much easier to follow service logic if we keep to one level of abstraction i.e. `start_of_day(timezone_offset)` <- which returns NaiveDateTime option.

Also i really like doing default derives on trait, since when they are used in test you only need to override one (rather then all trait methods)

service/src/dashboard/invoice_count.rs
clemens-msupply: its a bit tricky and I got confused myself for which reason I tried to make it clear through the type system...

Regarding timezone calculation, the client would have to provide times for today, start_of_this_week, etc... Or as in the chat, we let the client do multiple calls with the default api...

+1 for default derives

PR #652, opened 2021-12-09
650 add master list query

### LINKED ISSUE #650: Add master lists query
https://github.com/openmsupply/remote-server/issues/639

andreievg:
Closes #650 

* Existing master list repository renamed to MasterListRowRepository (sorry this didn't show that well in git diff), so can ignore diff in MasterListRowRepository
* Added domain, service, repository, graphql
* I found name_table inconsistent with other dsl, so renamed it to name, but the name field of name table had to be changed in dsl to name_
* For filtering master lists by names, added a query within create_filtered_query, this is tested in service.
* Relocated OrderByExtensions to be imported in apply_sort_no_case macro, so that it doesn't need to be imported wherr apply_sort_no_case macro is used

Some variations from pattern:

* Used MasterList type as MasterListRow
* Wanted to avoid changes in types and sort_and_filter files, thus put it all in queries/master_list, I think it's easier to follow (btw i also put the responses there too).


clemens-msupply (approved):
Looks good. Only suggestion is to merge all service/src/master_list/* files into one file.

Good gql tests, I plan to write some more gql tests myself! got a bit sloppy

andreievg:
> Only suggestion is to merge all service/src/master_list/* files into one file. 
Hmm, since there is a test there too, i think it's better to keep it in a folder. Could potentially combine query and mod in one file, but i really like consistency with the structure of other services (i think it adds to readability when reader encounters the same patterns)

PR #653, opened 2021-12-09
651 add master list line to master list query

andreievg:
Closes: #651 

I've also added standard graphql error, since needed to avoid structured errors for loaders


graphql/src/schema/types/master_list_line.rs
```diff
@@ -0,0 +1,68 @@
...
+            nodes: Vec::new(),
+        }
+    }
+
+    pub fn from_domain_vec(from: Vec<MasterListLine>) -> MasterListLineConnector {
+        MasterListLineConnector {
+            total_count: usize_to_u32(from.len()),
+            nodes: from
+                .into_iter()
+                .map(MasterListLineNode::from_domain)
```
clemens-msupply, line 34: Looks nice, even though in this case I would have just wrote `|master_list_line| MasterListLineNode { master_list_line }` since it probably only use once here and its even in the same file...

repository/src/db_diesel/master_list_line.rs
```diff
@@ -13,35 +23,92 @@ impl<'a> MasterListLineRepository<'a> {
...
+            f.master_list_id,
+            master_list_line_dsl::master_list_id
+        );
+    }
+
+    Ok(query)
+}
+
+#[cfg(test)]
+mod test {
```
clemens-msupply, line 79: The other repository tests are actually in the parent test directory. The reason was that we initially had multiple repository implementations and all of them should work with the same tests...

clemens-msupply (approved):
Look all good, just one test is in the "wrong" location (unsure if its worth keeping the old structure, see comment)

graphql/src/schema/types/master_list_line.rs
andreievg: Sure, I intuitively like method here, I think it's because it describes why the conversion needs to be made (that it's converting from domain object)

repository/src/db_diesel/master_list_line.rs
andreievg: hmm, for consistency i've relocated e149bc54919181788488be270862aa0fbd1ec252, i kind of think they should be in repositories tbh (i can't quite remember our decision to relocate them, i think it was because import of mock data for the tests were using other repositories, and it was deemed more of an integration test)

clemens-msupply: The reason was that I didn't wanted to test the diesel specific implementation, e.g. the same test should work with sqlx repositories...

PR #655, opened 2021-12-10
Move item query function into its own file

clemens-msupply:


andreievg (approved):
Looks good. Would be great also for Connector<ItemNode> to be a specific struct in queries/item.rs

also i think the following would be removed with additions of standard graphql error (i.e. nested item nodes would just return ItemNode rather then ItemResponse
```
#[derive(Union)]
pub enum ItemResponseError {
    InternalError(InternalError),
}

#[derive(SimpleObject)]
pub struct ItemError {
    pub error: ItemResponseError,
}

#[derive(Union)]
pub enum ItemResponse {
    Error(ItemError),
    Response(ItemNode),
}
```

clemens-msupply:
Use specific connectors for items and names: c0c20e5

Yes these types can be removed with standard errors :)

PR #656, opened 2021-12-13
579 stock expiry

### LINKED ISSUE #579: Dashboard endpoint
The dashboard section is making use of some queries which aren't currently supported.

Screen Shot 2021-11-18 at 2 15 49 PM

This is an area which I expect will be expanded, as we think of new statistics to add.. for now though what we have is 
- stock count, showing stock which is expired and expiring soon ( for this, I've used within 3 months 🤷  )
- invoices
  - number of outbound to be picked i.e. the count of outbound shipments which are `ALLOCATED`
  - number of inbound created today
  - number of inbound created this week

This is just what I've put together and have implemented in the mock API - can share the schema as well, but I thought that's probably not required? Names and shapes are up for discussion of course - the requirement is just to have something to show on the dashboard. I've used three separate calls, so that each dashboard area can be a separate component.

### Query
```
query invoiceCounts($type: InvoiceNodeType!) {
  invoiceCounts(type: $type) {
    ... on InvoiceCountsConnector {
      __typename
      created {
        today
        thisWeek
      }
      toBePicked
    }
    ... on ConnectorError {
      __typename
      error {
        description
      }
    }
  }
}
```
### Response
With the variables as `{ "type": "OUTBOUND_SHIPMENT" }` the response is:

```
{
    "data": {
        "invoiceCounts": {
            "__typename": "InvoiceCountsConnector",
            "toBePicked": 27
        }
    }
}
```

With the variables as `{ "type": "INBOUND_SHIPMENT" }` the response is:

```
{
    "data": {
        "invoiceCounts": {
            "__typename": "InvoiceCountsConnector",
            "created": {
                "today": 0,
                "thisWeek": 0
            }
        }
    }
}
```

### Query
```
query stockCounts {
  stockCounts {
    ... on StockCountsConnector {
      __typename
      expired
      expiringSoon
    }
    ... on ConnectorError {
      __typename
      error {
        description
      }
    }
  }
}
```

### Response
```
{
    "data": {
        "stockCounts": {
            "__typename": "StockCountsConnector",
            "expired": 0,
            "expiringSoon": 189
        }
    }
}
```

clemens-msupply:
Add count_stock endpoint and refactor count_invoices a little bit.

Closes #579

### DIFF
--- a/graphql/src/schema/queries/invoice_counts.rs
+++ b/graphql/src/schema/queries/invoice_counts.rs
@@ -1,9 +1,7 @@
 use async_graphql::*;
 use chrono::{FixedOffset, Utc};
 use domain::invoice::{InvoiceStatus, InvoiceType};
-use service::dashboard::invoice_count::{
-    CountTimeRange, InvoiceCountError, InvoiceCountService, InvoiceCountServiceTrait,
-};
+use service::dashboard::invoice_count::{CountTimeRange, InvoiceCountError};
 use util::timezone::offset_to_timezone;
 
 use crate::{standard_graphql_error::StandardGraphqlError, ContextExt};
@@ -17,7 +15,7 @@ fn do_invoice_count(
 ) -> Result<i64> {
     let service_provider = ctx.service_provider();
     let service_ctx = service_provider.context()?;
-    let service = InvoiceCountService {};
+    let service = &service_provider.invoice_count_service;
     let count = service
         .invoices_count(
             &service_ctx,
@@ -88,7 +86,7 @@ impl OutboundInvoiceCounts {
             source: None,
             extensions: None,
         })?;
-        let service = InvoiceCountService {};
+        let service = &service_provider.invoice_count_service;
         let to_by_picked = service
             .outbound_invoices_pickable_count(&service_ctx)
             .map_err(|_| Error {
--- a/graphql/src/schema/queries/invoice_counts.rs
+++ b/graphql/src/schema/queries/invoice_counts.rs
@@ -1,25 +1,22 @@
 use async_graphql::*;
-use chrono::Utc;
+use chrono::{FixedOffset, Utc};
 use domain::invoice::{InvoiceStatus, InvoiceType};
 use service::dashboard::invoice_count::{
     CountTimeRange, InvoiceCountError, InvoiceCountService, InvoiceCountServiceTrait,
 };
+use util::timezone::offset_to_timezone;
 
-use crate::ContextExt;
+use crate::{standard_graphql_error::StandardGraphqlError, ContextExt};
 
 fn do_invoice_count(
     ctx: &Context<'_>,
     invoice_type: &InvoiceType,
     invoice_status: &InvoiceStatus,
     range: &CountTimeRange,
-    timezone_offset: &Option<i32>,
+    timezone_offset: &FixedOffset,
 ) -> Result<i64> {
     let service_provider = ctx.service_provider();
-    let service_ctx = service_provider.context().map_err(|_| Error {
-        message: "InternalError".to_string(),
-        source: None,
-        extensions: None,
-    })?;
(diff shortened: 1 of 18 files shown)


graphql/src/schema/queries/invoice_counts.rs
```diff
@@ -31,16 +28,10 @@ fn do_invoice_count(
...
-                extensions: None,
-            },
-            InvoiceCountError::BadTimezoneOffset => Error {
-                message: "BadUserInput".to_string(),
-                source: None,
-                extensions: None,
-            },
+            InvoiceCountError::RepositoryError(err) => StandardGraphqlError::from(err),
+            InvoiceCountError::BadTimezoneOffset => {
+                StandardGraphqlError::BadUserInput("Invalid timezone offset".to_string())
```
andreievg, line 33: Hmm, i thought we were going with debug output of service error ? (since we wanted to simplify mapping between service error and api errors)

andreievg: Also it would mean we can avoid indirection of 'from'

graphql/src/schema/queries/stock_counts.rs
```diff
@@ -0,0 +1,51 @@
...
+    timezone_offset: FixedOffset,
+    days_till_expired: Option<i32>,
+}
+
+#[Object]
+impl StockCounts {
+    async fn expired(&self, ctx: &Context<'_>) -> Result<i64> {
+        let service_provider = ctx.service_provider();
+        let service_ctx = service_provider.context()?;
+        let service = StockExpiryCount {};
```
andreievg, line 18: Shouldn't this be in service provider ?

graphql/src/schema/queries/invoice_counts.rs
```diff
@@ -125,24 +116,27 @@ impl InboundInvoiceCounts {
...
-            timezone_offset: self.timezone_offset,
+            timezone_offset: self.timezone_offset.clone(),
         }
     }
 }
 
-pub fn invoice_counts(timezone_offset: Option<i32>) -> InvoiceCounts {
-    InvoiceCounts { timezone_offset }
+pub fn invoice_counts(timezone_offset: Option<i32>) -> Result<InvoiceCounts> {
+    let timezone_offset = offset_to_timezone(&timezone_offset).ok_or(
```
andreievg, line 138: this feels like service related mapping/check

graphql/src/schema/queries/invoice_counts.rs
clemens-msupply: I see I think I got you slightly wrong. You mean squashing all errors into a single standard error, right? thought we report at least some error code... Guess we can merge internal and bad user input but I think we need a special code for not_authenticated to give the frontend a change to re-login. However not authenticated (and not permitted) only needs to be handled once when validating the request so this wouldn't cause any extra work.

graphql/src/schema/queries/stock_counts.rs
clemens-msupply: Done in 7534440 (also for the invoice count service)

graphql/src/schema/queries/invoice_counts.rs
andreievg: I just meant to avoid string literals, as per code snippet here: https://github.com/openmsupply/remote-server/pull/646#discussion_r764809031

andreievg (approved):

graphql/src/schema/queries/invoice_counts.rs
clemens-msupply: So just format!("{}", err)? can do, but just wanted to provided a nicer error msg

andreievg: sure, up to you i guess, i like the simple format since it requires least amount of mapping.

PR #659, opened 2021-12-15
623 stock take service

clemens-msupply:
The stock take finalization should be seen as a draft, depending on the ongoing discussion... Had the feeling the changes are getting quite large so I do an intermediate PR.


graphql/src/schema/mutations/inbound_shipment/line/insert.rs
```diff
@@ -44,7 +44,15 @@ pub fn get_insert_inbound_shipment_line_response(
     input: InsertInboundShipmentLineInput,
 ) -> InsertInboundShipmentLineResponse {
     use InsertInboundShipmentLineResponse::*;
-    match insert_inbound_shipment_line(connection_manager, input.into()) {
+    let connection = match connection_manager.connection() {
+        Ok(con) => con,
+        Err(err) => {
+            return InsertInboundShipmentLineResponse::Error(ErrorWrapper {
```
andreievg, line 50: i think existing inbound shipment service won't work (since we can't pass on stock lines). 
Similar with outbound shipment (it would be different type).

I think we either create a new service for InventoryAdjustment or just do it within the stocktake service

repository/migrations/postgres/00000119_stock_take_table/up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE stock_take (
+    id TEXT NOT NULL PRIMARY KEY,
+    store_id TEXT NOT NULL REFERENCES store(id),
+    comment	TEXT,
+    description TEXT,
+    status TEXT,
```
andreievg, line 6: i think enums are important, see what pps think on omSupply chat

repository/migrations/postgres/00000119_stock_take_table/up.sql
```diff
@@ -0,0 +1,11 @@
...
+    created_datetime TIMESTAMP NOT NULL,
+    finalised_datetime TIMESTAMP,
+    inventory_additions_id TEXT REFERENCES invoice(id),
```
andreievg, line 9: I think we want to consolidate this into one (at some stage)

repository/migrations/sqlite/00000119_stock_take_table/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP VIEW IF EXISTS stock_take CASCADE;
```
andreievg, line 1: drop view ? (might be an issue somewhere else too, haven't really paid much attention to the down.sqls

repository/src/db_diesel/stock_take.rs
```diff
@@ -0,0 +1,167 @@
...
+            .limit(pagination.limit as i64)
+            .load::<StockTakeJoin>(&self.connection.connection)?;
+
+        Ok(result
+            .into_iter()
+            .map(|(stock_take, store)| StockTake { stock_take, store })
+            .collect())
+    }
+
+    pub fn find_one_by_id(&self, row_id: &str) -> Result<StockTakeJoin, RepositoryError> {
```
andreievg, line 161: personally i prefer class like this to return one type of parameter (this is why i think there should be two repositories, on for row and one for possible 'domain' type). You would need upsert method too (for sync and at least for testing), which i've been putting in row repository

repository/src/db_diesel/stock_take.rs
```diff
@@ -0,0 +1,167 @@
...
+
+type StockTakeJoin = (StockTakeRow, StoreRow);
+
+pub type StockTakeSort = Sort<StockTakeSortField>;
+
+type BoxedStockTakeQuery = IntoBoxed<'static, InnerJoin<stock_take::table, store::table>, DBType>;
+
+pub fn create_filtered_query<'a>(filter: Option<StockTakeFilter>) -> BoxedStockTakeQuery {
+    let mut query = stock_take_dsl::stock_take
+        .inner_join(store_dsl::store)
```
andreievg, line 86: i don't think we need store join, as per conversations we had, there is no need to support multi store operations (supservisor mode)

repository/src/db_diesel/stock_take.rs
andreievg: there is find_one_by_id on the stocktake_row repo, any need for this ?

repository/src/db_diesel/stock_take_row.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    pub fn find_one_by_id(&self, id: &str) -> Result<Option<StockTakeRow>, RepositoryError> {
+        let result = stock_take_dsl::stock_take
+            .filter(stock_take_dsl::id.eq(id))
+            .first(&self.connection.connection)
+            .optional();
+        result.map_err(|err| RepositoryError::from(err))
+    }
+
+    pub fn find_many_by_id(&self, ids: &[String]) -> Result<Vec<StockTakeRow>, RepositoryError> {
```
andreievg, line 50: is this one used ?

repository/src/schema/diesel_schema.rs
```diff
@@ -248,6 +277,8 @@ joinable!(master_list_name_join -> master_list (master_list_id));
 joinable!(master_list_name_join -> name (name_id));
 joinable!(item_is_visible -> item(id));
 joinable!(location -> store(store_id));
+joinable!(stock_take -> store(store_id));
+joinable!(stock_take_line -> location (location_id));
```
andreievg, line 281: joinable for stock_line too i think

service/src/stock_take/delete.rs
```diff
@@ -0,0 +1,53 @@
+use repository::{RepositoryError, StockTakeRowRepository, StorageConnection, TransactionError};
+
+use crate::{service_provider::ServiceContext, validate::check_store_id_matches};
+
+use super::validate::check_stock_take_exist;
+
+#[derive(Debug, PartialEq)]
+pub enum DeleteStockTakeError {
+    DatabaseError(RepositoryError),
+    StockTakeDoesNotExist,
```
andreievg, line 10: stock take is not empty (i.e. has lines)

service/src/stock_take/insert.rs
```diff
@@ -0,0 +1,96 @@
...
+    if !check_store_exists(connection, &stock_take.store_id)? {
+        return Err(InsertStockTakeError::InvalidStoreId);
+    }
+    Ok(())
+}
+
+fn generate(
+    InsertStockTakeInput {
+        id,
+        store_id,
```
andreievg, line 52: I really like what is being done with, store_id as graphql input

service/src/stock_take/insert.rs
```diff
@@ -0,0 +1,96 @@
...
+            let new_stock_take = generate(input);
+            StockTakeRowRepository::new(&connection).upsert_one(&new_stock_take)?;
+
+            let stock_take =
+                get_stock_take(ctx, new_stock_take.id).map_err(InsertStockTakeError::from)?;
+            stock_take.ok_or(InsertStockTakeError::InternalError(
+                "Failed to read the just inserted stock take!".to_string(),
+            ))
+        })
+        .map_err(|error: TransactionError<InsertStockTakeError>| error.to_inner_error())?;
```
andreievg, line 88: I think technically error types is not needed (if last error is typed to InsertStockTakeError)

service/src/stock_take/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+#[cfg(test)]
+mod tests;
+
+pub trait StockTakeServiceTrait: Sync + Send {
+    fn get_stock_takes(
+        &self,
+        ctx: &ServiceContext,
+        pagination: Option<PaginationOption>,
+        filter: Option<StockTakeFilter>,
+        sort: Option<StockTakeSort>,
```
andreievg, line 28: i feel like store_id should be here too (and we add filter by store_id), since when f/e request list of stock_takes it should filter by store_id, we can auto add that to filter in service)

service/src/stock_take/update.rs
```diff
@@ -0,0 +1,304 @@
...
+) -> Result<StockTakeInvoices, UpdateStockTakeError> {
+    // load stock take lines and matching stock lines
+    let stock_take_lines = StockTakeLineRepository::new(connection).query_by_filter(
+        StockTakeLineFilter::new().stock_take_id(EqualFilter::equal_to(stock_take_id)),
+    )?;
+    let stock_line_ids = stock_take_lines
+        .iter()
+        .map(|line| line.line.stock_line_id.to_owned())
+        .collect();
+    let stock = StockLineRepository::new(connection)
```
andreievg, line 101: maybe it's worth joining stock_line as result of StockTakeLineRepository ? (there is also a check that we would need to do in validation that snapshot quantity is current, or at the very least that we are not reducing below zero, so i think it would be useful)

service/src/stock_take/update.rs
```diff
@@ -0,0 +1,304 @@
...
+        let stock_line = stock
+            .iter()
+            .find(|stock| stock.id == stock_take_line.line.stock_line_id)
+            .ok_or(UpdateStockTakeError::InternalError(format!(
+                "Stock take line refers to invalid stock line: {:?}",
+                stock_take_line
+            )))?;
+        if stock_take_line.line.counted_number_of_packs > stock_line.total_number_of_packs {
+            // additions
+            added_inbound_lines.push(InsertInboundShipmentLine {
```
andreievg, line 119: as per comment further up, I think we shouldn't be reusing inbound and outbound shipment, it would be pretty easy to add InvoiceRow, InvoiceLineRow and StockLineRow manually here. And i think it would be easier as we are moving to just one invoice (rather then two)

andreievg (commented):
I've reviewed most of it, apart from stock_take/update, a few comments

graphql/src/schema/mutations/inbound_shipment/line/insert.rs
clemens-msupply: Yes this was just the initial attempt before the discussion

repository/migrations/sqlite/00000119_stock_take_table/down.sql
clemens-msupply: copy and past 8e55b31

repository/migrations/postgres/00000119_stock_take_table/up.sql
clemens-msupply: Done in 4a24bf4

repository/src/db_diesel/stock_take.rs
clemens-msupply: I don't think so, even though it might be useful for the validation checks where you don't need the join... I just added it because its in all the repos...

repository/src/db_diesel/stock_take.rs
clemens-msupply: Yes agree the store is not needed here 41ae383

repository/src/db_diesel/stock_take_row.rs
clemens-msupply: Same as above just included because other repos have it, should we remove these? wouldn't hurt much to keep them? maybe useful for test? don't really mind either way

repository/src/schema/diesel_schema.rs
clemens-msupply: 76eb614 Not sure if we would need it though but probably doesn't hurt to have it

service/src/stock_take/delete.rs
clemens-msupply: thanks done in b1964fe (I leave the test for another PR because I already have stock line mock data in a different branch) test is now also failing because of it will fix in next iteration

service/src/stock_take/insert.rs
clemens-msupply: Correct 4592350

service/src/stock_take/mod.rs
clemens-msupply: good point, should we have store_id in all service methods for consistency? (would pull it out of InsertStockTakeInput for the insert method)

service/src/stock_take/update.rs
clemens-msupply: Yes think that makes sense. Added in 3858323. Simplified the code quite a bit

clemens-msupply:
thanks for the review! will work on the stock take lines now, plan to get everything merged and then tackle stock take finalization afterwards

clemens-msupply:
thanks for the review! will work on the stock take lines now, plan to get everything merged and then tackle stock take finalization afterwards

service/src/stock_take/mod.rs
andreievg: could this be a good candidate for extra field in ServiceContext ?

andreievg (approved):
Thanks for changes. You've mentioned that some tests are failing, could potentially turn this into a features branch (or just branch from this branch), instead of merging to dev (if you want to fix tests in downstream branches)

service/src/stock_take/mod.rs
clemens-msupply: I think there are some request without a store_id like /me and /login...

PR #660, opened 2021-12-16
623 stock take line service

clemens-msupply:


repository/migrations/postgres/00000120_stock_take_line_table/up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE stock_take_line (
+    id TEXT NOT NULL PRIMARY KEY,
+    stock_take_id TEXT NOT NULL REFERENCES stock_take(id),
+    stock_line_id TEXT NOT NULL REFERENCES stock_line(id),
```
andreievg, line 4: as per convo, this would be optional (which also indicates that stock_line should be created, rather than updated), could potentially use the same shared method for inbound_shipment_line insert, it's currently using `generate_batch`.

service/src/stock_take_line/tests.rs
```diff
@@ -0,0 +1,342 @@
...
+                    batch: None,
+                    comment: None,
+                    cost_price_pack: None,
+                    sell_price_pack: None,
+                    snapshot_number_of_packs: None,
+                    counted_number_of_packs: None,
+                },
+            )
+            .unwrap();
+
```
andreievg, line 263: did you miss an assert ? (that stock_take_line_a is still the same in db as in mock) ?

service/src/stock_take_line/delete.rs
```diff
@@ -0,0 +1,62 @@
...
+    validate::check_store_id_matches,
+};
+
+use super::validate::check_stock_take_line_exist;
+
+#[derive(Debug, PartialEq)]
+pub enum DeleteStockTakeLineError {
+    DatabaseError(RepositoryError),
+    InternalError(String),
+    StockTakeLineDoesNotExist,
```
andreievg, line 14: StockTake is finalised

service/src/stock_take_line/delete.rs
```diff
@@ -0,0 +1,62 @@
...
+    service_provider::ServiceContext, stock_take::validate::check_stock_take_exist,
+    validate::check_store_id_matches,
+};
+
+use super::validate::check_stock_take_line_exist;
+
+#[derive(Debug, PartialEq)]
+pub enum DeleteStockTakeLineError {
+    DatabaseError(RepositoryError),
+    InternalError(String),
```
andreievg, line 13: I thought we would keep stock take enums as static (all enums, so you can see all possible errors, including internal errors just by looking at the error list), and do the generics in graqphl ?

service/src/stock_take_line/delete.rs
```diff
@@ -0,0 +1,62 @@
...
+    stock_take_line_id: &str,
+) -> Result<(), DeleteStockTakeLineError> {
+    let line = match check_stock_take_line_exist(connection, stock_take_line_id)? {
+        Some(line) => line,
+        None => return Err(DeleteStockTakeLineError::StockTakeLineDoesNotExist),
+    };
+    let stock_take = match check_stock_take_exist(connection, &line.stock_take_id)? {
+        Some(stock_take) => stock_take,
+        None => {
+            return Err(DeleteStockTakeLineError::InternalError(format!(
```
andreievg, line 30: Look a lot nice with `None => return Err(DeleteStockTakeLineError::StockTakeLinkedToStockLineIsMissing),`, mapping of internal error can be done in graphql (with debug stringify as suggested previously)

service/src/stock_take_line/insert.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub counted_number_of_packs: i32,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum InsertStockTakeLineError {
+    DatabaseError(RepositoryError),
+    InternalError(String),
+    InvalidStockTakeId,
+    InvalidStockLineId,
+    InvalidStoreId,
```
andreievg, line 35: stock take is finalised

service/src/stock_take_line/insert.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub sell_price_pack: f64,
+    pub snapshot_number_of_packs: i32,
+    pub counted_number_of_packs: i32,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum InsertStockTakeLineError {
+    DatabaseError(RepositoryError),
+    InternalError(String),
+    InvalidStockTakeId,
```
andreievg, line 33: StockTakeLineId already exists

service/src/stock_take_line/insert.rs
```diff
@@ -0,0 +1,119 @@
...
+        .connection
+        .transaction_sync(|connection| {
+            validate(connection, store_id, &input)?;
+            let new_stock_take_line = generate(input);
+            StockTakeLineRowRepository::new(&connection).upsert_one(&new_stock_take_line)?;
+
+            let line = get_stock_take_line(ctx, new_stock_take_line.id)
+                .map_err(InsertStockTakeLineError::from)?;
+            line.ok_or(InsertStockTakeLineError::InternalError(
+                "Failed to read the just inserted stock take line!".to_string(),
```
andreievg, line 108: As per earlier comments, I really dont' like hard coded strings, will keep on pushing for enums ; )

service/src/stock_take_line/update.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub counted_number_of_packs: Option<i32>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum UpdateStockTakeLineError {
+    DatabaseError(RepositoryError),
+    InternalError(String),
+    InvalidStockTakeLineId,
+    InvalidStoreId,
+    InvalidStockLineId,
```
andreievg, line 34: stock take is finalised

service/src/stock_take_line/update.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub sell_price_pack: Option<f64>,
+    pub snapshot_number_of_packs: Option<i32>,
+    pub counted_number_of_packs: Option<i32>,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum UpdateStockTakeLineError {
+    DatabaseError(RepositoryError),
+    InternalError(String),
+    InvalidStockTakeLineId,
```
andreievg, line 32: I think it's good to be more specific and have StockTakeLineIdDoesNotExists (or just RecordNotFound)

service/src/stock_take_line/insert.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub comment: Option<String>,
+    pub cost_price_pack: f64,
+    pub sell_price_pack: f64,
+    pub snapshot_number_of_packs: i32,
+    pub counted_number_of_packs: i32,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum InsertStockTakeLineError {
+    DatabaseError(RepositoryError),
```
andreievg, line 31: StockLine is already in stocktake

service/src/stock_take_line/update.rs
```diff
@@ -0,0 +1,130 @@
...
+) -> Result<StockTakeLine, UpdateStockTakeLineError> {
+    let result = ctx
+        .connection
+        .transaction_sync(|connection| {
+            let existing = validate(connection, store_id, &input)?;
+            let new_stock_take_line = generate(existing, input);
+            StockTakeLineRowRepository::new(&connection).upsert_one(&new_stock_take_line)?;
+
+            let line = get_stock_take_line(ctx, new_stock_take_line.id)
+                .map_err(UpdateStockTakeLineError::from)?;
```
andreievg, line 117: don't need to map_err here, question mark is sufficient

service/src/stock_take_line/tests.rs
```diff
@@ -0,0 +1,342 @@
...
+            .unwrap_err();
+        assert_eq!(error, DeleteStockTakeLineError::InvalidStoreId);
+
+        // success
+        let store_a = mock_store_a();
+        let existing_line = mock_stock_take_line_a();
+        let deleted_id = service
+            .delete_stock_take_line(&context, &store_a.id, &existing_line.id)
+            .unwrap();
+        assert_eq!(existing_line.id, deleted_id);
```
andreievg, line 340: do we need to assert that stock line was deleted ?

service/src/stock_take_line/insert.rs
clemens-msupply: thanks missed this. Just had a thought about it, why not have it upsert? Upsert would remove one error that frontend doesn't need to handle. For example, if frontend looses the connection during the insert request, frontend could just retry the insert even though backend already inserted the line from the previous call.

Disadvantage is that a buggy frontend could accidentally overwrite the wrong line but this seems to be easy to catch during testing.

service/src/stock_take_line/tests.rs
clemens-msupply: thanks done in 40205d4 (and revealed a bug 8-) )

service/src/stock_take_line/tests.rs
clemens-msupply: Thanks, yes that was the intention 40205d4

service/src/stock_take_line/delete.rs
clemens-msupply: thanks 751380d

service/src/stock_take_line/insert.rs
clemens-msupply: thats for insert but I also fixed it for stock_take 751380d

service/src/stock_take_line/update.rs
clemens-msupply: 751380d

service/src/stock_take_line/delete.rs
clemens-msupply: Sorry I missed that this was the plan. InternalError shouldn't really happen and I would keep it simple and just use the InternalError to hold a debug message. The InternalError also holds implementation dependent details and I wouldn't leak the implementation details into the upper layers.

service/src/stock_take_line/delete.rs
clemens-msupply: It misses the information that its an internal error, i.e. that it shouldn't happen at all.

service/src/stock_take_line/insert.rs
clemens-msupply: As commented above, a string literal is perfectly fine here ;)

service/src/stock_take_line/update.rs
clemens-msupply: thanks c9f4c3a

service/src/stock_take_line/insert.rs
clemens-msupply: fd45a72

service/src/stock_take_line/update.rs
clemens-msupply: Yes agree I improved the error names.  69ad40a

service/src/stock_take_line/insert.rs
andreievg: But we there could be a request to insert stock_take_line into finalised stock take

service/src/stock_take_line/insert.rs
andreievg: Do you mean have an upsert mutation ? If so, typically there a bunch of optionals in update mutation that are required in insert mutation, this makes error handling harder.

service/src/stock_take_line/delete.rs
andreievg: Oki, if it shouldn't happen at all do we need to add a message to it ?

service/src/stock_take_line/insert.rs
clemens-msupply: Yes sorry done 1fed463

service/src/stock_take_line/insert.rs
clemens-msupply: yes have an update and an insert operation but let the insert behave like an upsert so that it can't fail if applied multiple times.

Added the missing check earlier...

service/src/stock_take_line/delete.rs
clemens-msupply: Think its still valuable as a comment. And in case it happens anyway as a debug message.

In this case we could also avoid the error by having an inner join query but don't really want to add a new query just for this purpose (?)

repository/migrations/postgres/00000120_stock_take_line_table/up.sql
clemens-msupply: I think I start a new feature branch for this one...

andreievg (approved):

service/src/stock_take_line/delete.rs
andreievg: Sure, let me clarify:

In previous comment you mentioned: `The InternalError also holds implementation dependent details and I wouldn't leak the implementation details into the upper layers.` and here `Think it's still valuable as a comment`, do you mean that as an enum it would need to be destructured in error conversion to graphql (and thus leak to graphql layer) and as a string it wouldn't need to be processed by the destructuring and thus not leak ?

I still like the idea of using enums instead of strings (and just printing out it's debug output). But to retain the information of it being internal error and to not leak it would need to be another enum type.

andreievg (approved):
Approving after the latest round of comments, I still think enums vs strings, but don't think arguing about it is worth it.

service/src/stock_take_line/delete.rs
clemens-msupply: Yes for me the error is part of the service trait/interface. I think its ok to report that something bad happened but upper layers shouldn't really care about what exactly happen because this could change depending on the implementation, e.g. a different query could avoid the error.

Another reason is ease of dev: since the errors shouldn't really happen I don't want to spend time defining/updating an enum for "dead" code, I just want to do something better than unwrap().

Issue #661, opened 2021-12-20
Add invoice line type


andreievg:
Needed for https://github.com/openmsupply/remote-server/issues/531, a little bit de-normalised (can be deduced by combination of optional stock_line and item.type) but simplifies things quite a lot

Three type, StockIn, StockOut, Unallocated, Service

PR #662, opened 2021-12-20
661 add invoice line type

andreievg:
Fixes: #661 

Also updated test to make sure correct type is set on inserts, and also added invoice line type check in invoice query.
One other change was in outbund_shipment_line_insert -> generate method, reducing codes slightly

Added a commit to make sure that outbound shipment update generate is only adjusting StockIn or StockOut line (i think without this commit, even before this PR, if invoice has service line there would be an error when it's 'picked')


service/src/invoice_line/outbound_shipment_service_line/update/generate.rs
```diff
@@ -7,45 +7,30 @@ pub fn generate(
...
-        existing_line.item_name
-    };
+    if let Some(input_name) = input.name {
+        update_line.item_name = input_name
+    } else if let Some(input_item_id) = input.item_id {
+        if input_item_id != update_line.item_id {
+            update_line.item_name = item.name;
+            update_line.item_id = input_item_id;
+        }
+    }
```
clemens-msupply, line 21: The comment is a slightly out of sync now.

I would avoid having a mut here or mutating the input in general, e.g. you are comparing the input to the existing line and then you are mutating the existing line in the next step. This becomes hard to verify especially if you edit the code at a later point, e.g. are you still comparing to the existing line or has it already been updated earlier?

If its not for performance reason I think it was better the way it was before (or at least clone the whole input object).

I also like the explicit way it has earlier since you see exactly what you are generating with no details hidden.

clemens-msupply (approved):
LGTM just one comment

repository/src/schema/mod.rs
```diff
@@ -40,7 +40,7 @@ pub enum DatabaseRow {
 pub use central_sync_buffer::CentralSyncBufferRow;
 pub use central_sync_cursor::CentralSyncCursorRow;
 pub use invoice::{InvoiceRow, InvoiceRowStatus, InvoiceRowType};
-pub use invoice_line::InvoiceLineRow;
+pub use invoice_line::{InvoiceLineRow, InvoiceLineRowType};
```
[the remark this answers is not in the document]
andreievg: i think this was done in the original prototype and carried over (didn't have a chance to review original prototype).

With pub use it re-exports (so they are all available in through schema::, rather then say schema::invoice_line), i am keen to just do either `pub mod` or `pub use ...::*` (the later would mean less changes to existing imports)

service/src/invoice/outbound_shipment/update/generate.rs
```diff
@@ -80,7 +80,15 @@ pub fn generate_batches(
...
 ) -> Result<Vec<StockLineRow>, UpdateOutboundShipmentError> {
-    let invoice_lines = InvoiceLineRowRepository::new(connection).find_many_by_invoice_id(id)?;
+    let invoice_lines: Vec<InvoiceLineRow> = InvoiceLineRowRepository::new(connection)
+        .find_many_by_invoice_id(id)?
+        .into_iter()
+        .filter(|line| {
+            line.r#type == InvoiceLineRowType::StockIn
+                || line.r#type == InvoiceLineRowType::StockOut
+        })
+        .collect();
```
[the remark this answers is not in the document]
andreievg: True, I think i've used this check since `type` is not yet available as filter (it is in further PR), added todo: b8f5bc67059ae90d91d3cfe6f594c8d713a93bda

service/src/invoice_line/outbound_shipment_service_line/update/generate.rs
andreievg: The ownership of the record and input is transferred to generate, so it's not mutating input, it owns it now. 

> This becomes hard to verify especially if you edit the code at a later point, e.g. are you still comparing to the existing line or has it already been updated earlier?

This makes not difference if it's not mutating since the new record will be constructed from existing record (unless we query for the record again in generate)

To me it's more readable with the change, it wasn't straight away evident what `input.item_id.is_some() && input.item_id != Some(existing_line.item_id.to_owned())` means

clemens-msupply: Your are still mutating the input and reading it at the same time which is dangerous. Rust only prevents that somebody else does the same thing somewhere else.

Agree for the last point.

andreievg: 0266fdc4441e81af4c6fdd84384e8dd52bde18fb, also found a bug

PR #663, opened 2021-12-20
623 gql stock take line

clemens-msupply:
Only the delete stock take line endpoint since data struct for the other endpoints will change.

Additional changes:

- Move ValidationService into the service provider.
- Update PermissionsService to make it configurable for tests and use an Admin user as a default user for now.
- Disable auth on default for now
- Make store id optional in the endpoint

### DIFF
--- a/graphql/src/schema/mutations/stock_take/line/delete.rs
+++ b/graphql/src/schema/mutations/stock_take/line/delete.rs
@@ -1,6 +1,6 @@
 use async_graphql::*;
 use service::{
-    permission_validation::{Resource, ResourceAccessRequest, ValidationService},
+    permission_validation::{Resource, ResourceAccessRequest},
     stock_take_line::delete::DeleteStockTakeLineError,
 };
 
@@ -29,7 +29,8 @@ pub fn delete_stock_take_line(
     let service_provider = ctx.service_provider();
     let service_ctx = service_provider.context()?;
 
-    ValidationService::new(&service_ctx.connection).validate(
+    service_provider.validation_service.validate(
+        &service_ctx,
         ctx.get_auth_data(),
         &ctx.get_auth_token(),
         &ResourceAccessRequest {
--- a/graphql/src/schema/mutations/stock_take/line/delete.rs
+++ b/graphql/src/schema/mutations/stock_take/line/delete.rs
@@ -1,0 +1,57 @@
+use async_graphql::*;
+use service::{
+    permission_validation::{Resource, ResourceAccessRequest, ValidationService},
+    stock_take_line::delete::DeleteStockTakeLineError,
+};
+
+use crate::{standard_graphql_error::StandardGraphqlError, ContextExt};
+
+#[derive(InputObject)]
+pub struct DeleteStockTakeLineInput {
+    pub id: String,
+}
+
+#[derive(SimpleObject)]
+pub struct DeleteResponse {
+    pub id: String,
+}
+
+#[derive(Union)]
+pub enum DeleteStockTakeLineResponse {
+    Response(DeleteResponse),
+}
+
+pub fn delete_stock_take_line(
+    ctx: &Context<'_>,
+    store_id: &String,
+    input: &DeleteStockTakeLineInput,
+) -> Result<DeleteStockTakeLineResponse, StandardGraphqlError> {
+    let service_provider = ctx.service_provider();
+    let service_ctx = service_provider.context()?;
+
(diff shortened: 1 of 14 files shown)


graphql/src/schema/mutations/stock_take/line/delete.rs
```diff
@@ -0,0 +1,58 @@
...
+        Err(err) => Err(match err {
+            DeleteStockTakeLineError::DatabaseError(err) => err.into(),
+            DeleteStockTakeLineError::InternalError(err) => {
+                StandardGraphqlError::InternalError(err)
+            }
+            DeleteStockTakeLineError::StockTakeLineDoesNotExist(_) => {
+                StandardGraphqlError::BadUserInput(format!("{:?}", err))
+            }
+            DeleteStockTakeLineError::InvalidStoreId(_) => {
+                StandardGraphqlError::BadUserInput(format!("{:?}", err))
```
andreievg, line 54: isn't this part of authorisation error ?

graphql/src/standard_graphql_error.rs
```diff
@@ -37,3 +38,19 @@ impl From<RepositoryError> for StandardGraphqlError {
         StandardGraphqlError::InternalError(format!("{:?}", err))
     }
 }
+
+impl From<ValidationError> for StandardGraphqlError {
```
andreievg, line 42: I thought we were trying to avoid indirection ? (could be a case where it seems direct to the writer but not to the reader ?). Would prefer StandardGraphqlError::validation_error(..)

service/src/stock_take_line/delete.rs
```diff
@@ -22,7 +22,11 @@ fn validate(
 ) -> Result<(), DeleteStockTakeLineError> {
     let line = match check_stock_take_line_exist(connection, stock_take_line_id)? {
         Some(line) => line,
-        None => return Err(DeleteStockTakeLineError::StockTakeLineDoesNotExist),
+        None => {
+            return Err(DeleteStockTakeLineError::StockTakeLineDoesNotExist(
+                stock_take_line_id.to_string(),
```
andreievg, line 27: Isn't this known in the request ?

andreievg (approved):
I see that most changes are not related to stock take line (i've also done this before, where changes in the PR are out of scope of the issue being solved). Would be great if we can try and target specific issue with PR (creating new issue is pretty fast)

graphql/src/schema/mutations/stock_take/line/delete.rs
clemens-msupply: Yes right, it should already be caught there but I guess the service doesn't know about this fact and has to double check it again.

graphql/src/standard_graphql_error.rs
clemens-msupply: Its for the `validate()?` method that is called in every endpoint. Similar to RepositoryError I would say From is ok here (?).

Alternative would be:
`validate().map_err(::to_standard_gql_error)?`

service/src/stock_take_line/delete.rs
clemens-msupply: I actually added it for batch requests so that these errors become a bit more meaningful when there where multiple deletes. Should I remove?

andreievg: Ahh i see the motivation now, this is for batched operations that result in StandardGraphqlError, perhaps we should just log the whole input at this stage (for example in the following error just stock_take_id is logged, would be good to know which line it refers to)

graphql/src/schema/mutations/stock_take/line/delete.rs
andreievg: I think it's good to have this error in service, since it checks stock_take id against store_id. What i mean is, is it bad user input error or authorisation/permission type error ?

graphql/src/standard_graphql_error.rs
andreievg: i like the alternative, but as per our convo, if it's frequently used, i guess it's ok as from

andreievg (approved):

graphql/src/schema/mutations/stock_take/line/delete.rs
clemens-msupply: Yes agree should already be caught earlier in auth (even though not 100% sure if there other cases, e.g. user has access to a store but line belongs to different store) Think its doesn't hurt to leave it in.

Issue #664, opened 2021-12-20
Add unallocated line insert service


andreievg:
https://github.com/openmsupply/remote-server/issues/531

PR #668, opened 2021-12-20
664 add unallocated line insert service

andreievg:
Closes: #664 

Pretty straight forward, apart from last commit changes: https://github.com/openmsupply/remote-server/commit/c7c7cbd16fe6cbc604f646c35f276cc29a469478, i needed to add 'type' filter on invoice line and item, didnt' want to use domain `type` since filtering macros wouldn't work, so decided to relocate those filters to respective repos. Also changed ItemType (in diesel schema struct derives) to ItemRowType, since all other statuses and types that refer to rows are labeled ItemRowType (i think anything that is directly database data type related should have row in their name)


repository/migrations/sqlite/00000106_create_invoice_table/up.sql
```diff
@@ -8,7 +8,7 @@ CREATE TABLE invoice (
     store_id text NOT NULL REFERENCES store (id),
     invoice_number integer NOT NULL,
     type TEXT CHECK (type IN ('OUTBOUND_SHIPMENT', 'INBOUND_SHIPMENT')) NOT NULL,
-    status text CHECK (status IN ('PICKED', 'SHIPPED', 'NEW', 'DELIVERED', 'VERIFIED')) NOT NULL,
+    status text CHECK (status IN ('NEW','ALLOCATED', 'PICKED', 'SHIPPED',  'DELIVERED', 'VERIFIED')) NOT NULL,
```
clemens-msupply, line 11: When status is ALLOCATED and you add a new unallocated line does the status go back to NEW?
EDIT answered my question from the implementation. So ALLOCATED means "finalized/done/not editable", right?

repository/src/mock/mod.rs
```diff
@@ -66,6 +68,25 @@ pub struct MockData {
     pub full_master_list: HashMap<String, FullMockMasterList>,
     pub numbers: Vec<NumberRow>,
 }
+
+impl MockData {
+    pub fn new() -> Self {
```
clemens-msupply, line 73: The existing code uses Default::default() to create empty MockData.

repository/src/mock/test_unallocated_line.rs
```diff
@@ -0,0 +1,84 @@
...
+    result
+        .invoices
+        .push(mock_unallocated_line_allocated_invoice());
+    result
+        .invoice_lines
+        .push(mock_unallocated_line_new_invoice_line_1());
+    result
+}
+
+pub fn mock_unallocated_line_new_invoice() -> InvoiceRow {
```
clemens-msupply, line 21: The name is confusing because it returns an InvoiceRow and not a line

repository/src/mock/test_unallocated_line.rs
```diff
@@ -0,0 +1,84 @@
...
+        total_before_tax: 0.0,
+        total_after_tax: 0.0,
+        tax: None,
+        r#type: InvoiceLineRowType::UnallocatedStock,
+        number_of_packs: 1,
+        note: None,
+    }
+}
+
+pub fn mock_unallocated_line_allocated_invoice() -> InvoiceRow {
```
clemens-msupply, line 65: As above, better name? maybe just `mock_allocated_invoice`?

service/src/invoice/validate.rs
```diff
@@ -93,6 +93,20 @@ pub fn check_invoice_exists(
...
+pub fn check_invoice_exists_option(
+    id: &str,
+    connection: &StorageConnection,
+) -> Result<Option<InvoiceRow>, RepositoryError> {
+    let result = InvoiceRepository::new(connection).find_one_by_id(id);
+
+    match result {
+        Ok(invoice_row) => Ok(Some(invoice_row)),
+        Err(RepositoryError::NotFound) => Ok(None),
+        Err(error) => Err(error),
```
clemens-msupply, line 106: I found that there is an .optional(); method in diesel that does exactly this :) e.g.

```rust
let result: Result<Option<...>, ...> = stock_take_line_dsl::stock_take_line
            .filter(stock_take_line_dsl::id.eq(id))
            .first(&self.connection.connection)
            .optional();
```

service/src/invoice/validate.rs
```diff
@@ -93,6 +93,20 @@ pub fn check_invoice_exists(
     }
 }
 
+// TODO replace check_invoice_exists with this
+pub fn check_invoice_exists_option(
```
clemens-msupply, line 97: I always a bit confused if there is option in the fn name. Think it is clear enough if its just check_invoice_exists

service/src/invoice_line/outbound_shipment_unallocated_line/insert.rs
```diff
@@ -0,0 +1,357 @@
...
+    NotAnOutboundShipment,
+    //NotThisStoreInvoice,
+    CanOnlyAddLinesToNewOutboundShipment,
+    ItemNotFound,
+    NotAStockItem,
+    UnallocatedLineForItemAlreadyExistsInInvoice,
+    NewlyCreatedLineDoesNotExist,
+}
+
+type OutError = InsertOutboundShipmentUnallocatedLineError;
```
clemens-msupply, line 42: Hm maybe our error names get too long if we need to alias them... Not sure if that is better `InsertOutShipmentUnallocLineError`

service/src/invoice_line/outbound_shipment_unallocated_line/insert.rs
```diff
@@ -0,0 +1,357 @@
...
+    };
+
+    Ok(new_line)
+}
+
+pub fn check_unallocated_line_does_not_exist(
+    connection: &StorageConnection,
+    item_id: &str,
+) -> Result<bool, RepositoryError> {
+    let result = InvoiceLineRepository::new(connection).query_by_filter(
```
clemens-msupply, line 137: Use count() here to avoid retrieving all items (quite like the repository count()/query() pattern :))

service/src/invoice_line/outbound_shipment_unallocated_line/insert.rs
```diff
@@ -0,0 +1,357 @@
...
+    if invoice_row.r#type != InvoiceRowType::OutboundShipment {
+        return Err(OutError::NotAnOutboundShipment);
+    }
+
+    if invoice_row.status != InvoiceRowStatus::New {
+        return Err(OutError::CanOnlyAddLinesToNewOutboundShipment);
+    }
+
+    if !check_unallocated_line_does_not_exist(connection, &input.item_id)? {
+        return Err(OutError::UnallocatedLineForItemAlreadyExistsInInvoice);
```
clemens-msupply, line 92: I think this check also has to check that the unallocated line belong to the current invoice, or do I miss something?

service/src/invoice_line/validate.rs
```diff
@@ -19,6 +19,20 @@ pub fn check_line_does_not_exists(
     }
 }
 
+// TODO use this one instead of check_line_does_not_exists
+pub fn check_line_does_not_exists_new(
+    connection: &StorageConnection,
+    id: &str,
+) -> Result<bool, RepositoryError> {
+    let result = InvoiceLineRowRepository::new(connection).find_one_by_id(id);
```
clemens-msupply, line 27: Why the _new in the function name? (because the name is already used, right? think I got it now from the other todo)

use find_one_by_id(id).optional()

clemens-msupply (approved):
Github says this one is empty: `service/src/invoice_line/outbound_shipment_unallocated_line/validate.rs` should it be removed?

andreievg:
> Github says this one is empty: service/src/invoice_line/outbound_shipment_unallocated_line/validate.rs should it be removed?

852e9264e1c0850fed7a4245c2bbc64b34d1df19

repository/migrations/sqlite/00000106_create_invoice_table/up.sql
andreievg: https://github.com/openmsupply/remote-server/blob/852e9264e1c0850fed7a4245c2bbc64b34d1df19/graphql/src/schema/types/invoice_query.rs#L96-L100

I wonder if that's enough info ?

> When status is ALLOCATED and you add a new unallocated line does the status go back to NEW

No actually there is a check that will not allow unallocated lines to be added if invoice is allocated

repository/src/mock/mod.rs
andreievg: Thanks, i missed that. a3a41aea826e81e8faf7944b9077ddf9b576a91c

repository/src/mock/test_unallocated_line.rs
andreievg: yep, thanks: 7f24b6d5ae6fdfe7a31748aae507b6a18ca3e856

repository/src/mock/test_unallocated_line.rs
andreievg: 7f24b6d5ae6fdfe7a31748aae507b6a18ca3e856

service/src/invoice/validate.rs
andreievg: Nice, so we can refactor existing find_one_by_id methods sometime, I thought this work was outside of the scope of the issue

service/src/invoice_line/outbound_shipment_unallocated_line/insert.rs
andreievg: hmm, i think it's fine to alias, it makes further code more readable (even if the name is short). 
I don't like abbreviations btw, it's much clearer to use full names. 

We could just call them 'Error' and reference them in other methods by:
`outbound_shipment_unalloacted_line::Error`

Saying that, i think `InsertOutboundShipmentUnallocatedLineError` is fine

service/src/invoice_line/outbound_shipment_unallocated_line/insert.rs
andreievg: done, thanks: 384675d32037ff478d84cf0506ad2a6abad510ba

service/src/invoice_line/outbound_shipment_unallocated_line/insert.rs
andreievg: Thanks, nice catch ! I missed it in this PR, did it here: https://github.com/openmsupply/remote-server/pull/675

Issue #666, opened 2021-12-20
Add unallocated line delete service


andreievg:
https://github.com/openmsupply/remote-server/issues/531

PR #670, opened 2021-12-20
665 666 669 unallocated line delete/update service and outbound shipment status update restriction

### LINKED ISSUE #665: Add unallocated line update service
https://github.com/openmsupply/remote-server/issues/531

### LINKED ISSUE #669: Outbound status change to allocated only allowed when there are no unallocated lines

andreievg:
Closes #665 
Coses #666 
Closes #669

Should be pretty straight forward, as per the issues

### DIFF
--- a/service/src/invoice_line/outbound_shipment_unallocated_line/update.rs
+++ b/service/src/invoice_line/outbound_shipment_unallocated_line/update.rs
@@ -1,0 +1,166 @@
+use domain::invoice_line::InvoiceLine;
+use repository::{
+    schema::{InvoiceLineRow, InvoiceLineRowType},
+    InvoiceLineRowRepository, RepositoryError, StorageConnection,
+};
+
+use crate::{
+    invoice_line::{get_invoice_line_ctx, validate::check_line_exists_option},
+    service_provider::ServiceContext,
+    u32_to_i32,
+};
+
+pub struct UpdateOutboundShipmentUnallocatedLine {
+    pub id: String,
+    pub quantity: u32,
+}
+
+#[derive(Debug, PartialEq)]
+
+pub enum UpdateOutboundShipmentUnallocatedLineError {
+    LineDoesNotExist,
+    DatabaseError(RepositoryError),
+    LineIsNotUnallocatedLine,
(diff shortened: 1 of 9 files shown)


service/src/invoice_line/validate.rs
```diff
@@ -92,6 +92,19 @@ pub fn check_line_exists(
         Err(error) => Err(WithDBError::db(error)),
     }
 }
+// TODO use this one instead of check_line_exists
+pub fn check_line_exists_option(
+    connection: &StorageConnection,
+    id: &str,
+) -> Result<Option<InvoiceLineRow>, RepositoryError> {
+    let result = InvoiceLineRowRepository::new(connection).find_one_by_id(id);
```
clemens-msupply, line 100: .find_one_by_id(id).optional()

service/src/invoice_line/outbound_shipment_unallocated_line/update.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+fn generate(
+    UpdateOutboundShipmentUnallocatedLine {
+        id: _,
+        quantity,
+    }: UpdateOutboundShipmentUnallocatedLine,
+    mut line: InvoiceLineRow,
+) -> Result<InvoiceLineRow, UpdateOutboundShipmentUnallocatedLineError> {
+    line.number_of_packs = u32_to_i32(quantity);
```
clemens-msupply, line 70: As in the other PR, in general I would  keep the input immutable and clone it instead (even though this method is very simple)

clemens-msupply (approved):

service/src/invoice_line/outbound_shipment_unallocated_line/update.rs
andreievg: There is no conditions here, i am not sure why it needs to be immutable, it's not a parameter that is leaked out of this method.

service/src/invoice_line/validate.rs
andreievg: as per our convo, i think optional would only work when operations are done on dsl

PR #672, opened 2021-12-21
Update stock take line schema and services

clemens-msupply:


clemens-msupply:
will add more stuff here..

PR #673, opened 2021-12-21
Stores endpoint

clemens-msupply:
Add a simple stores endpoint. No auth right now since other changes need to be merged first... 

Proposed future work:
- Add default store to /me endpoint


graphql/src/standard_graphql_error.rs
```diff
@@ -37,3 +38,13 @@ impl From<RepositoryError> for StandardGraphqlError {
         StandardGraphqlError::InternalError(format!("{:?}", err))
     }
 }
+
+impl From<ListError> for StandardGraphqlError {
```
andreievg, line 42: As per earlier convo, kind of like the alternative ::to_graphql_error(), but nothing that stops approval

andreievg (approved):
Nice!

I think this shows that we are moving in the right direction:
* Less file changes
* Changes are clear and grouped by context

Would potentially keep StoreNode in queries/types (since it might be used by other end points, like 'me' and loader on 'name', to know that name is a store, thus stock is a transfer in outbound shipment, which is something we show in existing mSupply).

> // TODO add auth validation and restrict returned stores according to the user's permissions

I think it's potentially possible for user to see stores that they cannot log in to, like example above for names that are visible and that are linked to a store

PR #676, opened 2021-12-23
623 stock take line schema update

clemens-msupply:
There are still some tests missing that I haven't finished yet but I added a todo and plan to finish it after the break.

I also don't handle the case (error?) where the available count goes below zero...


repository/migrations/postgres/00000120_stock_take_line_table/up.sql
```diff
@@ -1,12 +1,16 @@
...
+    stock_line_id TEXT REFERENCES stock_line(id),
     location_id TEXT REFERENCES location(id),
-    batch TEXT,
     comment	TEXT,
-    cost_price_pack DOUBLE PRECISION NOT NULL,
-    sell_price_pack DOUBLE PRECISION NOT NULL,
     snapshot_number_of_packs INTEGER NOT NULL,
-    counted_number_of_packs INTEGER NOT NULL
+    counted_number_of_packs INTEGER,
+    item_id TEXT REFERENCES item(id),
```
andreievg, line 9: this should be not null ?

service/src/invoice/validate.rs
```diff
@@ -43,6 +43,14 @@ pub fn check_invoice_is_editable(invoice: &InvoiceRow) -> Result<(), InvoiceIsNo
             InvoiceStatus::Picked => false,
             InvoiceStatus::Verified => false,
         },
+        InvoiceRowType::InventoryAdjustment => match status {
```
andreievg, line 46: would just return false for now, without checking status (so far b/e doesn't expose inventory adjustment type invoices)

service/src/stock_take/update.rs
```diff
@@ -240,41 +380,27 @@ pub fn update_stock_take(
...
+            for line in result.shipment_lines {
+                shipment_line_repo.upsert_one(&line)?;
             }
-            // update stock take
-            StockTakeRowRepository::new(&connection).upsert_one(&stock_take_row)?;
+            StockTakeRowRepository::new(connection).upsert_one(&result.stock_take)?;
 
-            let stock_take = get_stock_take(ctx, stock_take_row.id)?;
+            // return the updated stock take
+            let stock_take = get_stock_take(ctx, stock_take_id)?;
```
andreievg, line 403: was wondering, in your opinion, what would be a limit of file size before extracting validate and generate to it's own files ? I personally think it's a good idea to keep file size below around 200 lines (less intimidating to someone new looking at it, and can expose composing 'bits' of functionality through file explorer, rather then parsing the whole file to see them)

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
+
+    Ok(StockTakeGenerateJob {
+        stock_take,
+        shipment: Some(shipment),
+        shipment_lines,
+        stock_lines,
+    })
 }
 
 pub fn update_stock_take(
```
andreievg, line 375: I like having this at the top, as it's the entry point of the whole `update_stock_take` service, devs looking at update functionality shouldn't need to parse/scroll the whole file to get an overview

service/src/stock_take/update.rs
```diff
@@ -240,41 +380,27 @@ pub fn update_stock_take(
...
+            // write data to the DB
+            let stock_line_repo = StockLineRowRepository::new(connection);
+            for stock_line in result.stock_lines {
+                stock_line_repo.upsert_one(&stock_line)?;
+            }
+            if let Some(shipment) = result.shipment {
+                let shipment_repo = InvoiceRepository::new(connection);
+                shipment_repo.upsert_one(&shipment)?;
+            }
+            let shipment_line_repo = InvoiceLineRowRepository::new(connection);
```
andreievg, line 396: I think it's better to name it inventory_adjustment vs shippment here.

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
     if !check_store_id_matches(store_id, &existing.store_id) {
         return Err(UpdateStockTakeError::InvalidStore);
     }
+    let stock_take_lines = StockTakeLineRepository::new(connection).query_by_filter(
+        StockTakeLineFilter::new().stock_take_id(EqualFilter::equal_to(&input.id)),
+    )?;
+
+    if let Some(uncounted) = check_stock_take_lines_counted(&stock_take_lines) {
+        return Err(UpdateStockTakeError::LinesNotCounted(uncounted));
```
andreievg, line 95: i think this check would only happen when stock take is finalised ?

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
+    }
 
-    Ok(existing)
-}
+    if let Some(mismatches) = check_snapshot_matches_current_count(&stock_take_lines) {
+        return Err(UpdateStockTakeError::SnapshotCountCurrentCountMismatch(
```
andreievg, line 99: i think this check would only happen when stock take is finalised ?

andreievg: I think we should check this if stocktake is finalised ?

Would assume something like this:
```rust
if input.status == StockTakeRowStatus::Finalised {
// using abstraction here allows for the higher level method be more literal (keeping to the same level of abstraction)
let stock_take_lines = get_lines_for_stocktake(&input.id, connection)?;

if stock_take_lines.len() == 0 {
     return Errr(NoLines);
}
    if let Some(mismatches) = check_snapshot_matches_current_count(&stock_take_lines) {
        return Err(UpdateStockTakeError::SnapshotCountCurrentCountMismatch(
            mismatches,
        ));
    }
}
```

service/src/stock_take/update.rs
andreievg, line 95: does this mean we can only update stock_take if all lines are counted ? I think we should be able to finalise a stock take with all lines uncounted if other things have changed

service/src/stock_take/update.rs
```diff
@@ -40,19 +27,56 @@ pub struct UpdateStockTakeInput {
...
+        .collect();
+    if uncounted.len() > 0 {
+        return Some(uncounted);
+    }
+    None
+}
+
+fn check_snapshot_matches_current_count(
+    stock_take_lines: &Vec<StockTakeLine>,
+) -> Option<Vec<String>> {
```
andreievg, line 58: i think this method should return StockTakeLine. In API we would want to return an error with array of stock take lines (connector), this makes it very easy for front end to action on the error. Returning just ids would mean we need to make another loader call in API or f/e would need to require the stock take with lines an filter out the ids

service/src/stock_take/update.rs
```diff
@@ -40,19 +27,56 @@ pub struct UpdateStockTakeInput {
...
+    for line in stock_take_lines {
+        let stock_line = match &line.stock_line {
+            Some(stock_line) => stock_line,
+            None => continue,
+        };
+        if line.line.snapshot_number_of_packs != stock_line.total_number_of_packs {
+            mismatches.push(line.line.id.clone());
+        }
+    }
+    if mismatches.len() > 0 {
```
andreievg, line 69: just saw there is `is_empty` method on vec

service/src/stock_take/update.rs
```diff
@@ -40,19 +27,56 @@ pub struct UpdateStockTakeInput {
...
+    /// Holds list of affected stock take line ids
+    LinesNotCounted(Vec<String>),
+    /// Holds list of affected stock take line ids
+    SnapshotCountCurrentCountMismatch(Vec<String>),
+}
+
+/// returns uncounted stock take lines
+fn check_stock_take_lines_counted(stock_take_lines: &Vec<StockTakeLine>) -> Option<Vec<String>> {
+    let uncounted: Vec<String> = stock_take_lines
+        .into_iter()
```
andreievg, line 46: hmm, i am not 100% sure what into_iter() would do on a reference, I was previously using it to take ownership during iteration, i think just iter() would be more cleaner, just an ovbservation.

service/src/stock_take/update.rs
```diff
@@ -40,19 +27,56 @@ pub struct UpdateStockTakeInput {
...
+        return Some(uncounted);
+    }
+    None
+}
+
+fn check_snapshot_matches_current_count(
+    stock_take_lines: &Vec<StockTakeLine>,
+) -> Option<Vec<String>> {
+    let mut mismatches = Vec::new();
+    for line in stock_take_lines {
```
andreievg, line 60: hmm, i've mentioned earlier about my avoidance to use map/reduce etc, in the context of having a mutable argument an a loop, I hope i am not contradicting myself here but i think in this example filter is better.
i.e. if we are return StockTakeLine

```rust
    let mismatches = stock_take_lines
        .iter()
        .filter(|line| match &line.stock_line {
            Some(stock_line) => {
                line.line.snapshot_number_of_packs != stock_line.total_number_of_packs
            }
            None => false,
        })
        .map(|line| line.clone())
        .collect::<Vec<StockTakeLine>>();

    if mismatches.is_empty() {
        None
    } else {
        Some(mismatches)
    }
```

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
-            finalised_datetime = Some(Utc::now().naive_utc());
+    stock_take_lines: &Vec<StockTakeLine>,
+    store_id: &str,
+) -> Result<StockTakeGenerateJob, UpdateStockTakeError> {
+    if input.status != Some(StockTakeStatus::Finalized) {
+        // just update the existing stock take
+        let stock_take = StockTakeRow {
+            id: existing.id,
+            store_id: existing.store_id,
+            comment: input.comment.or(existing.comment),
```
andreievg, line 272: As per our discussion earlier (still to be conclusive), I prefer to destructure input (so that when new input is added it won't be missed). And we are updating stock take, here it looks like we are generating new stock take (in fact you are doing this twice in this method). So would prefer if existing stock take that mutated and saved (this method now owns it, so it's safe to edit it).

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
+        if let Some(shipment_line) = shipment_line {
+            shipment_lines.push(shipment_line);
         }
     }
-    let stock_take_row = StockTakeRow {
-        id,
+
+    // find inventory adjustment name:
+    let name_result = NameQueryRepository::new(connection)
+        .query_by_filter(NameFilter::new().code(SimpleStringFilter::equal_to("invad")))?;
```
andreievg, line 328: "invad" should be a global constant string

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
+    // finalize the stock take
+    let mut shipment_lines: Vec<InvoiceLineRow> = Vec::new();
+    let mut stock_lines: Vec<StockLineRow> = Vec::new();
+    let shipment_id = uuid();
+    for stock_take_line in stock_take_lines {
+        let counted_number_of_packs = match stock_take_line.line.counted_number_of_packs {
+            Some(count) => count,
+            None => {
+                return Err(UpdateStockTakeError::InternalError(
+                    "Uncounted stock take line".to_string(),
```
andreievg, line 296: As per earlier comment, i think it should be possible to finalise a stock take with no counts, just updating other changes to stock_line

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
-    }: UpdateStockTakeInput,
+    input: UpdateStockTakeInput,
     existing: StockTakeRow,
-) -> Result<(StockTakeRow, Option<StockTakeInvoices>), UpdateStockTakeError> {
-    let mut finalised_datetime = None;
-    let mut invoices = None;
-    if let Some(ref status) = status {
-        if *status == StockTakeStatus::Finalized {
-            finalised_datetime = Some(Utc::now().naive_utc());
+    stock_take_lines: &Vec<StockTakeLine>,
```
andreievg, line 264: A the point of generating, StockTakeLines can be moved vs passing them are reference

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
-                color: None,
-            },
+    let quantiy_change = i32::abs(delta);
+    let shipment_line = if quantiy_change > 0 {
+        let line_type = if delta > 0 {
+            InvoiceLineRowType::StockIn
+        } else {
+            InvoiceLineRowType::StockOut
+        };
+        Some(InvoiceLineRow {
```
andreievg, line 161: I think it shouldn't be optional, if stock_line is update, invoice_line should record the update even if there is no quantity change (to preserve the history of update/mutation)

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
+            Some(count) => count,
+            None => {
+                return Err(UpdateStockTakeError::InternalError(
+                    "Uncounted stock take line".to_string(),
+                ))
+            }
+        };
 
-            invoices = Some(generate_invoices(connection, &id, &existing.store_id)?);
+        let (stock_line, shipment_line) = if let Some(ref stock_line) = stock_take_line.stock_line {
```
andreievg, line 301: If generate method now owns StockTakeLines, we don't need to `ref` here, or all the clones in helper methods.

``rust
stock_take_lines.into_iter().map(|line|  {
  if let Some(stock_line) = line.stock_line {
   generate_stock_line_update(store_id, &shipment_id, line.stock_take_line, stock_line)
  } else {
   generate_new_stock_line(store_id, &shipment_id, line.stock_take_line)
  }
 
}).collect::<Vec<Option<(StockLineRow, InvoiceLineRow>>>()
```

Also you can note there is no need for `connection` or error in generate helpers, item_id shouldn't be optional on stock_take_line.

// in update method
```rust
for update in result.stock_and_invoice_line.into_iter() {
 if let Some((stock_line, invoice_line) in result.stock_and_invoice_line {
   stock_line_repo.upsert(&stock_line)?;
   invoice_line_repo.upsert(&invoice_line)?;
 }
}
```

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
-                    as u32,
-                total_before_tax: 0.0,
-                total_after_tax: 0.0,
-                tax: None,
-            });
+    let item = match ItemRepository::new(connection).find_one_by_id(&stock_line.item_id)? {
+        Some(item) => item,
+        None => {
+            return Err(UpdateStockTakeError::InternalError(format!(
+                "Can't find item {} for new stock take line {}!",
```
andreievg, line 148: for existing

service/src/stock_take/update.rs
```diff
@@ -240,41 +380,27 @@ pub fn update_stock_take(
...
             }
-            // update stock take
-            StockTakeRowRepository::new(&connection).upsert_one(&stock_take_row)?;
+            StockTakeRowRepository::new(connection).upsert_one(&result.stock_take)?;
 
-            let stock_take = get_stock_take(ctx, stock_take_row.id)?;
+            // return the updated stock take
+            let stock_take = get_stock_take(ctx, stock_take_id)?;
             stock_take.ok_or(UpdateStockTakeError::InternalError(
                 "Failed to read the just inserted stock take!".to_string(),
```
andreievg, line 405: updated

service/src/stock_take/update.rs
andreievg: Also this could be a helper method, it's used twice and pretty much a replica

service/src/stock_take/update.rs
```diff
@@ -63,173 +87,289 @@ fn validate(
...
+    invoice_id: &str,
+    stock_take_line: &StockTakeLine,
+    counted_number_of_packs: i32,
+) -> Result<(StockLineRow, Option<InvoiceLineRow>), UpdateStockTakeError> {
+    let item_id = stock_take_line
+        .line
+        .item_id
+        .clone()
+        .ok_or(UpdateStockTakeError::InternalError(
+            "Stock take line without stock line and without item id".to_string(),
```
andreievg, line 200: i think item_id should not be optional on stock_take_line

service/src/stock_take_line/update.rs
```diff
@@ -11,19 +13,24 @@ use crate::{
...
     pub stock_line_id: Option<String>,
     pub location_id: Option<String>,
-    pub batch: Option<String>,
     pub comment: Option<String>,
-    pub cost_price_pack: Option<f64>,
-    pub sell_price_pack: Option<f64>,
     pub snapshot_number_of_packs: Option<i32>,
     pub counted_number_of_packs: Option<i32>,
+
+    pub item_id: Option<String>,
```
andreievg, line 27: i don't think we need item_id or stock_line_id on update, it should just be set once on insert

service/src/stock_take_line/insert.rs
```diff
@@ -12,20 +16,24 @@ use crate::{
...
     pub location_id: Option<String>,
-    pub batch: Option<String>,
     pub comment: Option<String>,
-    pub cost_price_pack: f64,
-    pub sell_price_pack: f64,
-    pub snapshot_number_of_packs: i32,
-    pub counted_number_of_packs: i32,
+    pub counted_number_of_packs: Option<i32>,
+
+    pub item_id: Option<String>,
```
andreievg, line 30: I think item_id should be required, we would need to check it though against stock_line_id (if one is passed through), but that should be pretty easy.

service/src/stock_take_line/insert.rs
```diff
@@ -38,6 +46,9 @@ pub enum InsertStockTakeLineError {
     StockLineDoesNotExist,
     LocationDoesNotExist,
     CannotEditFinalised,
+    /// Either stock take line xor item must be set
+    StockTakeLineXOrItem,
+    ItemDoesNotExist,
```
andreievg, line 51: StockLineAlreadyExistsInStocktake

service/src/stock_take_line/update.rs
```diff
@@ -83,27 +118,93 @@ fn generate(
...
+        None
+    };
+    let update_item_id = if existing.item_id != item_id {
+        item_id
+    } else {
+        None
+    };
+    match (update_stock_line_id, update_item_id) {
+        // normal update: (stock_line nor item_id have changed):
+        (None, None) => Ok(StockTakeLineRow {
```
andreievg, line 145: As per earlier comment, we should remove item_id and stock_line_id from update gql, which makes this a lot simpler

andreievg (commented):
Sorry, quite a few comments, I think we can simplify things quite a bit.

One this for us to get an agreement on is using existing record in updates rather then re-creating it (it's owned by the method, it can be changed by the method), technically speaking we are updating the record, not re-creating it. This together with destructuring input, would simplify code imo.

Would also suggest, for the future, to break this PR up into two (one for stock take and one for line)

repository/migrations/postgres/00000120_stock_take_line_table/up.sql
clemens-msupply: NULL means not counted yet
0 means zero items have been counted

Would that make sense?

service/src/invoice/validate.rs
clemens-msupply: thanks done in 7dbbcd5

service/src/stock_take/update.rs
clemens-msupply: I would go with "one bit of functionality" or "closely related functionality" per file rather than a somewhat arbitrary line count. This particular file it contains "update stock take". It hasn't got into my way yet that there are some more lines. Can split though if you think its too big (just easier for me to develop in one file). 

As you pointed out I missed to make some checks conditional dependent on the status. For this reason I could split the file into update.rs and update_finalized.rs, but probably not a good idea, just thinking out loud...

Even though I don't use it much there is also the file outline tab. I also just rediscovered file split view and wonder if I should use it more (never seriously used it before). 

Just had some random looks at other projects sources (pg, linux, golang, zig, fossil) and they all seem to have one file for one bit functionality and some of the files get quite big. Only golang seem to have a lot of small files with few lines and only some files are > 500lines.

service/src/stock_take/update.rs
clemens-msupply: think its more common to have the "main"/"entry" method (if there is any) at the bottom of the file (?) at least that where I usually look for entry methods...

service/src/stock_take/update.rs
clemens-msupply: I kept the repo name but adjusted the other variables to make it more clear 1812a98

service/src/stock_take/update.rs
clemens-msupply: I added NoLines check and put LinesNotCounted and SnapshotCountCurrentCountMismatch into the `if Some(StockTakeRowStatus::Finalised) ` block.

6a0fc32

clemens-msupply: What do you mean with the comment in the code block you posted?

service/src/stock_take/update.rs
clemens-msupply: Makes sense, I removed the check. fec0648

service/src/stock_take/update.rs
clemens-msupply: Makes sense but how is FE suppose to solve the conflict. Should it ask the user to recount?

I return the stock take line id and the stock line now.

348a6cf

service/src/stock_take/update.rs
clemens-msupply: They forgot `is_not_empty()` though. And !is_empty() is IMHO not better than len > 0.

service/src/stock_take/update.rs
clemens-msupply: already removed it but yes iter() should work as well

service/src/stock_take/update.rs
clemens-msupply: I now return the stock_take_line.id and the stock_line so filter and map would make that a bit more complicated...

In general I think at least map and filter are very common nowadays (at least when coming from js).

service/src/stock_take/update.rs
clemens-msupply: I destructured the input 0ac644e

We should really avoid mutating variables if possible. There are too many things that can go wrong (its a logic thing not a Rust memory safety issue). Writing it out twice is safer and IMHO easier to follow.

service/src/stock_take/update.rs
clemens-msupply: I introduced a `util/src/constants.rs` file 88c7e90.
Not sure how often we are going to use this const though so it might be simpler to keep it inlined.

service/src/stock_take/update.rs
clemens-msupply: done

service/src/stock_take/update.rs
clemens-msupply: 37114ce

service/src/stock_take/update.rs
clemens-msupply: clippy actual suggests this as well so seem to be a Rust thing, I changed it bebb08f

should we actually use clippy?

service/src/stock_take/update.rs
clemens-msupply: Lets discuss with Craig once he is back,  invoice/ shipment is a bit of a god table and can/should probably be cleaned up.

Had a look a the helper method but think it gets way to complex (for what it does) and its much simpler to just write it out twice.

service/src/stock_take/update.rs
clemens-msupply: I want to safely unwrap the Option and then I need two references...

service/src/stock_take/update.rs
clemens-msupply: thanks fixed it e10f5ff

service/src/stock_take/update.rs
clemens-msupply: e10f5ff

service/src/stock_take/update.rs
clemens-msupply: The item_id its only for creating a new stock_line. Otherwise the item id comes from the existing stock_line.

That there is either a item_id of a stock_line_id in the stock_take_line is validated when inserting/updating a stock_take_line.

service/src/stock_take_line/update.rs
clemens-msupply: That simplifies a couple of things :) 4c370e5

service/src/stock_take_line/insert.rs
clemens-msupply: Think its redundant information and can stay optional (see previous comment)

service/src/stock_take_line/update.rs
clemens-msupply: 4c370e5

service/src/stock_take_line/insert.rs
clemens-msupply: Thanks added in bbfd1b8

What about new stock line items that haven't been created yet?

clemens-msupply:
I added some missing tests e943f0e

repository/migrations/postgres/00000120_stock_take_line_table/up.sql
andreievg: sorry i meant item_id

service/src/stock_take/update.rs
andreievg: > I also just rediscovered file split view and wonder if I should use it more (never seriously used it before).

I find it really good.

> Just had some random looks at other projects sources

Hmm, are all of those written in C ? (the examples sounds like quite low level language project)  I was under the impression that higher level language projects tend to lean towards reduces file size)

It sounds like it's a bit of a preference, I guess I have a bit of a phobia of large files (probably because the hard to digest legacy code that I've previously worked with have large files, most likely a coincidence and not the cause of reduced readability)

service/src/stock_take/update.rs
andreievg: This ties slightly with the comment about file size, for me it makes a lot of sense to see what the file is about at the first glance, when opening it. Like in this example, scrolling to line 375 to see what 'update.rs' actually exposes doesn't seem ideal (I would assume that 90% of the time this is what the dev would like to see first when they open that file).

I guess if the file size is reduce to be like one or two screen sizes, and the functionality is still grouped but within the folder, it's should be easy to explore code base (without being too intimidated, especially for new devs)

service/src/stock_take/update.rs
andreievg: > What do you mean with the comment in the code block you posted?

Ohh it was just about trying not to use repo's directly in the `validate` method, my idea of validate method was that it's a higher level of abstraction, should read more like a `prose`, having a method that calls a repo and using it in validate would be more in line with that (even though it might seem redundant, as in the method would just be a one liner, but I think it's small things like that which make things more readable)

service/src/stock_take/update.rs
andreievg: >  Should it ask the user to recount?

Yeah no idea how they would solve them, but giving them the info would allow them all the options.

service/src/stock_take/update.rs
andreievg: ooh, never tried clippy, not sure tbh. I'll be using is_empty now though, seems more 'natural'

service/src/stock_take/update.rs
andreievg: > We should really avoid mutating variables if possible.
> Writing it out twice is safer and IMHO easier to follow.

I don't understand still. We are 'updating' a record in DB, not creating it. To me this is better represented as doing a mutation on it's instance (struct), to update the fields that are being changed (surely this is more readable then listing all the fields even if they are not changed, and also listing all the fields would result in more lines and more likely errors ?). 

Would it be better if mut variables is created from a .clone() ?

service/src/stock_take/update.rs
andreievg: We already have some types in sync that ideally would be constants (for system created names/stores, I think we are ignoring them now)

service/src/stock_take/update.rs
andreievg: ok, we just have to be mind full, when discussing, to be aware of translations that would need to be done if things are changed from how mSupply does ii

service/src/stock_take/update.rs
andreievg: > I want to safely unwrap the Option and then I need two references...

Hmm, I'll do an example.

service/src/stock_take_line/insert.rs
andreievg: Hmm, so in graphql, in order to get related item, you would need to:
* check stock_line is present, if so use item_id
* if stock_line is not present check item_id is some, then use item_id
* otherwise throw internal error
* finally use item_id to query for item, and throw internal error if one is not present

vs 
* set item_id from either stock_line or from passed through item_id in mutation

*.in query for item through stock_line, use non optional item_id
* use item_id to query for item, and throw internal error if one is not present

I don't think it's redundant, it helps to simplify logic

service/src/stock_take_line/insert.rs
andreievg: I think it's fine right ? Can have as many as you want new stock_lines for an item in the same stock take. Maybe we can do a check in the future when finalising that new stock line is unique (have to be combination of a lot of fields, since can have the same stock_line with the same properties in different locations or in with different expiry etc..)

andreievg (approved):
Thanks for the changes, looking good. 

We need a medium to help us agree on a few things
* functionality in a file to be visible at a glance or is it ok to have long files that require thorough look to see their function
* mutating variables that represent update operations in db
* how to best record metadata changes on stock_line

service/src/stock_take/update.rs
clemens-msupply: Yes c is probably is bit more verbose than Rust. For some reason golang always seem to be very short even though they have explicit error handling.

service/src/stock_take/update.rs
clemens-msupply: I did some more research on this but couldn't find some good guideline but from looking at some other code and from own experience I have the feeling that many people use it like this:

1) Interface/struct definitions first, e.g. GQL UpdateStockTakeInput, other input enums and returns types
2) The "main" method, e.g. update_stock_take 
3) other methods that are used, e.g. From, to_domain, map_err

So the main bits are still at the top and implementation details are at the bottom.

service/src/stock_take/update.rs
clemens-msupply: thanks I see what you mean, not 100% sure about it though since the repo one liner should almost as easy to read and gives all the details at the same time... Done in 00af44d

service/src/stock_take/update.rs
clemens-msupply: is like in functional programming where you never mutate and then you can make sure that what a once written structs will not be changed anywhere else.

Especially when there are two output structs and you start with a common struct it can become quite complicated to follow what has been changed.

service/src/stock_take/update.rs
clemens-msupply: will keep it in mind

service/src/stock_take_line/insert.rs
clemens-msupply: Yes makes sense will change in the gql endpoint branch so that I can update the endpoints as well (if needed)

PR #684, opened 2022-01-05
668 add unallocated line update and delete graphql

andreievg:
Closes #668 

Pretty much copy, paste and replace


graphql/src/schema/mutations/outbound_shipment/unallocated_line/delete.rs
```diff
@@ -0,0 +1,84 @@
...
+    standard_graphql_error::StandardGraphqlError,
+    ContextExt,
+};
+
+#[derive(InputObject)]
+pub struct DeleteOutboundShipmentUnallocatedLineInput {
+    pub id: String,
+}
+
+use DeleteOutboundShipmentUnallocatedLineInput as Input;
```
clemens-msupply, line 18: Still stumbling on what what Input, Error, ErrorInterface is when reading the code.

I also think the names are simply too long but not exactly sure how to fix it in a good way either, or if it's indicate that there is some underlying problem... Maybe have shorter rust struct names and use graphql macro to specify the gql name?

clemens-msupply (approved):
looks good, haven't done success gql tests but guess I should do this as well..

graphql/src/schema/mutations/outbound_shipment/unallocated_line/delete.rs
andreievg: Do you mean something like this: https://github.com/openmsupply/remote-server/compare/668-Add-unallocated-line-update-and-delete-graphql...668-simplified-types-example

If so i really like your suggestion

clemens-msupply: yes! could be many a little be more verbose, e.g DeleteInput, DeleteResponse, DeleteError. Since these are only used in the context of outbound shipment unallocated line it should be clear what they mean.

andreievg: https://github.com/openmsupply/remote-server/pull/684/commits/978a73230ab8ec65b8e19333a4d725e8b2b82d79

PR #687, opened 2022-01-05
677 fix inbound line update location

### LINKED ISSUE #667: Add unallocated line insert graphql
https://github.com/openmsupply/remote-server/issues/531

### LINKED ISSUE #677: Can't update locations for inbound lines
When I sent an update mutation for an inbound line, it has no effect. 

Inserting does insert with the location correctly.

Is it because the location ID is not assigned here? https://github.com/openmsupply/remote-server/blob/ce12f47445a6cd72458903d111a4b14f5aeb2b3d/service/src/invoice_line/inbound_shipment_line/update/generate.rs#L49

andreievg:
Closes #677

### DIFF
--- a/service/src/invoice_line/inbound_shipment_line/update/generate.rs
+++ b/service/src/invoice_line/inbound_shipment_line/update/generate.rs
@@ -50,11 +50,14 @@ fn generate_line(
     UpdateInboundShipmentLine {
         pack_size,
         batch,
-        expiry_date,
-        sell_price_per_pack,
         cost_price_per_pack,
+        sell_price_per_pack,
+        expiry_date,
         number_of_packs,
-        ..
+        location_id,
+        id: _,
+        invoice_id: _,
+        item_id: _,
     }: UpdateInboundShipmentLine,
     current_line: InvoiceLineRow,
     new_item_option: Option<ItemRow>,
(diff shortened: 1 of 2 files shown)


service/src/invoice_line/inbound_shipment_line/update/generate.rs
```diff
@@ -63,6 +66,7 @@ fn generate_line(
 
     update_line.pack_size = pack_size.map(u32_to_i32).unwrap_or(update_line.pack_size);
     update_line.batch = batch.or(update_line.batch);
+    update_line.location_id = location_id.or(update_line.location_id);
```
clemens-msupply, line 69: Thats why I like to write it out and not mutated it ;) However, when writing it out you can introduce other bugs so don't take this comment too serious, e.g.

```rust
{
  batch: location_id,
  location_id: batch,
}
```

clemens-msupply (approved):

service/src/invoice_line/inbound_shipment_line/update/generate.rs
andreievg: hmm, yeah it would have helped, likewise if input parameter was exhaustively destructured.

PR #688, opened 2022-01-06
623 gql stock take endpoints

clemens-msupply:
I made some to_domain method public (stock line and location). I think this is not optimal and is not needed if we use row db types as domain objects (StockTakeLine contains raw db types while StockLine and Location is a domain object).

Also added the missing extend() call for gql standard errors. To make this easier I added a validate_auth gql helper which also simplifies the code a little bit. 

Also contains "Make item_id in stock_take_line row NOT NULL" I thought it requires gql endpoint changes but it doesn't so its a bit out of place here, sorry about that.


graphql/src/schema/mutations/stock_take/delete.rs
```diff
@@ -0,0 +1,65 @@
...
+        })),
+        Err(err) => {
+            let formatted_error = format!("{:#?}", err);
+            let graphql_error = match err {
+                ServiceError::DatabaseError(err) => err.into(),
+                ServiceError::InvalidStore => StandardGraphqlError::BadUserInput(formatted_error),
+                ServiceError::StockTakeDoesNotExist => {
+                    StandardGraphqlError::BadUserInput(formatted_error)
+                }
+                ServiceError::StockTakeLinesExist => {
```
andreievg, line 55: If this would be a structured error, I think it helps front end to return the lines that exists, similar to `CannotDeleteInvoiceWithLines` types of errors

graphql/src/schema/mutations/stock_take/line/update.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+#[derive(SimpleObject)]
+pub struct UpdateStockTakeLineNode {
+    pub stock_take_line: StockTakeLineNode,
+}
+
+#[derive(Union)]
+pub enum UpdateStockTakeLineResponse {
+    Response(UpdateStockTakeLineNode),
```
andreievg, line 39: I don't think we need to define extra types, should just be Response(StockTakeLineNode) and Response(StockTakeNode), for inserts and update of stock take and line

graphql/src/schema/mutations/stock_take/line/insert.rs
```diff
@@ -0,0 +1,138 @@
...
+};
+
+#[derive(InputObject)]
+pub struct InsertStockTakeLineInput {
+    pub id: String,
+    pub stock_take_id: String,
+    pub stock_line_id: Option<String>,
+    pub location_id: Option<String>,
+    pub comment: Option<String>,
+    pub counted_number_of_packs: Option<i32>,
```
andreievg, line 23: we should use async graphql positive integer checks (change all i32 graphql input types to u32).

Also for pack_size we can use `NonZeroU32` https://docs.rs/async-graphql/latest/async_graphql/trait.InputType.html

graphql/src/schema/mutations/stock_take/line/insert.rs
```diff
@@ -0,0 +1,138 @@
...
+                InsertStockTakeLineError::StockLineAlreadyExistsInStockTake => {
+                    StandardGraphqlError::BadUserInput(formatted_error)
+                }
+                InsertStockTakeLineError::LocationDoesNotExist => {
+                    StandardGraphqlError::BadUserInput(formatted_error)
+                }
+                InsertStockTakeLineError::CannotEditFinalised => {
+                    StandardGraphqlError::BadUserInput(formatted_error)
+                }
+                InsertStockTakeLineError::StockTakeLineXOrItem => {
```
andreievg, line 94: Hmm if i got this error i would be quite confused (I think enum variant can be a bit longer)

graphql/src/schema/mutations/stock_take/mod.rs
```diff
@@ -1 +1,70 @@
...
+impl StockTakeNode {
+    pub async fn id(&self) -> &str {
+        &self.stock_take.id
+    }
+
+    pub async fn store_id(&self) -> &str {
+        &self.stock_take.store_id
+    }
+
+    pub async fn comment(&self) -> Option<String> {
```
andreievg, line 47: Can use `&Option<String>` i think, then don't need to clone

graphql/src/schema/mutations/stock_take/mod.rs
```diff
@@ -1 +1,70 @@
...
+    pub async fn description(&self) -> Option<String> {
+        self.stock_take.description.clone()
+    }
+
+    pub async fn status(&self) -> StockTakeNodeStatus {
+        from_domain(&self.stock_take.status)
+    }
+
+    pub async fn created_datetime(&self) -> NaiveDateTime {
+        self.stock_take.created_datetime
```
andreievg, line 60: how does this work ? (not reference and no clone)

graphql/src/schema/mutations/stock_take/update.rs
```diff
@@ -0,0 +1,126 @@
...
+#[Object]
+impl SnapshotCountCurrentCountMismatch {
+    pub async fn description(&self) -> &'static str {
+        "Snapshot count doesn't match the current stock count"
+    }
+
+    pub async fn lines(&self) -> Vec<StockTakeLineNode> {
+        self.0
+            .iter()
+            .map(|line| StockTakeLineNode { line: line.clone() })
```
andreievg, line 41: I wonder if we should be consistent with using Connector for all list outputs, see `CannotDeleteInvoiceWithLines`

graphql/src/schema/types/stock_take_line.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use crate::schema::types::{LocationNode, StockLineNode};
+
+pub struct StockTakeLineNode {
+    pub line: StockTakeLine,
+}
+
+#[Object]
+impl StockTakeLineNode {
+    pub async fn id(&self) -> &str {
```
andreievg, line 12: we need to expose most of the fields in StockLineNode, also should have ItemNode exposed, and maybe isNew (is_some stock_line_id)

graphql/src/schema/types/stock_take_line.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub async fn stock_take_id(&self) -> &str {
+        &self.line.line.stock_take_id
+    }
+
+    pub async fn stock_line(&self) -> Option<StockLineNode> {
+        self.line
+            .stock_line
+            .clone()
+            .map(|stock_line| StockLineNode {
+                stock_line: stock_line_to_domain((stock_line, self.line.location.clone())),
```
andreievg, line 25: self.line.location is not accurate here.

I think we need to use a loader here to get stock_line similar to `stock_line` on `InvoiceLineNode`

graphql/src/schema/types/stock_take_line.rs
```diff
@@ -0,0 +1,50 @@
...
+        self.line.location.clone().map(|location| LocationNode {
+            location: location_to_domain(location),
+        })
+    }
+
+    pub async fn comment(&self) -> Option<String> {
+        self.line.line.comment.clone()
+    }
+
+    pub async fn snapshot_number_of_packs(&self) -> i32 {
```
andreievg, line 43: u32 return

graphql/src/standard_graphql_error.rs
```diff
@@ -54,16 +71,16 @@ impl From<ValidationError> for StandardGraphqlError {
...
 
-impl From<ListError> for StandardGraphqlError {
-    fn from(err: ListError) -> Self {
-        match err {
-            ListError::DatabaseError(err) => err.into(),
-            ListError::LimitBelowMin(_) => StandardGraphqlError::BadUserInput(format!("{:?}", err)),
-            ListError::LimitAboveMax(_) => StandardGraphqlError::BadUserInput(format!("{:?}", err)),
-        }
-    }
+pub fn list_error_to_gql_err(err: ListError) -> AsyncGQLError {
```
andreievg, line 79: i wonder in this case if it should just be `async_graphql::Error` rather then aliasing to AsyncGQLError, I think I might have gone down the alias path previously too.

andreievg: Also, I think From derive was fine, where you've used list_error_to_gql_err, can just be StandardGraphqlError::From, i think attaching functionality to structs where possible is great. I was looking at some `to_domain` functions for inputs, and i thought similar, they should really be members of the struct

andreievg (commented):
Thanks, a few minor comments.

I skipped the tests, will look at them all together when stocktake graphql is finalised

graphql/src/schema/mutations/stock_take/line/update.rs
clemens-msupply: The reason I did it is to keep the door open to extend the api (and avoids introducing new api versions), e.g. return some extra information and not only the stock take line node.

andreievg: What would be extra information returned for insert or update ? Would we be changing all of the API routes (our api needs to be consistent) to use this new way ? Do we also add this on connector (i.e. nodes { row { InvoiceNode { id } someExtraInfo { id } } ?  

Would prefer if we kept the existing way, and only change from it if we foresee concrete examples.

clemens-msupply: Somehow thought we already do it like that for other endpoints, reverted in 495b844.

In general I would say its good practice to return an object at the endpoint root (not for every node object since the nodes can be reused in other endpoints as well and they are more like building blocks...)

I don't have concrete examples but the point for doing this is that you don't know what your future requirements are. For example, in the future we might need to return additional data about the update or add some debug information. Having an return object makes it very easy to add more data in the future without breaking the endpoint.

graphql/src/schema/mutations/stock_take/line/insert.rs
clemens-msupply: done in 7176897

graphql/src/schema/mutations/stock_take/line/insert.rs
clemens-msupply: There was also a type in the variant. I changed it to StockLineXOrItem and used a better error message for the graphql response (variant would become to long to do make it descriptive...)
c4cd0b3

graphql/src/schema/mutations/stock_take/mod.rs
clemens-msupply: NaiveDateTime implements Copy (was wondering the same when I did it)

graphql/src/schema/mutations/stock_take/mod.rs
clemens-msupply: 83d3fdd

graphql/src/schema/mutations/stock_take/update.rs
clemens-msupply: thanks done in e76fe80

graphql/src/schema/mutations/stock_take/line/update.rs
andreievg: > Having an return object makes it very easy to add more data in the future without breaking the endpoint.

I like this for mutations, I think we should have done it from the start, my comment was mainly about the consistency with existing mutations. Be keen to change sometime but for all existing ones too.

graphql/src/schema/mutations/stock_take/line/insert.rs
andreievg: I think enum variant `MissingStockLineOrItem` is not that long

graphql/src/schema/types/stock_take_line.rs
clemens-msupply: c4e42fa

graphql/src/schema/types/stock_take_line.rs
clemens-msupply: Fixed in c4e42fa. I had to add a LocationRow loader because the stock take line type uses the raw row types....

graphql/src/schema/types/stock_take_line.rs
clemens-msupply: done c4e42fa

graphql/src/standard_graphql_error.rs
clemens-msupply: I removed the type alias in cf1c2df. Rust doesn't allow me to impl From for ListError...

```
conflicting implementations of trait `std::convert::From<service::ListError>` for type `async_graphql::Error`
conflicting implementation in crate `async_graphql`:
- impl<T> From<T> for async_graphql::Error
  where T: 'static, T: std::fmt::Display, T: std::marker::Send, T: Sync;
upstream crates may add a new impl of trait `std::fmt::Display` for type `service::ListError` in future versions
```

Since the two types are in different crates I can't add to_domain to the service struct...

graphql/src/schema/mutations/stock_take/delete.rs
clemens-msupply: Would be helpful but client could also just reload the whole record... I know I am being lazy but its a lot of extra error handling since all the other error needs to be handled as well then... Will update in a new PR.

The real source of these errors is concurrent edits really. For this reason we may track concurrent edits instead. For example, every stock_take or line could have a version id and the client tells the server which version it wants to mutate. If the version has changed in the meantime the server returns a ConcurrentEdit error and the client can sync to the latest state. All the remaining errors stay bad input errors...

graphql/src/schema/mutations/stock_take/line/insert.rs
clemens-msupply: It also StockLine and Item can't be set at the same time...

PR #694, opened 2022-01-10
692 invoice by number

### LINKED ISSUE #692: XByInvoiceNumber Query
Have discussed previously but putting this here to not be forgotten.

A query for fetching a single shipment by invoice number, rather than ID.

andreievg:
Closes #692 

Ended up relocating existing get_invoice_respose to just get, and new query helper is just get_by_number


graphql/src/schema/mutations/inbound_shipment/insert.rs
```diff
@@ -44,7 +42,7 @@ pub fn get_insert_inbound_shipment_response(
         }
     };
     match insert_inbound_shipment(&connection, input.into()) {
-        Ok(id) => match get_invoice_response(connection_manager, id) {
+        Ok(id) => match invoice::get(connection_manager, id) {
```
clemens-msupply, line 45: maybe just get_invoice...

graphql/src/schema/queries/invoice.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+pub fn get(connection_manager: &StorageConnectionManager, id: String) -> InvoiceResponse {
+    match get_invoice(connection_manager, id) {
+        Ok(invoice) => InvoiceResponse::Response(invoice.into()),
+        Err(error) => InvoiceResponse::Error(error.into()),
+    }
+}
+
+pub fn get_by_number(
```
clemens-msupply, line 26: think get_invoice_by_number is clearer

service/src/invoice/query.rs
```diff
@@ -39,3 +42,72 @@ pub fn get_invoice(
         Err(SingleRecordError::NotFound(id))
     }
 }
+
+pub fn get_by_number(
```
clemens-msupply, line 46: as above why not use the a `query` method?

graphql/src/schema/queries/mod.rs
```diff
@@ -137,9 +138,18 @@ impl Queries {
...
+        invoice::get(connection_manager, id)
+    }
+
+    pub async fn invoice_by_number(
+        &self,
+        ctx: &Context<'_>,
+        invoice_number: u32,
+        r#type: InvoiceNodeType,
+    ) -> Result<invoice::InvoiceResponse> {
+        invoice::get_by_number(ctx, invoice_number, r#type)
```
clemens-msupply, line 152: Why not use the `invoices` endpoint and make the invoice number a filter parameter?

clemens-msupply (commented):
Is it on purpose that there is a new endpoint? and the number query is not added as a query filter parameter?

service/src/invoice/query.rs
andreievg: That's been available for awhile now, as per the issue request was for single invoice query by number

graphql/src/schema/mutations/inbound_shipment/insert.rs
clemens-msupply: get, insert and delete are very generic, think its better to give some more details what the method does. get_invoice also reads nicer than invoice::get. The methods also become harder to find when doing a global search.

When importing it you also need to use the module first to get help from rust analyser don't you? otherwise it would just import `get()`? Maybe too much convention to know to use it consistently...

graphql/src/schema/queries/invoice.rs
andreievg: a14b19fbe15464567c223f0594413cd3b51639d0

andreievg:
@clemens-msupply, i think i've made all the requested changes

clemens-msupply (approved):

PR #697, opened 2022-01-13
632 qlg stock take query and numbering

clemens-msupply:


graphql/src/schema/queries/stock_take_line.rs
```diff
@@ -0,0 +1,102 @@
...
+    )?;
+
+    let service_provider = ctx.service_provider();
+    let service_ctx = service_provider.context()?;
+    let service = &service_provider.stock_take_line_service;
+
+    match service.get_stock_take_lines(
+        &service_ctx,
+        store_id,
+        stock_take_id,
```
andreievg, line 66: this is a bit different to how lines for invoice are queried. I think lines should be a resolver on stock take (no need for seperate stock_line route)

graphql/src/schema/types/stock_take.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    pub async fn created_datetime(&self) -> &NaiveDateTime {
+        &self.stock_take.created_datetime
+    }
+
+    pub async fn finalised_datetime(&self) -> &Option<NaiveDateTime> {
+        &self.stock_take.finalised_datetime
+    }
+
+    pub async fn inventory_adjustment_id(&self) -> &Option<String> {
```
andreievg, line 52: Not required, but I think it's good to have InvoiceNode available in resolver

andreievg:
Sorry I missed this earlier, the `stock_line` in `StockTakeNode`, even thought stock_line is available in StockTakeLine, I think we should use the same loader as invoice_line. (Also I think the only place we should be constructing domain objects from their row parts should be is in the repository)

service/src/stock_take_line/query.rs
```diff
@@ -10,18 +11,41 @@ use crate::{
 pub const MAX_LIMIT: u32 = 1000;
 pub const MIN_LIMIT: u32 = 1;
 
+#[derive(Debug, PartialEq)]
+pub enum GetStockTakeLinesError {
+    DatabaseError(RepositoryError),
+    /// Stock take doesn't belong to the specified store
+    InvalidStore,
```
andreievg, line 18: Even thought I think stock take line query (without stock take) should be removed, I am wondering about this structured error `InvalidStore`, why would it be classified as structured vs standard ?

clemens-msupply: In this case a standard error because the user tried to access a stock take that doesn't belong to the store. Since the user specified the store id in the request its a bad user input.

graphql/src/schema/queries/stock_take_line.rs
clemens-msupply: Totally makes sense. Done in 4a4b38e

clemens-msupply:
> Sorry I missed this earlier, the `stock_line` in `StockTakeNode`, even thought stock_line is available in StockTakeLine, I think we should use the same loader as invoice_line. (Also I think the only place we should be constructing domain objects from their row parts should be is in the repository)

I use the StockLineByIdLoader in the stock take line node now.

graphql/src/schema/types/stock_take.rs
clemens-msupply: Done in 69517cf

andreievg (approved):
Thanks for changes

PR #698, opened 2022-01-14
Android java bindings

clemens-msupply:
- Adds a package for Android java bindings.
- Refactors a start_server method that can be used in main and from Android
- Add "off_switch" parameter to start_server to shutdown the server (e.g. when running on Android)
- Remove unwrap from sync (so that the error can be logged on Android)

### DIFF
--- a/android/src/android.rs
+++ b/android/src/android.rs
@@ -1,0 +1,67 @@
+#[allow(non_snake_case)]
+pub mod android {
+    extern crate jni;
+
+    use std::thread;
+
+    use jni::sys::jchar;
+    use repository::database_settings::DatabaseSettings;
+    use tokio::runtime::Runtime;
+
+    use server::settings::{AuthSettings, ServerSettings, Settings, SyncSettings};
+    use server::start_server;
+
+    use self::jni::objects::{JClass, JString};
+    use self::jni::sys::{jshort, jstring};
+    use self::jni::JNIEnv;
+
+    #[no_mangle]
+    pub unsafe extern "C" fn Java_org_openmsupply_mobile_RemoteServer_rustHelloWorld(
+        env: JNIEnv,
+        _: JClass,
+        name: JString,
+    ) -> jstring {
+        let name: String = env.get_string(name).unwrap().into();
+        let response = format!("Hello {}!", name);
+        env.new_string(response).unwrap().into_inner()
+    }
+
+    #[no_mangle]
+    pub unsafe extern "C" fn Java_org_openmsupply_mobile_RemoteServer_startServer(
+        _: JNIEnv,
+        _: JClass,
+        port: jchar,
+    ) -> jshort {
+        // run server in background thread
+        thread::spawn(move || {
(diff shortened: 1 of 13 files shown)


server/src/lib.rs
```diff
@@ -1,7 +1,127 @@
...
 pub mod environment;
 pub mod middleware;
 pub mod settings;
 pub mod sync;
 pub mod test_utils;
+
+/// Starts the server
+///
+/// This method doesn't return until a message is send to the off_switch.
+pub async fn start_server(
```
andreievg, line 37: is this copy paster straight from main.rs ? or any other changes here ?

server/src/sync/connection.rs
```diff
@@ -71,20 +71,20 @@ pub struct SyncConnection {
 }
 
 impl SyncConnection {
-    pub fn new(settings: &SyncSettings) -> SyncConnection {
+    pub fn new(settings: &SyncSettings) -> anyhow::Result<SyncConnection> {
```
andreievg, line 74: would this error only happen if Url parse fail ?

android/src/android.rs
```diff
@@ -0,0 +1,130 @@
...
+                    },
+                    database: DatabaseSettings {
+                        username: "n/a".to_string(),
+                        password: "n/a".to_string(),
+                        port: 0,
+                        host: "n/a".to_string(),
+                        database_name: db_path,
+                    },
+                    sync: SyncSettings {
+                        url: "http://localhost".to_string(),
```
andreievg, line 91: should this be a parameter ? Btw i think this sort of settings should be stored in DB, makes it easier to handle between platforms then toml file

android/src/android.rs
```diff
@@ -0,0 +1,130 @@
...
+                        username: "username".to_string(),
+                        password: "password".to_string(),
+                        interval: 300,
+                    },
+                    auth: AuthSettings {
+                        token_secret: "Make me configurable".to_string(),
```
andreievg, line 97: todo

andreievg (approved):
Looks good.

But since the makefile, android.rs and host.rs are quite technical, can we have in depth documentation about this please.? 

Liike why is there darwin host option for android build, what is server_bucket.

server/src/lib.rs
clemens-msupply: Mostly there is only an off_switch channel added to stop the server programmatically

server/src/sync/connection.rs
clemens-msupply: yes I was debugging and the panic don't show up in the Android logs so I needed the error to write it to the log...

android/src/android.rs
clemens-msupply: Done [d806b14](https://github.com/openmsupply/remote-server/pull/698/commits/d806b14d8affd1bd2af6b01edc8acb728c68608c)

clemens-msupply:
I added more comments to host.sh, makefile, and android.rs

PR #701, opened 2022-01-24
Create an example for spliting the graphql crate

clemens-msupply:
Spend too much time today waiting for cargo check so I did a little test having a second graphql crate and it seems to work much better, i.e. editing the smaller crate is much more pleasant :) 

For a proper endpoint we probably also need a `graphql_common` crate for common types, auth validation method etc...


andreievg:
Made a placeholder issue: https://github.com/openmsupply/remote-server/issues/781, will use this PR as reference but closing, doing a bit of a PR clean up

PR #711, opened 2022-01-27
710 requisition repositories

### LINKED ISSUE #710: Requisition repositories
https://github.com/openmsupply/remote-server/issues/703

andreievg:
Closes #710 

### New repos

* Requisition
* RequistionRow
* RequistionLine
* RequisitionLineRow

In tests I tried testing filter query results against actual sql statement results.

It seems like sort and filters are better places in repositories, this avoids extra types (domain filter, sort, requisition_type and status). What do you think ?


repository/src/db_diesel/requisition/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub r#type: Option<EqualFilter<RequisitionRowType>>,
+    pub status: Option<EqualFilter<RequisitionRowStatus>>,
+    pub created_datetime: Option<DatetimeFilter>,
+    pub sent_datetime: Option<DatetimeFilter>,
+    pub finalised_datetime: Option<DatetimeFilter>,
+    pub name_id: Option<EqualFilter<String>>,
+    pub name: Option<EqualFilter<String>>,
+    pub color: Option<EqualFilter<String>>,
+    pub their_reference: Option<EqualFilter<String>>,
+    pub comment: Option<EqualFilter<String>>,
```
clemens-msupply, line 24: SimpleStringFilter might be a better fit for comment

repository/src/tests.rs
```diff
@@ -896,42 +902,199 @@ mod repository_test {
...
-                .iter()
-                .enumerate()
-            {
-                assert_eq!(lines[count].id, line.id)
-            }
-        }
+            .unwrap();
+
+        let raw_result = sql_query(&format!(
+            r#"SELECT id from requisition where id = '{}'"#,
```
clemens-msupply, line 965: In general this is sqlite/postgres specific isn't it, i.e. test might fail depending on the feature.

The repository is the high level interface to the database. However, using raw sql makes the test dependent on the implementation. The test should ensure that the interface is not broken, e.g. when changing the database schema for this reason it shouldn't use SQL. Better is to fill the repo with some test data and test that you get the correct data back (the only thing you really care about).

clemens-msupply (approved):
You mean move the common sort and filter, e.g. EqualFilter, to repository as well? Yes please having all of these defined where they originate seem to be more natural to me (and easier to implement as well since you just need to put it in your repository file). Think some repos already do it.

If we ever need something different in the services layer we can easily define some special filter/sorts there.

Please see comment regarding raw SQL in repo tests.

repository/src/db_diesel/requisition/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+            id: None,
+            requisition_number: None,
+            r#type: None,
+            status: None,
+            created_datetime: None,
+            sent_datetime: None,
+            finalised_datetime: None,
+            name_id: None,
+            name: None,
+            color: None,
```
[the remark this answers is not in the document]
andreievg: thanks, 2a6aee7a9ac51e057164bc95e51579b54f8cf0bd

repository/src/db_diesel/requisition/mod.rs
andreievg: After talking to you and some thought, I agree, let's keep them seperate for now. 
also added sanity check (to make sure at least one record is matched).
11c39fc0be16056d59fe0b7cbd8ec5d0df41d9d2

repository/src/tests.rs
andreievg: I think if we are thinking that new repository implementation (api call etc..) are created then SQL test are not good, but if we are just testing the data access layer they seem pretty good to me (i think sql is pretty easy to read).

I'll try to remember to talk about it in our next meeting, hope it's ok to keep till then

Issue #722, opened 2022-02-03
Item stats service and query


andreievg:
https://github.com/openmsupply/remote-server/issues/703

PR #723, opened 2022-02-03
722 item stats service and query

andreievg:
closes https://github.com/openmsupply/remote-server/issues/722

For the loader, we need three things, item_id, store_id and look_back_datetime. Added ItemStatsLoaderInput struct with own hash implementation (only hashing by item_id), I don't know any other way of passing store_id and look_back_datetime to loader (don't think can institute new or edit state of existing loader due to it's asynchronous and 'sharable' (between resolvers in the same query) nature.

Ended up having two views, one for consumption and one for stock_info. Would have liked consumption view to have group_by and to pass through  look back date, but that would require a procedure, didn't want to use raw sql in diesel.
Summing up is done in the repo `to_domain`.


repository/src/db_diesel/item_stats.rs
```diff
@@ -0,0 +1,147 @@
...
+                    item_id: stock_info_row.item_id.clone(),
+                    stock_info_row,
+                    look_back_datetime,
+                },
+            )
+        })
+        .collect();
+
+    for consumption_row in consumption_rows.into_iter() {
+        map.entry(consumption_row.item_id.clone())
```
andreievg, line 99: technically entry will always exist, this could potentially could be an error (if entry doesn't exist) ? Or is default ok, or is it waiting to be invisible logic bug ?

repository/migrations/postgres/2021-12-25T10:00_item_stats/up.sql
```diff
@@ -0,0 +1,44 @@
+CREATE VIEW consumption AS
+SELECT 
+    'n/a' as id,
+    items_and_stores.item_id AS item_id, 
+    items_and_stores.store_id AS store_id,
+	COALESCE(consumption.quantity, 0) AS consumption_quantity,
+	consumption.date AS consumption_datetime
+FROM
+   (SELECT item.id AS item_id, store.id AS store_id FROM item, store) as items_and_stores
+LEFT OUTER JOIN
```
clemens-msupply, line 10: SQL question: why is `items_and_stores` needed here (same for the stock_info view)? wouldn't it work without the left outer join?

repository/src/db_diesel/item_stats.rs
clemens-msupply: would say its ok as it is, maybe just add a comment about it

repository/src/db_diesel/item_stats.rs
```diff
@@ -0,0 +1,147 @@
...
+        let number_of_days_since_look_back = now
+            .signed_duration_since(self.look_back_datetime)
+            .num_days();
+
+        let mut total_consumption = 0;
+        for row in &self.consumption_rows {
+            total_consumption += row.consumption_quantity
+        }
+
+        (total_consumption as f64 / number_of_days_since_look_back as f64 * 30 as f64) as i32
```
clemens-msupply, line 141: Probably good to have the 30 days as a float parameter, e.g. to make the test independent from the hard coded 30.

clemens-msupply (approved):

repository/src/db_diesel/item_stats.rs
andreievg: thanks: 9fd6fe8cbf4ece347206dff57b79b7692a7dba1d

repository/migrations/postgres/2021-12-25T10:00_item_stats/up.sql
andreievg: if there is no consumption or not batches the item/store record wouldn't show up. Technically it's not needed if the consumer knows that item/store will be skipped for records with no consumption and or stock data.

repository/src/db_diesel/item_stats.rs
andreievg: I've added it as a constants abae9e558dc17faaf77766f00293c84b12e092ca

andreievg: Is that what you meant ?

repository/migrations/postgres/2021-12-25T10:00_item_stats/up.sql
clemens-msupply: ok thanks that make sense now

repository/src/db_diesel/item_stats.rs
clemens-msupply: I thought about having it configurable as a parameter in average_monthly_consumption(days_in_month: Option<u8>). However, unlikely to be changed and having it as a constant it is easy to refactor now.

andreievg: I think it's good to have as a configurable thing, i have a question to ask in our v1 chat about amc calculation configurability

andreievg:

Issue #724, opened 2022-02-03
Request requisition line service


andreievg:
https://github.com/openmsupply/remote-server/issues/703

PR #725, opened 2022-02-03
724 Request requisition line service

andreievg:
closes https://github.com/openmsupply/remote-server/issues/724

Added insert, update, delete, pretty straight forward

### DIFF
--- a/service/src/requisition_line/request_requisition_line/insert.rs
+++ b/service/src/requisition_line/request_requisition_line/insert.rs
@@ -1,0 +1,287 @@
+use crate::{
+    requisition::{
+        common::check_requisition_exists, request_requisition::generate_requisition_lines,
+    },
+    requisition_line::{
+        common::{check_item_exists_in_requisition, check_requisition_line_exists},
+        query::get_requisition_line,
+    },
+    service_provider::ServiceContext,
+};
+
+use repository::{
+    schema::{RequisitionLineRow, RequisitionRow, RequisitionRowStatus, RequisitionRowType},
+    RepositoryError, RequisitionLine, RequisitionLineRowRepository, StorageConnection,
+};
+
+pub struct InsertRequestRequisitionLine {
+    pub id: String,
+    pub item_id: String,
+    pub requisition_id: String,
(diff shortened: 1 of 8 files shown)


service/src/requisition_line/request_requisition_line/insert.rs
```diff
@@ -0,0 +1,287 @@
...
+#[derive(Debug, PartialEq)]
+
+pub enum InsertRequestRequisitionLineError {
+    RequisitionLineAlreadyExists,
+    ItemAlreadyExistInRequisition,
+    RequistionDoesNotExist,
+    NotThisStoreRequisition,
+    CannotEditRequisition,
+    NotARequestRequisition,
+    ProblemCreatingRequisitionLine,
```
clemens-msupply, line 33: What problem? is it an internal error or an expected error? If expected maybe add a comment what the problem is. Otherwise I would just call it internal error.

Just thought about it, at service level a DatabaseError is really an InternalError as well. Other than the error message we don't want handle it anymore at a higher level. So it could be squashed into an InternalError(String) as well...

clemens-msupply (approved):

service/src/requisition_line/request_requisition_line/insert.rs
andreievg: Sorry, that error is bad, changed to (CannotFindItemStatusForRequisitionLine): eb4aec2a7f61bfa87c1892abccbca7270ed5ae74

Even though it's not really expected error (it's an internal error), in this case I still like to see a full list of all the possible errors via enum. (and it's no more work then adding InternalError(String))

clemens-msupply: It misses the information that it should never happen tough (at least add a comment in the enum). Higher level shouldn't really worry about it because its an implementation detail.

andreievg: Sure: 91ee15ab7fc70a43ca189575e5e3d732f8036b01 oops and 2649fa27e238bd75ba236a1546f964e904b4a840

andreievg:

PR #731, opened 2022-02-03
730 response requisition service to graphql mapping

### LINKED ISSUE #730: Response requisition service to graphql mapping
https://github.com/openmsupply/remote-server/issues/703

andreievg:
closes #730 

* Removed store_id from 
* Added some extra structured errors

Btw in the structured errors in requisition end points, I would like to see more info being returned, but wanted to implement as much functionality without those details (i.e. NothingRemainigToSupply can return shipments etc..)

### DIFF
--- a/graphql/src/schema/mutations/requisition/response_requisition/create_requisition_shipment.rs
+++ b/graphql/src/schema/mutations/requisition/response_requisition/create_requisition_shipment.rs
@@ -1,7 +1,18 @@
+use crate::{
+    schema::{
+        mutations::{requisition::errors::CannotEditRequisition, RecordDoesNotExist},
+        types::InvoiceNode,
+    },
+    standard_graphql_error::{validate_auth, StandardGraphqlError},
+    ContextExt,
+};
 use async_graphql::*;
 
-use crate::schema::{
-    mutations::requisition::errors::CannotEditRequisition, types::RequisitionNode,
+use service::{
+    permission_validation::{Resource, ResourceAccessRequest},
+    requisition::response_requisition::{
+        CreateRequisitionShipment as ServiceInput, CreateRequisitionShipmentError as ServiceError,
+    },
 };
 
 #[derive(InputObject)]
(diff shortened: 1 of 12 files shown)


graphql/src/schema/mutations/requisition/response_requisition/create_requisition_shipment.rs
```diff
@@ -10,26 +21,104 @@ pub struct CreateRequisitionShipmentInput {
 }
 
 #[derive(Interface)]
+#[graphql(name = "CreateRequisitionShipmentErrorInterface")]
 #[graphql(field(name = "description", type = "String"))]
-pub enum CreateRequisitionShipmentErrorInterface {
+pub enum DeleteErrorInterface {
+    RecordDoesNotExist(RecordDoesNotExist),
+    NothingRemainingToSupply(NothingRemainingToSupply),
     CannotEditRequisition(CannotEditRequisition),
```
clemens-msupply, line 29: are these errors all due to possible concurrent editing? maybe we should encode this information into the errors as well.

How would FE recover from these errors? if FE just pulls updated data from the server and tries to reapply changes the errors don't need to contain more information.

Btw. did you try to do an exhaustive error check in TypeScript? I had some type problem that prevented me to do so (I haven't fully investigated though)...

clemens-msupply (approved):

graphql/src/schema/mutations/requisition/response_requisition/create_requisition_shipment.rs
andreievg: > are these errors all due to possible concurrent editing? maybe we should encode this information into the errors as well.

yep i wanted to as per the pr comment, but ran out of time, i'll inform the front end crew tomorrow to let us know what other information can help them deal with standard/structured errors they encounter

> How would FE recover from these errors? if FE just pulls updated data from the server and tries to reapply changes the errors don't need to contain more information.

> Btw. did you try to do an exhaustive error check in TypeScript? I had some type problem that prevented me to do so (I haven't fully investigated though)...

Sure although good API would reduce number of calls (and give you back what you need to recover from the error at the time it happens)

https://www.dropbox.com/home?preview=zoom_0.mp4 around 1 hour and 5 mins

andreievg:

Issue #706, opened 2022-01-25
Transfers


andreievg:
https://github.com/openmsupply/remote-server/issues/703

### Scope

* Creating of Inbound from Outbound Shipment and Response
* Creating Response from Request Requisition
* Keeping all of the links for Order Cycle
* Site/Sync safe update of linked records

Diagram in https://github.com/openmsupply/remote-server/issues/703 describes most of how linked records are created and updated, and has a link diagram.

Here is a list of conditions/operations by event:

For all, only do if **name_id store belongs to current site**, and it's possible that store_id belongs to another site (record not editable, see below for further conditions)

1. **Request Requisition is SENT** and No Response Requisition linked to Request Requisition
  a. Create Response Requisition (linking to Request Requisition)
  b. Link Request Requisition to Response Requisition
2. **Response Requisition is FINALISED** and linked Request Requisition is not Finalised
  a. Finalise linked Request Requisition
3. **Outbound Shipment is PICKED** and No Inbound Shipments linked to Outbound Shipment
  a. Create Inbound Shipment (linking to Outbound Shipment)
  b. Link Inbound Shipment to Request Requisition, through Outbound Shipment linked Response Requisition and linked Request Requisition
  c. Link Outbound Shipment to Inbound Shipment
4. **Outbound Shipment is SHIPPED** and linked Inbound Shipments is Picked
  a. Set Inbound Shipment as Shipped
5. **Inbound Shipment is DELIVERED** and linked Outbound Shipments is Shipped/Picked
  a. Set Outbound Shipment as Delivered
6. **Inbound Shipment is VERIFIED** and linked Outbound Shipments is Shipped/Picked/Delivered
  a. Set Outbound Shipment as Verified

This is all easily done if site owns both of the stores, but if linked records belong to different sites, we need to have a mechanism to do 1.a, 2.b, etc..

I think a simple processing queue or just a method that takes record ids and table numbers for all records that might have triggers of 1-6 would work. Persisted processing queue is a bit more stable (in case of power failure or some unforeseen concurrent issue).

Then (I think they are all existing rules)

* Global sync rule: All requisitions and invoices are always sent to name_id -> store owner site.
* Global sync rule: Remote site cannot send or edit records not belonging to that site (can receive and integrate of course as per above)
* After integrating remote records, add all integrated requisitions and invoices to sync processing queue and trigger sync processing actor to run (not guaranteed that it will be processed, might not match criteria)
[...]

Issue #703, opened 2022-01-25
Requisitions and Transfers (Full Order Cycle)


andreievg:
Related: https://github.com/openmsupply/remote-server/issues/625, https://github.com/openmsupply/remote-server/issues/624

### Tasks

* [x] - https://github.com/openmsupply/remote-server/issues/703
* [x] -  https://github.com/openmsupply/remote-server/issues/708
* [x] - https://github.com/openmsupply/remote-server/issues/710
* [x] -  https://github.com/openmsupply/remote-server/issues/712
* [x] - https://github.com/openmsupply/remote-server/issues/716
* [x] - https://github.com/openmsupply/remote-server/issues/718
* [x] - https://github.com/openmsupply/remote-server/issues/720
* [x] - https://github.com/openmsupply/remote-server/issues/722
* [x] - https://github.com/openmsupply/remote-server/issues/722
* [x] - https://github.com/openmsupply/remote-server/issues/724
* [x] - https://github.com/openmsupply/remote-server/issues/726
* [x] - https://github.com/openmsupply/remote-server/issues/728
* [x] - https://github.com/openmsupply/remote-server/issues/730
* [x] - https://github.com/openmsupply/remote-server/issues/732
* [x] - https://github.com/openmsupply/remote-server/issues/737
* [ ] - https://github.com/openmsupply/remote-server/issues/739
* ~~[Transfers](https://github.com/openmsupply/remote-server/issues/706) (Sync Processing Queue)~~
   * RequestRequisition <-> ResponseRequisition
   * OutboundShipment <-> InboundShipment

This issue summarises area of work that allows for complete order cycle.

### Order Cycle
[drive](https://app.diagrams.net/#G1Iq1kR4ZoxKOGZXqHz6K9p7-pH6-h63j0)
![Order Cycle](https://user-images.githubusercontent.com/26194949/150956728-78155212-3f43-4058-b90c-085c853a4b17.png)

### Graphql Schema 

See issue and pr

#### Requisition

| mSupply                      | omSupply                  |                                                                         |
|------------------------------|---------------------------|-------------------------------------------------------------------------|
| date_entered                 |                           | if no om_created_datetime present use this to populate created_datetime |
| (new) om_created_datetime    | created_datetime          | also translates to date_entered in sync                                 |
| (new) om_sent_datetime       | sent_datetime             |                                                                         |
| (new) om_finalised_datetime  | finalised_datetime        |                                                                         |
[...]

PR #736, opened 2022-02-03
734 post sync processor service mechanism

### LINKED ISSUE #734: Post sync processor service mechanism
https://github.com/openmsupply/remote-server/issues/706

andreievg:
closes #734

As per the issue and underlying diagram from https://github.com/openmsupply/remote-server/issues/703, there a a bunch of operations to be done on some record change (duplicate into response or inbound, or update the other half)

Still missing a couple of operations for full sync to work (from https://github.com/openmsupply/remote-server/issues/706, point 7 and 8), in TODO (couldn't add as both stores are considered active atm)

Wanted to avoid a huge list of conditionals and a huge method which has been identified (for readability) as one of the issues in existing sync, thus created an array of ProcessRecord types for each operation.

Seperate CreateAndLink and just Create for sync, as per #706, since can't fully link if origin store in not active.

Since there is a seperate can_execute and process_record in the trait, have to do a little bit of extra extraction of some value in process_record (and do ConditionNotMetInProcessor error), not 100% happy with this one, potentially could be simpler with not using trait struct and just having a hard coded condition set in main mod.rs, i.e.:
```rust
if let Some(processor_result) = CreateAndLinkRequistionProcessor::can_execute(record) {
  CreateAndLinkRequistionProcessor::process_record(record, processor_result);
}
```

above would mean can_execute will provide anything that's need to process_record (i.e. already unwrapped option values).

You will notice that picked invoice is still editable, even though it appears in receiving store as not editable, but lines can be updated without receiving store realising until it's shipped (also when shipped have to update lines so we do full delete of lines and then insert, instead of more complex spot the difference logic)


service/src/sync_processor/invoice/common.rs
```diff
@@ -0,0 +1,228 @@
...
+        expiry_date,
+        note,
+        id: _,
+        stock_line_id: _,
+        invoice_id: _,
+        location_id: _,
+        location_name: _,
+        requisition_id: _,
+        cost_price_per_pack: _,
+    } in source_lines.into_iter()
```
clemens-msupply, line 185: In this case is there an advantage in destructuring InvoiceLine? if a field is added to it it would also be added in InvoiceLineRow wouldn't it?

service/src/sync_processor/invoice/create_and_link_invoice.rs
```diff
@@ -0,0 +1,55 @@
...
+    ProcessRecord, ProcessRecordError, ProcessRecordResult, Record, RecordForProcessing,
+};
+
+use super::common::{can_create_inbound_invoice, generate_and_integrate_linked_invoice};
+
+pub struct CreateAndLinkInvoiceProcessor {}
+
+impl ProcessRecord for CreateAndLinkInvoiceProcessor {
+    fn name(&self) -> String {
+        "Create linked invoice (linking source invoice)".to_string()
```
clemens-msupply, line 13: name() is more like a description, rename it?

service/src/sync_processor/invoice/create_and_link_invoice.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        false
+    }
+
+    fn process_record(
+        &self,
+        connection: &repository::StorageConnection,
+        record_for_processing: &RecordForProcessing,
+    ) -> Result<ProcessRecordResult, ProcessRecordError> {
+        if let Record::InvoiceRow(source) = &record_for_processing.record {
```
clemens-msupply, line 37: in general, to avoid one indent I would return early by doing:
```rust
let source = match &record_for_processing.record {
    Record::RequisitionRow(source) => source,
    Record::InvoiceRow(_) => return Ok(ProcessRecordResult::ConditionNotMetInProcessor),
};
```

service/src/sync_processor/invoice/common.rs
```diff
@@ -0,0 +1,228 @@
...
+            invoice_line_row_repository
+                .find_one_by_id(&line.id)
+                .unwrap(),
+        );
+    }
+
+    Ok(result_lines)
+}
+
+pub fn re_generate_linked_invoice_lines(
```
clemens-msupply, line 156: re_generate -> regenerate? or generate_updated_linked_invoice_lines? for a moment I was pondering what "re" means 8-)

service/src/sync_processor/invoice/common.rs
```diff
@@ -0,0 +1,228 @@
...
+    source_invoice: &InvoiceRow,
+    record_for_processing: &RecordForProcessing,
+) -> Result<(InvoiceRow, Vec<InvoiceLineRow>), ProcessRecordError> {
+    let invoice_row = generate_linked_invoice(connection, &source_invoice, record_for_processing)?;
+    let (lines_to_delete, invoice_line_rows) =
+        re_generate_linked_invoice_lines(connection, &invoice_row, &source_invoice)?;
+
+    InvoiceRepository::new(connection).upsert_one(&invoice_row)?;
+
+    let invoice_line_repository = InvoiceLineRowRepository::new(connection);
```
clemens-msupply, line 56: As mentioned in a previous PR we should go through the invoice services really to enforce the business logic in a single place...

service/src/sync_processor/invoice/update_inbound_shipment.rs
```diff
@@ -0,0 +1,85 @@
...
+        &self,
+        connection: &repository::StorageConnection,
+        record_for_processing: &RecordForProcessing,
+    ) -> Result<ProcessRecordResult, ProcessRecordError> {
+        if let (Record::InvoiceRow(source_invoice), Some(Record::InvoiceRow(linked_invoice))) = (
+            &record_for_processing.record,
+            &record_for_processing.linked_record,
+        ) {
+            let mut update_linked_invoice = linked_invoice.clone();
+            let (lines_to_delete, inserted_lines) = re_generate_linked_invoice_lines(
```
clemens-msupply, line 57: Since re_generate_linked_invoice_lines is only used here, move it here?

service/src/sync_processor/requisition/create_requisition.rs
```diff
@@ -0,0 +1,50 @@
...
+                    connection,
+                    &source,
+                    record_for_processing,
+                )?;
+
+            let result = ProcessRecordResult::Success(format!(
+                "generated: {:#?}\n{:#?}",
+                requisition_row, requisition_line_rows
+            ));
+            return Ok(result);
```
clemens-msupply, line 46: This processor doesn't do anything is it an duplicate of CreateAndLinkRequistionProcessor? in this case generate_and_integrate_linked_requisition can be moved to create_and_link_requisition.rs

clemens-msupply (approved):

service/src/sync_processor/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn name(&self) -> String;
+
+    fn can_execute(&self, record_for_processing: &RecordForProcessing) -> bool;
+
+    fn process_record(
+        &self,
+        connection: &StorageConnection,
+        record_for_processing: &RecordForProcessing,
+    ) -> Result<ProcessRecordResult, ProcessRecordError>;
+}
```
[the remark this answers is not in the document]
clemens-msupply: Alternatively, the more OOP way is to have a stateful ProcessRecord that is dynamically created by a ProcessRecordFactory through a method `fn can_execute(record) -> Option<Box<dyn ProcessRecord>>`. But think this is too much boilerplate...

service/src/sync_processor/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+                &record,
+            )?,
+            linked_record: get_linked_record(connection, &record)?,
+            other_party_store,
+            source_name,
+            record,
+        };
+
+        // iterators don't work for dyn box ? (have to annotate type somehow), thus just looping
+        let mut i = 0;
```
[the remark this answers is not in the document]
andreievg: i can't replicate now, but had a type issue, cheers for trying to reproduce

service/src/sync_processor/mod.rs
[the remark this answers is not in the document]
andreievg: I thought about this for awhile, in the context of indirection that comes with trait (interface). And i think this is a better way to do it: https://github.com/openmsupply/remote-server/compare/734-Post-sync-processor-service-mechanism...734-sync-processors-without-trait

What do you think ?

service/src/sync_processor/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+                    processor_name,
+                    record: record_for_processing.clone(),
+                });
+                break;
+            } else {
+                results.push(ProcessRecordResultSet {
+                    result: ProcessRecordResult::ConditionNotMet,
+                    processor_name,
+                    record: record_for_processing.clone(),
+                })
```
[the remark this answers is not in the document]
andreievg: Hmm, possibly, i think this sort of core processing should be logged in a file (actually I think we should be doing a lot of file logging, it's really hard to diagnose remote problems I think overly verbose logging to file would help quite a bit)

service/src/sync_processor/invoice/common.rs
andreievg: Not really, i prefer it this way, it seems very clean

service/src/sync_processor/invoice/create_and_link_invoice.rs
andreievg: it gone in the example i have of traitless sync processor above

service/src/sync_processor/invoice/create_and_link_invoice.rs
andreievg: Nice, i really like this, added to the traitless example: 35411a8eaca83d267649bfe360b19e128fc4ee6c

service/src/sync_processor/invoice/common.rs
andreievg: it deletes and remove lines, so generate doesn't quite explain it, i've remove `_` in the tratiless example: 6d2441281f65bee353110977ddf763fb165edbe4

service/src/sync_processor/invoice/common.rs
andreievg: I agree, i've made an issue: https://github.com/openmsupply/remote-server/issues/762, keen to discuss tomorrow

service/src/sync_processor/invoice/update_inbound_shipment.rs
andreievg: it's also use in common file

service/src/sync_processor/requisition/create_requisition.rs
andreievg: There is a difference between just create and create and link processors (and the helpers are shared between the two), the difference is, sometimes the source record is not editable on this site, thus we cannot update it (link it).

Potentially we could just have create and a seperate one to link. Thus on a site where both form and to are active, first run of sync processor would create the duplicated record (it would then be put on a change_log), and then sync processor would process the newly created one and link source from it. I prefer this way but i thought it might be a bit complex

service/src/sync_processor/mod.rs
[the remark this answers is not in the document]
clemens-msupply: I actually liked having everything belonging to one processor in one trait implementation. Think its a relatively common pattern as well but maybe I am biased because I like this abstraction.

In the new branch the ProcessRecordError and ProcessRecordResult have IMHO too much details. I don't think this information is needed, probably just a string for logging is enough? Furthermore, for every new processor you need to update these enums as well.

service/src/sync_processor/requisition/create_requisition.rs
clemens-msupply: thanks for clarifying, think I misread the code

service/src/sync_processor/mod.rs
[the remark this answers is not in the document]
clemens-msupply: Just to clarify, I like the "visitor" process pattern here because it really applies to many processors and there are probably more to come. Sometimes you only have a few use cases for an abstraction and the abstraction becomes a bit questionable but here is seems useful.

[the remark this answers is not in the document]
andreievg: My worry with the abstraction is that it becomes indirection in rust, it's hard to know what is happening for each iteration of the loop, but if processors are listed it makes it easier to digest. Of course it means we have to enforce the pattern via PR reviews, keen to chat about it tomorrow

service/src/sync_processor/requisition/create_requisition.rs
andreievg: i've added to agenda for tomorrow discussion to chat about circulars relations (which would exist with current thinking, requisition<->requisition, invoice<->invoice)

service/src/sync_processor/mod.rs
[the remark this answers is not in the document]
andreievg: > In the new branch the ProcessRecordError and ProcessRecordResult have IMHO too much details. I don't think this information is needed,

Hmm i think it's needed for logging (`MHO too much details` -> we should be logging more then just ids of records, this is super useful when debugging or diagnosing remote problems, i can go back to the database error example https://github.com/openmsupply/remote-server/pull/289#discussion_r699723193, where I was leaning toward more info of the error being stored, and it was really hard to know what database error happened until we added full database error info to repository error).

And I am not convinced that string errors are faster to add, it does require also editing enum and creating a response struct but at the same time it doesn't require writing string literals (just see what you have left at the end of processor, create a struct to hold it and add it to the struct as result)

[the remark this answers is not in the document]
andreievg: I've refactored again, using trait and string return, is that what you were suggesting ? 
I've been deliberately avoiding things that can cause indirection with rust, I think sync is a complex domain and we should not try to be tricky with it. I personally would do it the trait-less way (for sake of future devs coming on board to maintain the project, devs that might not be as familiar with rust), but if the target audience is me and you then using a trait is way better.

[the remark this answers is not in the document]
clemens-msupply: I think with the processor array inlined into the method it should be very easy to understand how it works now :) And the trait enforces how the processors are supposed to be implemented.

Only suggestions I have:
- The ProcessRecordError is dependent on the processors that implement the trait (dependent on the errors they throw). Would an anyhow error work as well? i.e. you still have the full error for logging but the trait is "clean".
-  try_process_record could return a struct instead of a string to leave it to the logger to decide how to compose this information:

```rust
  struct SyncProcessorResult {
     /// Processor name (needed?)
     name: String,
     description: String,
     /// details about what has been done
     message: String
  }
```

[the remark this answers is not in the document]
andreievg: if it's ok I'll do anyhow error in the https://github.com/openmsupply/remote-server/issues/762.

> try_process_record could return a struct instead of a string to leave it to the logger to decide how to compose this information:

hmm, string seems sufficient for now

PR #748, opened 2022-02-08
714 sync api and changelog

clemens-msupply:
Sorry for mixing to somewhat independent commit in one PR:

1) sync api: I added a sync v3 API implementation which will be needed for sync. While doing this I refactored the existing SyncConnection (which was the sync v5 api). Both sync APIs now follow the same pattern:
   - the API file (e.g. sync_api_v5.rs) only contains API related methods
   - the sync API is standalone and doesn't contain references to other server structs
   - the API file contains all structs related to the API
  
2) Changelog table and example trigger for the stock_take table.
  - changelog table
  - changelog view with duplicated rows
  - triggers


repository/migrations/postgres/2022-01-27T08:00_create_changelog_table/up.sql
```diff
@@ -0,0 +1,55 @@
...
+        t1.row_id,
+        t1.row_action
+    FROM changelog t1
+    WHERE t1.id = (SELECT max(t2.id) 
+                    from changelog t2
+                    where t2.row_id = t1.row_id);
+
+-- Helper trigger function for updating the changelog when a row has been mutated.
+-- This function should be used in table triggers.
+CREATE OR REPLACE FUNCTION update_changelog()
```
andreievg, line 36: nice

repository/migrations/postgres/2022-01-27T08:00_create_changelog_table/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE IF EXISTS changelog CASCADE;
+DROP VIEW IF EXISTS changelog_cleaned CASCADE;
```
andreievg, line 2: not sure if we ever use down.sql but anyways, i think it should be changelog_deduped

server/src/sync/synchroniser.rs
```diff
@@ -78,13 +94,29 @@ impl Synchroniser {
...
+                sync_batch
+                    .data
+                    .unwrap_or(vec![])
+                    .into_iter()
+                    .map(|record| {
+                        Ok(CentralSyncBufferRow {
+                            id: record.id,
+                            table_name: record.table_name,
+                            record_id: record.record_id,
+                            data: serde_json::to_string(&record.data)?,
```
andreievg, line 113: maybe map_err here is better (keeping to the place where it can originate). Also i kind of like the idea of single layer of abstraction in this loop and having a helper method for the translation.

let central_sync_records = to_central_sync_buffer.to_central_sync_buffer_row().map_err(...)

andreievg: I do like the new way of using serde though

repository/src/db_diesel/changelog_row.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// # Arguments:
+    /// * earliest the first cursor to be included in the returned
+    /// * count the number of entries to be returned
+    pub fn changelogs(
+        &self,
+        earliest: u64,
+        count: u32,
+    ) -> Result<Vec<ChangelogRow>, RepositoryError> {
+        let result = changelog_deduped_dsl::changelog_deduped
+            .filter(changelog_deduped_dsl::id.ge(earliest.try_into().unwrap_or(0)))
```
andreievg, line 27: there are a bunch of methods like `i32_to_u32`, maybe a good addition

repository/src/db_diesel/changelog_row.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl<'a> ChangelogRepository<'a> {
+    pub fn new(connection: &'a StorageConnection) -> Self {
+        ChangelogRepository { connection }
+    }
+
+    /// # Arguments:
+    /// * earliest the first cursor to be included in the returned
+    /// * count the number of entries to be returned
+    pub fn changelogs(
```
andreievg, line 21: With this view, how does a subscriber know what ID it's up to ? I think it becomes difficult with this view. I'll give an example tomorrow, but just consider a subscriber needing to process X number of records at a time from the change log, and there a bunch of records with more then one operation scattered (operations on those records are not consecutive records in the table)

repository/src/tests.rs
```diff
@@ -902,6 +906,114 @@ mod repository_test {
...
+                    id: 3,
+                    table_name: ChangelogTableName::StockTake,
+                    row_id: stock_take_b.id.clone(),
+                    row_action: ChangelogAction::Upsert,
+                }
+            ]
+        );
+
+        // delete an entry
+        stock_take_repo.delete(&stock_take_b.id).unwrap();
```
andreievg, line 996: as per earlier comment, I think these tests should include a test of subscriber going through change queue X records at a time

server/src/sync/synchroniser.rs
```diff
@@ -119,25 +151,27 @@ impl Synchroniser {
...
         info!("Initialising remote sync records...");
-        let mut sync_batch: RemoteSyncBatch = self
-            .connection
-            .initialise_remote_records()
-            .await
-            .map_err(|source| RemoteSyncError {
-                msg: "Failed to initialise remote sync records",
-                source: anyhow::Error::from(source),
-            })?;
+        let mut sync_batch: RemoteSyncBatchV5 =
```
andreievg, line 159: fyi, we should only do post_initilise, once, when site is doing it's first sync

server/src/sync/synchroniser.rs
```diff
@@ -149,8 +183,8 @@ impl Synchroniser {
             if let Some(data) = sync_batch.data {
```
andreievg, line 183: Saw an artifact comment `// TODO: acknowledge after integration`, that's not quite right, we should acknowledge as we sync into buffer.

server/src/sync/sync_api_v5.rs
```diff
@@ -0,0 +1,445 @@
...
+            server_url,
+            credentials,
+            client,
+        }
+    }
+
+    // Initialize remote sync queue.
+    //
+    // Should only be called on initial sync or when re-initializing an existing data file.
+    pub async fn post_initialise(&self) -> Result<RemoteSyncBatchV5, SyncConnectionError> {
```
andreievg, line 81: I think it's good to keep remote and central in these methods, the concepts already are a little bit confusing, it's good to know what methods deals with what type.

andreievg (commented):
Would like to approve, but wondering about the changelog_deduped view and how it can be used by subscriber (i've made a couple of comment about that)

repository/migrations/postgres/2022-01-27T08:00_create_changelog_table/down.sql
clemens-msupply: thanks done in 0821e8b

server/src/sync/synchroniser.rs
clemens-msupply: Done in ffed519 

Think it was clearer before though, too much indirection and the loop was short enough before. Have a look what you think.

repository/src/db_diesel/changelog_row.rs
clemens-msupply: I know, they are in `service` though and I didn't want to do this refactor in this PR. We could move i32_to_u32, etc into `util`

repository/src/db_diesel/changelog_row.rs
clemens-msupply: Yes an example would be good. Isn't it up to the subscriber to manage the cursor and what state it is in?

Do we actually plan to "vacuum" the changelog? In this case we also need a mechanism for the subscriber to what point they have handled the changelog...

server/src/sync/sync_api_v5.rs
clemens-msupply: I named all method after the actual endpoints, i.e. SyncApiV5 is just a thin wrapper around the underlying api. For this reason I would keep the naming to make it easy for somebody who knows the api to find the required method...

server/src/sync/synchroniser.rs
andreievg: i think it's better.

btw in my view `indirection` is lack of straightforwardness (like when compiler figures out from conversions and it's not clear to reader what is happening), here it's abstraction, and it's very clear what `central_sync_batch_records_to_buffer_row` is doing, and keeps reader focused on the steps of the sync loop more (rather then parsing through mapping details).

`to_central_sync_buffer_rows` is more explanatory imo (since the the struct is called CentralSyncBuffer)

repository/src/db_diesel/changelog_row.rs
andreievg: sure, perhaps a to do would be helpful. There was a bunch of times where i need to use a similar conversion, and it was nice to know I don't have to look it up again, just type `i32_to` and analyzer gives the option.

server/src/sync/sync_api_v3.rs
```diff
@@ -0,0 +1,181 @@
+#![allow(dead_code)]
+
+use reqwest::{
```
Chris-Petty, line 3: request?

repository/src/db_diesel/changelog_row.rs
andreievg: Yep it's up to the subscriber, but I don't see how this view can help subscriber do that in this example:

Subscriber does 3 records at a time, and here is a content of change log:

| id              | op     | record_id                                 |
|-----------------|--------|-------------------------------------------|
| 1               | upsert | a                                         |
| 2               | upsert | b                                         |
| 3               | upsert | c                                         |
| 4               | upsert | d                                         |
| 5               | delete | b                                         |
| 6               | upsert | c                                         |
| 7               | upsert | a                                         |
| 8               | upsert | c                                         |
| 9               | delete | c                                         |

subscriber asks for first 3, it would get this right ?

| id              | op     | record_id                                 |
|-----------------|--------|-------------------------------------------|
| 7               | upsert | a                                         |
| 5               | delete | b                                         |
| 9               | delete | c                                         |

It can't use the lowest id as it's next cursosr because record d is missed, if it just goes 0 = 3 for it's next cursor it would get

| id              | op     | record_id                                 |
|-----------------|--------|-------------------------------------------|
| 4               | upsert | d                                         |
| 5               | delete | b                                         |
| 9               | delete | c                                         |

So now it would need to double handle b and c

server/src/sync/sync_api_v3.rs
Chris-Petty: I hear it's an external lib 😁

repository/src/db_diesel/changelog_row.rs
clemens-msupply: Just added a test case to make sure I got the changelog_deduped views right for sqlite and postgres.

After the first query `changelogs(0, 3)` I get:
|id|op|record_id|
|-|--|-|
|4| upsert| d|
|5| delete|b|
|7| upsert|a|

The next cursor is then 7 + 1 = 8. The query `changelogs(8, 3)` then gives

|id|op|record_id|
|-|--|-|
|9| delete|c|

and I reached the end of the queue.

repository/src/tests.rs
clemens-msupply: I included the example from the comment in a test case: [4454665](https://github.com/openmsupply/remote-server/pull/748/commits/4454665e7f8730244d9c5872f9e67c2f626a0f53)

repository/src/db_diesel/changelog_row.rs
andreievg: Thanks, that looks great, I am wondering, there is not sort on the changelog_deduped, or in diesel, is the order guaranteed ?

server/src/sync/sync_api_v5.rs
andreievg: I've asked Chris about renaming those, to make it more clear

andreievg (approved):
Thanks for the changes

repository/src/db_diesel/changelog_row.rs
clemens-msupply: Good point thanks! the sqlite view was ordered but the postgres view wasn't

[287d170](https://github.com/openmsupply/remote-server/pull/748/commits/287d17039bc2dae17e25af80c0ea3cc38cbde03e)

server/src/sync/sync_api_v5.rs
clemens-msupply: Ok cool great, I leave it for now and updated it once v5 has been updated.

PR #750, opened 2022-02-08
746 stocktake batch mutation

clemens-msupply:


service/src/permission_validation.rs
```diff
@@ -46,6 +47,7 @@ fn all_permissions() -> HashMap<Resource, PermissionDSL> {
     map.insert(Resource::InsertStockTake, default());
     map.insert(Resource::UpdateStockTake, default());
     map.insert(Resource::DeleteStockTake, default());
+    map.insert(Resource::BatchStockTake, default());
```
andreievg, line 50: why would this be a new permission ? 
I think we want to check Insert/Update/Delete permissions for each operation in the batch

graphql/src/schema/mutations/stock_take/batch.rs
```diff
@@ -0,0 +1,296 @@
...
+        &ResourceAccessRequest {
+            resource: Resource::BatchStockTake,
+            store_id: Some(store_id.to_string()),
+        },
+    )?;
+
+    let service_provider = ctx.service_provider();
+    let service_ctx = service_provider.context()?;
+
+    let result = service_ctx.connection.transaction_sync(|_| {
```
andreievg, line 104: Hmm, not a big fan of accessing db/repo level functionality here, but not sure how else to do this apart from relocating to service or making another transaction helper in service context/provider (maybe provider can can have .context_transaction method that makes context with a transaction already

graphql/src/schema/mutations/stock_take/batch.rs
```diff
@@ -0,0 +1,296 @@
...
+    params(DeleteStockTakeLineResponse)
+))]
+pub struct MutationWithId<T: OutputType> {
+    pub id: String,
+    pub response: T,
+}
+
+#[derive(InputObject)]
+pub struct BatchStocktakeInput {
+    pub insert_stocktake: Option<InsertStockTakeInput>,
```
andreievg, line 52: this input is different from batch mutations of shipments, I think f/e wanted to be consistent with the way they use api (thus the need to add batch mutations), so I think we need to have the same input shape patters and shipments

graphql/src/schema/mutations/stock_take/batch.rs
```diff
@@ -0,0 +1,296 @@
...
+            let mut list = Vec::<MutationWithId<InsertStockTakeLineResponse>>::new();
+            for input in input {
+                list.push(MutationWithId {
+                    id: input.id.clone(),
+                    response: do_insert_stock_take_line(
+                        service_ctx,
+                        service_provider,
+                        store_id,
+                        input,
+                    )?,
```
andreievg, line 174: Hmm, even though we discussed that bad input shouldn't be actioned on by f/e and it's something they need to sort out during dev, record id or stringified input would still be useful, this can be done here or in the `do_` methods

andreievg (commented):
Look great, a couple of comment, main one is about input not being consistent with other batch mutations

graphql/src/schema/mutations/stock_take/batch.rs
clemens-msupply: yes agree, had similar thoughts. There is a little bit of logic in there which I haven't anticipated, e.g. to abort the transaction when there is a failure and this is tests in the graphql layer... I think the best is to have a batch method in service. I could then reuse the result mapping from the other gql endpoints...

I created a ticket #757

graphql/src/schema/mutations/stock_take/batch.rs
clemens-msupply: You mean something like:

let formatted_error = format!("Delete line {}: {:#?}", line.id, err);

graphql/src/schema/mutations/stock_take/batch.rs
clemens-msupply: ok can do

service/src/permission_validation.rs
clemens-msupply: Should I actually just have a MutateStocktake permission?

graphql/src/schema/mutations/stock_take/batch.rs
andreievg: yep, although as per our discussion maybe it's not that important (it sounds like the priority is to get batch and stocktake query in)

service/src/permission_validation.rs
andreievg: That would be great, we can be more granular in the future (that's what i did for requisition)

PR #751, opened 2022-02-09
Rename stock_take to stocktake

clemens-msupply:


andreievg (approved):
Approve all the renaming, although I haven't looked at the full diff

PR #752, opened 2022-02-09
747 stocktake query

### LINKED ISSUE #747: Add stocktake query

clemens-msupply:
Closes #747 

The endpoints returns a `Option<StocktakeNode>` which seem to be inconsistent. What should it return? Letting it return a Response object seems overkill since the API user would have to explicitly catch the not found error...


andreievg:
Wanted to review but there are too many changes from other branches, should this be based of some other branch ? (i tried changing base to the batch mutation branch but still had a lot of changes).

> The endpoints returns a Option<StocktakeNode> which seem to be inconsistent. What should it return? Letting it return a Response object seems overkill since the API user would have to explicitly catch the not found error...

I think its should be consistent with existing queries. Is it that much work ?
And f/e is currently handling all the other single queries with an error, and I think standard graphql on single queries is to return an error.

clemens-msupply:
Yes sorry it should be merged into 746-stocktake-rename...

PR #754, opened 2022-02-13
714 key value store

clemens-msupply:
Add a key value store table for storing various local server state.

Use the key value store to store the central pull cursor and remove the previously used table.


andreievg (approved):
Looks good, thanks, i like the helper get_string on the repo

PR #759, opened 2022-02-14
747 stocktake query

clemens-msupply:


graphql/src/schema/queries/stocktake.rs
```diff
@@ -117,3 +121,51 @@ pub fn stocktakes(
...
+                }),
+            };
+            Ok(result)
+        }
+
+        Err(err) => {
+            let formatted_error = format!("{:#?}", err);
+            let graphql_error = match err {
+                ListError::DatabaseError(err) => err.into(),
+                ListError::LimitBelowMin(_) => StandardGraphqlError::BadUserInput(formatted_error),
```
andreievg, line 165: I think in previous end points i've made the single query it's own service to avoid the ListError mapping in graphql, but I guess this still works

andreievg (approved):

graphql/src/schema/queries/stocktake.rs
clemens-msupply: Sorry missed this comment earlier. Will keep it in mind, but yes its no real logic in it so its fine I guess.

PR #770, opened 2022-02-15
769 stocktake by number

clemens-msupply:
I ended up to not introduce new methods in the stocktake service. After simplifying the `stocktake` endpoint the code was actually so simple that I didn't see the need for a special service method, i.e. the service only has one query method. 

Somehow I ran into a std error while adding stocktake tests and the output was a bit misleading. I now print the std error in the assert_graphql_query macro.

### DIFF
--- a/graphql/src/schema/queries/stocktake.rs
+++ b/graphql/src/schema/queries/stocktake.rs
@@ -21,6 +21,7 @@ use crate::schema::types::{
     },
     PaginationInput,
 };
+use crate::standard_graphql_error::list_error_to_gql_err;
 use crate::standard_graphql_error::validate_auth;
 use crate::standard_graphql_error::StandardGraphqlError;
 use crate::ContextExt;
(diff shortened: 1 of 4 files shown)


server/tests/graphql/mod.rs
```diff
@@ -181,6 +181,15 @@ macro_rules! assert_graphql_query {
         )
         .await;
 
+        match actual.get("errors").and_then(serde_json::Value::as_array) {
```
andreievg, line 184: is this really needed, standard error would still be printed neatly without it

server/tests/graphql/stocktake_query.rs
```diff
@@ -127,4 +127,60 @@ mod graphql {
...
+        let query = r#"query QueryStocktakeByNumber($storeId: String, $stocktakeNumber: String) {
+            stocktakeByNumber(storeId: $storeId, stocktakeNumber: $stocktakeNumber) {
+                ... on StocktakeNode {
+                    stocktakeNumber
+                }
+            }
+        }"#;
+        let expected_stocktake = mock_stocktake_a();
+        let variables = Some(json!({
+            "storeId": expected_stocktake.store_id,
```
andreievg, line 176: I thought we are testing mapping in graphql ? I mean that was the main reason for the complexity added with service traits (I know i varied from it in loaders, but i thought that was idea for simple queries ?)

graphql/src/schema/queries/stocktake.rs
```diff
@@ -158,14 +159,44 @@ pub fn stocktake(ctx: &Context<'_>, store_id: &str, id: &str) -> Result<Stocktak
...
+    match service.get_stocktakes(
+        &service_ctx,
+        store_id,
+        None,
+        Some(StocktakeFilter::new().stocktake_number(EqualFilter::equal_to_i64(stocktake_number))),
+        None,
+    ) {
+        Ok(mut stocktakes) => {
+            let result = match stocktakes.rows.pop() {
+                Some(stocktake) => StocktakeResponse::Response(StocktakeNode { stocktake }),
```
andreievg, line 192: that looks better then the query by id, but personally i would still make a seperate service (to keep graphql layer as purely mapping layer)

andreievg (approved):
Looks good, had a question about consistency in graphql tests (are we testing mapping between service and graphql or full integration test) ? Mainly asking this since it's more work and more complexity implementing the service trait, and it was done for that reason

server/tests/graphql/stocktake_query.rs
clemens-msupply: Don't we also test the success case? Think that would be good to test that there is no major issue with the api. At least thats the main purpose of this test (even though it does the full thing).

server/tests/graphql/mod.rs
clemens-msupply: I will give an example tomorrow

clemens-msupply: Yes it actually does show the errors in `expected` but it also shows the `actual` and `query`. Do we need this when there is an error? after this patch it shows something like:

```bash
thread 'graphql::stocktake_query::graphql::test_graphql_stocktakes_query' panicked at 'Request failed with standard error(s): [
  {
    "locations": [
      {
        "column": 175,
        "line": 1
      }
    ],
    "message": "Failed to parse \"Int\": Invalid number (occurred while parsing \"PaginationInput\")",
    "path": [
      "stocktakes"
    ]
  }
]', server/tests/graphql/stocktake_query.rs:122:9
```

graphql/src/schema/queries/stocktake.rs
clemens-msupply: For me its just another view on the same underlying data. Since we have already have a powerful service query function I think we can use it as it is. Can discuss later

Issue #768, opened 2022-02-15
Change batch mutations to use input object


andreievg:
Both inbound and outbound shipment batch mutations (to align with stocktake batch mutation)

Issue #749, opened 2022-02-08
Refactor/Alignments/Docs


andreievg:
Want to apply latest patterns consistently in the codebase and have solid docs (pre requisite of V1 release)

## Repositories

- [ ] Rename repositories (i.e. when we have [Type]Query to [Type] and [Type]Row
- [x] https://github.com/openmsupply/remote-server/issues/832
- [x] https://github.com/openmsupply/remote-server/issues/822
- [ ] upsert_many (see draft PR)
- [ ] one file for schema definition (see draft https://github.com/openmsupply/remote-server/pull/797)

### Graphql
- [x] https://github.com/openmsupply/remote-server/issues/849
- [x] https://github.com/openmsupply/remote-server/issues/638
- [x] https://github.com/openmsupply/remote-server/issues/699
- [x] https://github.com/openmsupply/remote-server/issues/781
- [x] Remove `graphql_client` from tests
- [ ] remove invoice_id from line mutations

### Service
- [x] Use service provider
- [ ] All service validate methods to map_err
- [x] https://github.com/openmsupply/remote-server/issues/850
- [ ] https://github.com/openmsupply/remote-server/issues/460 (might be done already ?)
- [x] https://github.com/openmsupply/remote-server/issues/739
- [ ] https://github.com/openmsupply/remote-server/issues/762
- [x] https://github.com/openmsupply/remote-server/issues/757

** Docs
- [ ] https://github.com/openmsupply/remote-server/issues/230
- [ ] Update all graphql resolvers with comments
- [ ] Update docs with patterns and shared diagrams
- [ ] Update api docs
- [ ] How to add new end point with mutations
- [ ] How to add new sync record
- [ ] Maybe (as this would be quite a bit of work) https://github.com/openmsupply/remote-server/issues/504

andreievg:
Closing, cary over is in https://github.com/openmsupply/remote-server/issues/993 and https://github.com/openmsupply/remote-server/issues/994

PR #773, opened 2022-02-15
768 change batch mutations to use input object

andreievg:
closes: #768 

Also added unallocated line mutations to outbound shipment.

There is a mix and match of trait services and older services taking connection manager, to decided to keep it simple for now and made a bulletpoint to address this (and use transaction, similar to stocktake batch mutation), anyways added https://github.com/openmsupply/remote-server/issues/749

### DIFF
--- a/graphql/src/schema/mutations/mod.rs
+++ b/graphql/src/schema/mutations/mod.rs
@@ -241,53 +241,19 @@ impl Mutations {
     async fn batch_inbound_shipment(
         &self,
         ctx: &Context<'_>,
-        insert_inbound_shipments: Option<Vec<InsertInboundShipmentInput>>,
-        insert_inbound_shipment_lines: Option<Vec<InsertInboundShipmentLineInput>>,
-        update_inbound_shipment_lines: Option<Vec<UpdateInboundShipmentLineInput>>,
-        delete_inbound_shipment_lines: Option<Vec<DeleteInboundShipmentLineInput>>,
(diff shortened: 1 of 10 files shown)


graphql/src/schema/mutations/mod.rs
```diff
@@ -706,6 +672,18 @@ impl InvoiceLineBelongsToAnotherInvoice {
...
+))]
+#[graphql(concrete(
+    name = "UpdateOutboundShipmentUnallocatedLineResponseWithId",
+    params(outbound_shipment::unallocated_line::UpdateResponse)
+))]
+#[graphql(concrete(
+    name = "DeleteOutboundShipmentUnallocatedLineResponseWithId",
+    params(outbound_shipment::unallocated_line::DeleteResponse)
+))]
 pub struct MutationWithId<T: OutputType> {
```
clemens-msupply, line 687: Don't know if you noticed, in my other PR I duplicated the MutationWithId struct for stocktake_batch.rs in an attempt to not edit this file and keep the number of "#[graphql(concrete(" annotations low...

can discuss more in a chat if thats a good idea...

clemens-msupply (approved):
Looks good thanks

graphql/src/schema/mutations/mod.rs
andreievg: that is a good idea, in #749 i've mentioned that we should replicate stocktake batch mutation

PR #775, opened 2022-02-15
Automatically run DB migrations when starting the server

clemens-msupply:
Checks for outstanding migrations and applies them.

Should also make developing remote sync a bit more easier, i.e. to start with a fresh DB.

Should automatically create an sql db on Android as well (untested)

### DIFF
--- a/repository/src/lib.rs
+++ b/repository/src/lib.rs
@@ -1,5 +1,7 @@
 #[macro_use]
 extern crate diesel;
+#[macro_use]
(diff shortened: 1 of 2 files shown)


andreievg:
if we are not deleting database, migrations wont' work if new fields are added or tables are inserted in between existing tables

clemens-msupply:
Yes correct you would have to clear the DB manually when changing existing migration files.

andreievg (approved):

PR #777, opened 2022-02-16
714 initial pull: get and store data from the central server

clemens-msupply:
No integration yet.

I panic if the initial pull fails or is there anything better to do right now?

I also did a bit of refactoring:
- split synchroniser into a central and a remote part
- the main synchroniser also manages the existing actor structs now which simplifies the code in the server lib

I was about to further move the existing actor structs into the main synchroniser and effectively making the synchroniser the actor (for sync it would manages how often it will get notified and also receives the sync messages). Just wanted to check first if you had other ideas how the actors should be implemented....


server/src/sync/remote_data_synchroniser.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    pub fn sync_queue_initalised(&self) -> Result<bool, RepositoryError> {
+        let value = self
+            .key_value_store
+            .get_string(KeyValueType::RemoteSyncQueueV5Initalised)?;
+        Ok(value.map(|value| value == DONE_VALUE).unwrap_or(false))
+    }
+
+    pub fn set_sync_queue_initalised(&self) -> Result<(), RepositoryError> {
+        self.key_value_store.set_string(
```
andreievg, line 156: wouldn't it be a boolean ?

clemens-msupply: Oh yes forgot to discuss. Should I add set_bool, set_i64, set_f64, etc? Was wondering if a string value is enough/simpler... But fine to add different column depending on the type

repository/migrations/postgres/2022-02-11T015:00_create_key_value_store/up.sql
```diff
@@ -1,6 +1,8 @@
 CREATE TYPE key_type AS ENUM (
     -- Cursor for pulling central records from the central server
-    'CENTRAL_SYNC_PULL_CURSOR'
+    'CENTRAL_SYNC_PULL_CURSOR',
+    'REMOTE_SYNC_QUEUE_V5_INITALISED',
```
andreievg, line 4: maybe REMOTE_SYNC_INITILISATION_STARTED, REMOTE_SYNC_INITILISATION_FINISHED (or something similar) ?

repository/src/db_diesel/remote_sync_buffer.rs
```diff
@@ -0,0 +1,59 @@
...
+    connection: &'a StorageConnection,
+}
+
+impl<'a> RemoteSyncBufferRepository<'a> {
+    pub fn new(connection: &'a StorageConnection) -> Self {
+        RemoteSyncBufferRepository { connection }
+    }
+
+    #[cfg(feature = "postgres")]
+    pub fn upsert_many(&self, rows: &Vec<RemoteSyncBufferRow>) -> Result<(), RepositoryError> {
```
andreievg, line 24: maybe we should be using this instead of just upsert in all of the repos (more often then not we insert more then one record for a table)

repository/src/db_diesel/remote_sync_buffer.rs
```diff
@@ -0,0 +1,59 @@
...
+        diesel::insert_into(remote_sync_buffer_dsl::remote_sync_buffer)
+            .values(rows)
+            .on_conflict(remote_sync_buffer_dsl::id)
+            .do_update()
+            .set(rows)
```
andreievg, line 29: AsChangeSet for RemoteSyncBufferRow

repository/src/schema/remote_sync_buffer.rs
```diff
@@ -0,0 +1,22 @@
+use super::diesel_schema::remote_sync_buffer;
+use diesel_derive_enum::DbEnum;
+
+#[derive(DbEnum, Debug, Clone, PartialEq, Eq)]
+#[DbValueStyle = "SCREAMING_SNAKE_CASE"]
+pub enum RemoteSyncBufferAction {
+    Create,
+    Update,
+    Delete,
```
andreievg, line 9: I'll talk to Chris, i am pretty sure this is never allowed (only owner can delete), is this coming from API or an assumption ?

server/src/lib.rs
```diff
@@ -96,20 +79,21 @@ pub async fn start_server(
         }
     }
     let mut running_sever = http_server.run();
-
-    let mut synchroniser = Synchroniser::new(&settings.sync).unwrap();
+    let mut synchroniser = Synchroniser::new(settings.sync, connection_manager).unwrap();
+    // Do the initial pull before doing anything else
+    synchroniser
```
andreievg, line 84: > I panic if the initial pull fails or is there anything better to do right now?

hmm i think it's better as just a log for now

server/src/lib.rs
```diff
@@ -96,20 +79,21 @@ pub async fn start_server(
         }
     }
     let mut running_sever = http_server.run();
-
-    let mut synchroniser = Synchroniser::new(&settings.sync).unwrap();
+    let mut synchroniser = Synchroniser::new(settings.sync, connection_manager).unwrap();
```
andreievg, line 82: To answer your question a bit more, i think there will need some channels rx/tx for manual sync (via api call) and for scheduler to reschedule next sync, not now of course, eventually

server/src/sync/central_data_synchroniser.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+impl CentralDataSynchroniser {
+    async fn pull_central_records(
+        &self,
+        connection: &StorageConnection,
+    ) -> Result<(), CentralSyncError> {
+        let central_sync_cursor = CentralSyncPullCursor::new(&connection);
+        let mut cursor: u32 = central_sync_cursor.get_cursor().unwrap_or_else(|_| {
+            info!("Initialising new central sync cursor...");
```
andreievg, line 45: should this be database error? already unwrapping to zero in get_cursor

server/src/sync/central_data_synchroniser.rs
```diff
@@ -0,0 +1,289 @@
...
+            0
+        });
+
+        // Arbitrary batch size.
+        const BATCH_SIZE: u32 = 500;
+
+        Ok(loop {
```
andreievg, line 52: I am a big fan of reducing indentation, personally would rather Ok(()) at the end, or break Ok(());

server/src/sync/central_data_synchroniser.rs
```diff
@@ -0,0 +1,289 @@
...
+            let sync_batch: CentralSyncBatchV5 = self
+                .sync_api_v5
+                .get_central_records(cursor, BATCH_SIZE)
+                .await
+                .map_err(|source| CentralSyncError::PullCentralSyncRecordsError { source })?;
+            let central_sync_records = central_sync_batch_records_to_buffer_rows(sync_batch.data)
+                .map_err(|err| {
+                CentralSyncError::PullCentralSyncRecordsError { source: err.into() }
+            })?;
+
```
andreievg, line 63: I think we want a transaction inside loop iteration so that we save batches of sync record together with update of central queue pointer

server/src/sync/central_data_synchroniser.rs
```diff
@@ -0,0 +1,289 @@
...
+            for central_sync_record in central_sync_records {
+                Self::insert_one_and_update_cursor(&connection, &central_sync_record)
+                    .await
+                    .map_err(
+                        |source| CentralSyncError::UpdateCentralSyncBufferRecordsError { source },
+                    )?;
+            }
+            info!("Successfully inserted central sync records into central sync buffer");
+
+            cursor = central_sync_cursor
```
andreievg, line 83: not a biggy but i think it's simpler to just check `central_sync_records` length and get cursor at the start of the loop

server/src/sync/central_data_synchroniser.rs
```diff
@@ -0,0 +1,289 @@
...
+
+            if cursor >= sync_batch.max_cursor - 1 {
+                info!("All central sync records pulled successfully");
+                break;
+            }
+        })
+    }
+
+    /// insert row and update cursor in a single transaction
+    async fn insert_one_and_update_cursor(
```
andreievg, line 95: ohh i see this is where transaction is.. ignore previous comment, upsert_many would be so much nicer here imo

server/src/sync/central_data_synchroniser.rs
```diff
@@ -0,0 +1,289 @@
...
+            records.append(&mut buffer_rows);
+        }
+
+        info!("Importing {} central sync buffer records...", records.len());
+        import_sync_records(connection, &records)
+            .await
+            .map_err(|source| CentralSyncError::ImportCentralSyncRecordsError { source })?;
+        info!("Successfully Imported central sync buffer records",);
+
+        // TODO needs to be done for M1 as name_store_joins are not synced yet but are required in API
```
andreievg, line 143: ooh, we want to remember to delete this when NameStoreJoins are synced

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            return Ok(());
+        }
+
+        info!("Pull remote records...");
+        self.pull_and_store_remote_records(connection)
+            .await
+            .map_err(|error| RemoteSyncError {
+                msg: "Failed to pull remote records",
+                source: error,
```
andreievg, line 47: I need to get up to speed with anyhow::Error, I might comment about this in future PRs

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /// Initalises the sync queue on the central server, pulls all records and stores them in the
+    /// DB.
+    pub async fn pull_and_store_remote_records(
+        &self,
+        connection: &StorageConnection,
+    ) -> anyhow::Result<()> {
+        let state = RemoteSyncState::new(connection);
+        if !state.sync_queue_initalised()? {
```
andreievg, line 72: I think this should be in initial_pull

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -0,0 +1,175 @@
...
+            info!("Initialising remote sync records...");
+            let _ = self.sync_api_v5.post_initialise().await?;
```
andreievg, line 74: why do we need let _ here and below ?

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -0,0 +1,175 @@
...
+
+            if let Some(data) = sync_batch.data {
+                let sync_ids: Vec<String> =
+                    data.iter().map(|record| record.sync_id.clone()).collect();
+
+                let _ = RemoteSyncBufferRepository::new(connection)
+                    .upsert_many(&remote_sync_batch_records_to_buffer_rows(&data)?);
+
+                info!("Acknowledging remote sync records...");
+                self.sync_api_v5.post_acknowledge_records(sync_ids).await?;
```
andreievg, line 95: should we do acknowledge and upsert_many inside a transact (to avoid duplicates if acknowledge fails)

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -0,0 +1,175 @@
...
+            if sync_batch.queue_length <= 0 {
+                break;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+fn to_row_action(action: &RemoteSyncActionV5) -> RemoteSyncBufferAction {
```
andreievg, line 108: I kind of like this on a RemoteSyncActionV5 struct and consume action

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -0,0 +1,175 @@
...
+    match action {
+        RemoteSyncActionV5::Create => RemoteSyncBufferAction::Create,
+        RemoteSyncActionV5::Update => RemoteSyncBufferAction::Update,
+        RemoteSyncActionV5::Delete => RemoteSyncBufferAction::Delete,
+        RemoteSyncActionV5::Merge => RemoteSyncBufferAction::Merge,
+    }
+}
+
+fn remote_sync_batch_records_to_buffer_rows(
+    records: &Vec<RemoteSyncRecordV5>,
```
andreievg, line 118: I think we should consume the vec at this stage and not clone below

server/src/sync/translation/mod.rs
```diff
@@ -192,40 +192,38 @@ fn integrate_record(
...
-    .await
-    .map_err(|error| match error {
-        TransactionError::Transaction { msg, level } => SyncImportError::as_integration_error(
-            RepositoryError::TransactionError { msg, level },
-            "",
-        ),
-        TransactionError::Inner(e) => e,
-    })
+    connection
+        .transaction(|con| async move {
```
andreievg, line 199: do we need async transaction ?

server/src/sync/remote_data_synchroniser.rs
andreievg: I kind of thought that was the idea, otherwise we would need to introduce boolean to string mapping or constantly hard code "true", "false", i think it's ok to just add them as needed (for now)

andreievg: Actually i just realised we are parsing string key value into u32 (pretty sure it's an artifact).. i think that's pretty bad

repository/migrations/postgres/2022-02-11T015:00_create_key_value_store/up.sql
andreievg: after reviewing the PR, i wonder if we need them ?

andreievg (approved):
Approving, don't want to hold you up, most things are pretty minor, and I think a lot of comments for code you copy pasted (so comments that I didn't make in initial sync PR)

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: ok will change to add more types to the table

repository/src/schema/remote_sync_buffer.rs
clemens-msupply: this was an assumption because the api struct had it as well. From real data I only got "update" so far. What about merge?

server/src/lib.rs
clemens-msupply: ok done also created a ticket to revise it before a release:

https://github.com/openmsupply/remote-server/issues/789

repository/src/db_diesel/remote_sync_buffer.rs
clemens-msupply: Postgres doesn't like it though :( Had to go through the vec and do it one by one.

server/src/sync/central_data_synchroniser.rs
clemens-msupply: The unwrap_or(0) is only for the case that the cursor wasn't set before

server/src/sync/central_data_synchroniser.rs
clemens-msupply: Done it doesn't reduced the indentation level but is IMHO nicer anyway

server/src/sync/central_data_synchroniser.rs
clemens-msupply: Yes the cursor is updated together with the write of each row in a transaction in insert_one_and_update_cursor.

We could probably write all rows in one DB request though.

server/src/sync/central_data_synchroniser.rs
clemens-msupply: will put it on my list!

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: done tanks

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: we don't thanks

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: done thanks

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: done

repository/src/schema/remote_sync_buffer.rs
andreievg: merge should exists, i just remembered something (not related to this PR), i've asked Chris, soft delete's of central data might be in V1

server/src/sync/translation/mod.rs
clemens-msupply: no I chaged it

server/src/sync/central_data_synchroniser.rs
andreievg: sure, i was just saying, in this case get_cursor() already returns 0 if no value

server/src/sync/remote_data_synchroniser.rs
andreievg: ohh true, it's better to have duplicates then not have any records, good call.

PR #780, opened 2022-02-16
714 number pull

clemens-msupply:
Is it correct that there is no RequestRequisition number in mSupply?


server/src/sync/translation_remote/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    Ok(())
+}
+
+fn do_translation(
+    sync_record: &RemoteSyncBufferRow,
+    records: &mut IntegrationRecord,
+) -> Result<(), SyncTranslationError> {
+    use IntegrationUpsertRecord::*;
+    if let Some(row) = LegacyNumberRow::try_translate_pull(sync_record)? {
```
andreievg, line 61: not going with visitor ?

server/src/sync/translation_remote/number.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+fn parse_number_name(value: String) -> Option<(NumberRowType, String)> {
+    let mut split = value.split("_for_store_");
+    let number_type = match split.next()? {
+        "stock_take_number" => NumberRowType::Stocktake,
+        "inventory_adjustment_serial_number" => NumberRowType::InventoryAdjustment,
+        "supplier_invoice_number" => NumberRowType::InboundShipment,
+        "customer_invoice_number" => NumberRowType::OutboundShipment,
+        // NumberRowType::RequestRequisition ?,
```
andreievg, line 56: > Is it correct that there is no RequestRequisition number in mSupply?

I've asked that question in mSupply dev, i guess so, just requisition_number_for_store

andreievg: ohh man this is ugly, it's just one, the only way i can think of to do it 'safely' is new field on mSupply, and if it exists consider requisition_number_for_store as response number and if it doesn't add another number for request number. In sync out make sure to populate this new field,

andreievg (approved):
looks good thanks, not sure what to do with requisition number for now

PR #792, opened 2022-02-18
714 name and stock line pull

clemens-msupply:
Both translation use a common trait now.

I must have merged something in the wrong order some changes didn't end up in develop so this PR contains some old changes...


server/src/sync/translation_remote/stock_line.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use serde::{Deserialize, Deserializer};
+
+use crate::sync::translation_central::SyncTranslationError;
+
+use super::{
+    IntegrationRecord, IntegrationUpsertRecord, RemotePullTranslation, TRANSLATION_RECORD_ITEM_LINE,
+};
+
+fn empty_str_as_option<'de, D: Deserializer<'de>>(d: D) -> Result<Option<String>, D::Error> {
```
andreievg, line 12: nice, can see this being a common on in remote record translations

server/src/sync/translation_remote/stock_line.rs
```diff
@@ -0,0 +1,105 @@
...
+                SyncTranslationError {
+                    table_name,
+                    source,
+                    record: sync_record.data.clone(),
+                }
+            })?;
+
+        Ok(Some(IntegrationRecord::from_upsert(
+            IntegrationUpsertRecord::StockLine(StockLineRow {
+                id: data.ID.clone(),
```
andreievg, line 67: why do we need to clone ?

andreievg (approved):
Looks good, only reviewed the translations for stock_line

server/src/sync/translation_remote/stock_line.rs
clemens-msupply: yes already done in my current work...

server/src/sync/translation_remote/stock_line.rs
clemens-msupply: thanks removed it [fa9d827](https://github.com/openmsupply/remote-server/pull/792/commits/fa9d827d8b53bd77e8177258f765be699507905f)

Issue #714, opened 2022-01-28
Remote server sync


clemens-msupply:
# mSupply tables to be synced (both ways):

| table               | sync |
|---------------------|------|
| pref                | X    |
| pref_blob           | X    |
| purchase_order      | X    |
| purchase_order_line | X    |
| name_tag_join       | X    |
| user                | ?    |
| user_store          | ?    |
| name                | X  |
| name_store_join     | yes  |
| number              | yes  (only currently used numbers)|
| number_reuse        | X    |
| Stock_take          | yes  |
| Stock_take_lines    | yes  |
| transact            | yes  |
| trans_line          | yes  |
| item_line           | yes  |
| item_store_join     | no  |
| requisition  (mSupply table name?)| yes| 
| requisition line (mSupply table name?)| yes|

### Pull
- [x] number
- [x] name_store_join
- [x] stocktake
- [x] stocktake_line
- [x] transact (shipment)
- [x] trans_line (shipment_line)
- [x] item_line (stock_line)
- [x] requisition
- [x] requisition_line

### Push
- [x] sql triggers
- [x] push trait and push logic
#### Push translations
- [x] number
- [x] name_store_join
- [x] stocktake
- [x] stocktake_line
- [x] transact (shipment)
- [x] trans_line (shipment_line)
- [x] item_line (stock_line)
- [x] requisition
- [x] requisition_line

### TODOs
- [x] address name_store_join TODO in server/src/sync/central_data_synchroniser.rs
- [ ] looks like pack_size is not editable in mSupply, so yeah. I think we need to restrict pack_size editability on Update/InsertStocktakeLineInput to only allow it for new stock_lines (https://github.com/openmsupply/remote-server/pull/814/files/6eda1263ef4f8bf4caf0a12fabcc7ac34980673f#diff-5cff3cda3f7a6d4f4d4eeff886aef5bf400f88d0cf5e70d51940d38a273f2ad1)
- [x] stocktake_line: item_line_id optional? Andrei: It looks like for new lines (where stock_line does not exist, item_line_id is generated but item_line is not made until finalisation). But I think this is the only line that appears with snapshot quantity = 0, so we can use that and set stock_line_id = Null rather the item_line_id (SOLUTION: item_line_id is always set)

andreievg:
oops sorry i never replied

> name 

this is central

> user

Sounds like we are doing fetch for this on every logging and then storing locally (if can't connect to central then use local), btw this is where permissions will come from and be updated

> user store

[...]

PR #794, opened 2022-02-20
Move tests and test data closer together

clemens-msupply:
#714


andreievg (commented):
Hmm, it's away from code that does the translation but closer to the data, why is it better to have them close data vs logic ?

clemens-msupply:
Its very annoying to write them / jumping from the data to the test file. However, since the whole translation including DB write is tested anyway there isn't a real need for them anyway... Plan to pass a StorageConnection to try_translate and plan to remove these tests then.

Issue #530, opened 2021-11-03
Invoice Status


andreievg:
New invoices statuses were suggested and since existing mSupply status needs to remain (and be in sync with mSupply status), new field to be added in existing mSupply for new omSupply status.

omSupply will use new field but will also translate it to mSupply field during sync (i.e. for `in sync` -> if status_new has a value it will be used as status_new, otherwise default status from mSupply status will be used, for `out sync`, omSupply status will sync as status_new, and will also map mSupply.status from omSupply.status, as per table below)

* (mSupply) New status: `transact.status_new`

## Outbound Shipment

#### Status Translations

| mSupply status | omSupply.status/mSupply.status_new | Comments                                                                                                                                                                                                             |
|----------------|------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| nw             |                                    |                                                                                                                                                                                                                      |
| sg             | NEW/DRAFT                          |                                                                                                                                                                                                                      |
| sg             | ALLOCATED                          | Allocated indicates that it's ready for picking, can only be set when all lines are allocated (suggestion was for UI to present user with option to either delete unallocated lines or create new invoice with them) |
| cn             | PICKED                             |                                                                                                                                                                                                                      |
| fn             | SHIPPED                            |                                                                                                                                                                                                                      |
| fn             | DELIVERED                          | Either enterable or automatically set when Inbound Shipment is set as Delivered                                                                                                                                      |
| fn             | VERIFIED                           | Automatically set when Inbound Shipment is set as Verified                                                                                                                                                           |

#### Stock Adjustments

|    | status    | adjustments                          |
|----|-----------|--------------------------------------|
| nw |           |                                      |
| sg | NEW/DRAFT | available_number_of_packs is reduced |
| sg | ALLOCATED | available_number_of_packs is reduced |
| cn | PICKED    | total_number_of_packs is reduced     |

#### DateTime

It should be ok to have a seperate date time field to record every date and time change:

| mSupply field | omSupply field     | Comment                                       |
|---------------|--------------------|-----------------------------------------------|
| entry_date    |                    | omSupply entry_datetime                       |
| entry_time    |                    | omSupply entry_datetime                       |
|               | entry_datetime     | mSupply entry_date/time                       |
|               | allocated_datetime |                                               |
| confirm_date  |                    | omSupply picked_datetime                      |
| confirm_time  |                    | omSupply picked_datetime                      |
[...]

PR #795, opened 2022-02-21
714 name store join shipment pull

clemens-msupply:
Translations for name_store_join and shipments. (shipments has some questions in it...)

RemotePullTranslation::try_translate_pull now takes a StorageConnection to be able to lookup other table needed for a translation.

Also refactored the sync errors that are shared between central and remote.


clemens-msupply:
Noticed that one test is failing / wasn't enabled will fix in followup PR

server/src/sync/translation_remote/shipment.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Supplier invoice
+    #[serde(rename = "si")]
+    Si,
+    /// Customer invoice
+    #[serde(rename = "ci")]
+    Ci,
+    // customer credit
+    //cc,
+    // supplier credit
+    //sc,
```
andreievg, line 27: I think we can do default with serde(other), there are quite a few transact types (don't really know them all)

andreievg: we can of course ignore them for now, but would help avoid unexpected `de` errors

server/src/sync/translation_remote/shipment.rs
```diff
@@ -0,0 +1,164 @@
...
+                name_id: data.name_ID,
+                // TODO is None correct?
+                name_store_id: None,
+                invoice_number: data.invoice_num,
+                r#type: shipment_type(&data._type),
+                status: shipment_status(&data.status),
+                on_hold: data.hold,
+                comment: data.comment,
+                their_reference: data.their_ref,
+                created_datetime: data_and_time_to_datatime(data.entry_date, data.entry_time),
```
andreievg, line 128: there would be a new field added (text field) for omSupply full timestamps for new statuses and existing statuses. In translations we would check if new field is present if it is it would be use otherwise deduce from date and time.

(good to add a TODO in the issue for it)

server/src/sync/translation_remote/shipment.rs
```diff
@@ -0,0 +1,164 @@
...
+                allocated_datetime: None,
+                picked_datetime: None,
+                shipped_datetime: data
+                    .ship_date
+                    .map(|ship_date| data_and_time_to_datatime(ship_date, 0)),
```
andreievg, line 133: `data` -> `date`

server/src/sync/translation_remote/shipment.rs
```diff
@@ -0,0 +1,164 @@
...
+                }
+            })?;
+
+        let confirm_time = data.confirm_time;
+        Ok(Some(IntegrationRecord::from_upsert(
+            IntegrationUpsertRecord::Shipment(InvoiceRow {
+                id: data.ID,
+                store_id: data.store_ID,
+                name_id: data.name_ID,
+                // TODO is None correct?
```
andreievg, line 120: It should be the store associated with the name (if any), Name `domain` object has store_id option

server/src/sync/translation_remote/shipment.rs
```diff
@@ -0,0 +1,164 @@
...
+fn shipment_type(_type: &LegacyTransactType) -> InvoiceRowType {
+    match _type {
+        LegacyTransactType::Si => InvoiceRowType::InboundShipment,
+        LegacyTransactType::Ci => InvoiceRowType::OutboundShipment,
+    }
+}
+
+fn shipment_status(status: &LegacyTransactStatus) -> InvoiceRowStatus {
+    match status {
+        LegacyTransactStatus::Nw => InvoiceRowStatus::New,
```
andreievg, line 157: https://github.com/openmsupply/remote-server/issues/530.

it's based on invoice type:
Screen Shot 2022-02-23 at 11 21 15 AM
Screen Shot 2022-02-23 at 11 21 23 AM

andreievg (approved):
Approving but a change request of status translations

server/src/sync/translation_remote/shipment.rs
clemens-msupply: Added in [497e5e4](https://github.com/openmsupply/remote-server/pull/795/commits/497e5e49f34ed0056e20a476c635344d857a4fa7)

I just remembered why I didn't included the known types previously. My original plan was to let serde fail and the error will later be logged. What I am doing now is to include all known types + Other into the enum and only through an error if the type can't be mapped (because there is no equivalent in omSupply)

server/src/sync/translation_remote/shipment.rs
clemens-msupply: thanks did this so many times...

server/src/sync/translation_remote/shipment.rs
clemens-msupply: thanks done in [d0caa68](https://github.com/openmsupply/remote-server/pull/795/commits/d0caa68196fa988af15fe8d67e09e555d87358e1)

server/src/sync/translation_remote/shipment.rs
clemens-msupply: Thanks! I improved the mapping in [6e2823c](https://github.com/openmsupply/remote-server/pull/795/commits/6e2823c966f6f0e84d02242a038aaad0a718bd8c)

For outbound shipments I wasn't sure how to figure out if sg means Allocated. I could check if there are un-allocated lines but it could also be that is hasn't been set to allocated (for the cases that omSupply pulls its own data back from the central server). Or is there another column in mSupply that indicates the Allocated status?

andreievg: I think it's safe to skip this until we get new omSupply field in place. I think for translation from legacy rows sg = new/draft

andreievg:
Thanks!
I think it would be good to revisit these translations once we get om_ fields for new status and datetimes in mSupply, maybe we can keep things simplified. 

You'll notice in the issue i've linked there is also date time translations, where confirm_date means something else based on record type. Anyways don't want to hold up the merge so i think it's ok to merge it and look at in depth next week

PR #797, opened 2022-02-21
Example Repository schema and row definitions in one file

andreievg:
Before I go and refactor repositories, want to run this past you @clemens-msupply.

* Removed requisition from db_diesel_schema and relocated to db_diesel/requistion/schema
* Relocate filter and sort definitions to db_diesel/requistion/schema
* Remove requisition row frome scehma/requisition to db_diesel/requistion/schema
* Add export of requisition types tp schema/mod (to avoid mass import changes for now)

Smaller repositories can be in one file but otherwise similar structure shown here.


clemens-msupply (commented):
Cool, I like where it is going! Some suggestion:
- In the schema file I would only put the diesel table! macro, the row struct and the row enums (everything directly DB table related)
- In the repo file I would put the Repository and the filter and sort defs (filter and sort are not directly DB specific) 

For my taste I would prefer something like `requistion_schema.rs` and `requisition_repo.rs`

andreievg:
going to close this, to keep PR list clean, will reference when doing repository juggling

PR #798, opened 2022-02-21
714 shipment line pull

clemens-msupply:
Also:
- Some test fixes
- Small refactoring and more mock data


server/src/sync/translation_remote/shipment_line.rs
```diff
@@ -0,0 +1,123 @@
...
+use super::{
+    empty_str_as_option, zero_date_as_option, IntegrationRecord, RemotePullTranslation,
+    TRANSLATION_RECORD_TRANS_LINE,
+};
+
+#[derive(Deserialize)]
+enum LegacyTransLineType {
+    #[serde(rename = "stock_in")]
+    StockIn,
+    // TODO check stock_out exist in mSupply:
```
andreievg, line 20: https://github.com/openmsupply/remote-server/issues/530

Screen Shot 2022-02-23 at 11 25 06 AM

Although the comment suggest i haven't quite confirmed (but at least service needs to be added)

server/src/sync/translation_remote/shipment_line.rs
```diff
@@ -0,0 +1,123 @@
+use chrono::NaiveDate;
```
andreievg, line 1: I think for both shipment and shipment_line we should name translation files as invoice and invoice_line, since invoices are not just shipments (also inventory adjustments, credits etc..)

server/src/sync/translation_remote/shipment_line.rs
```diff
@@ -0,0 +1,123 @@
...
+                })
+            }
+        };
+
+        Ok(Some(IntegrationRecord::from_upsert(
+            IntegrationUpsertRecord::ShipmentLine(InvoiceLineRow {
+                id: data.ID,
+                invoice_id: data.transaction_ID,
+                item_id: data.item_ID,
+                item_name: data.item_name,
```
andreievg, line 97: we need to preserver historic name of the item on shipment, this should come from trans_line. Also no need to query for item

server/src/sync/translation_remote/shipment_line.rs
```diff
@@ -0,0 +1,123 @@
...
+                item_code: item.code,
+                stock_line_id: data.item_line_ID,
+                location_id: data.location_ID,
+                batch: data.batch,
+                expiry_date: data.expiry_date,
+                pack_size: data.pack_size,
+                cost_price_per_pack: data.cost_price,
+                sell_price_per_pack: data.sell_price,
+                total_before_tax: data.sell_price * data.quantity as f64,
```
andreievg, line 106: as per our convo, I think there needs to be either a TODO added to issue or we use sell and cost based on shipment type

andreievg (approved):

server/src/sync/translation_remote/shipment_line.rs
clemens-msupply: Agree but I will leave that for later to avoid conflicts

server/src/sync/translation_remote/shipment_line.rs
clemens-msupply: thanks added them [c977e4a](https://github.com/openmsupply/remote-server/pull/798/commits/c977e4ae93e7e854d0a777ec62363dc16a0fcf78)

server/src/sync/translation_remote/shipment_line.rs
clemens-msupply: Its only for the code and the code is not present in trans_line

server/src/sync/translation_remote/shipment_line.rs
clemens-msupply: thanks done

server/src/sync/translation_remote/shipment_line.rs
andreievg: oops sorry, missed that

andreievg:
thanks

PR #814, opened 2022-02-22
714 stock take pull

clemens-msupply:
Sync pull for stocktake and stocktake line

### DIFF
--- a/server/src/sync/translation_remote/mod.rs
+++ b/server/src/sync/translation_remote/mod.rs
@@ -3,17 +3,18 @@ use log::{info, warn};
 use repository::{
     schema::{
         InvoiceLineRow, InvoiceRow, NameStoreJoinRow, NumberRow, RemoteSyncBufferRow, StockLineRow,
-        StocktakeRow,
+        StocktakeLineRow, StocktakeRow,
     },
     InvoiceLineRowRepository, InvoiceRepository, NameStoreJoinRepository, NumberRowRepository,
-    RepositoryError, StockLineRowRepository, StocktakeRowRepository, StorageConnection,
-    TransactionError,
+    RepositoryError, StockLineRowRepository, StocktakeLineRowRepository, StocktakeRowRepository,
+    StorageConnection, TransactionError,
 };
 use serde::{Deserialize, Deserializer};
 
 use crate::sync::translation_remote::{
     name_store_join::NameStoreJoinTranslation, shipment::ShipmentTranslation,
     shipment_line::ShipmentLineTranslation, stocktake::StocktakeTranslation,
+    stocktake_line::StocktakeLineTranslation,
 };
 
 use self::number::NumberTranslation;
@@ -27,6 +28,7 @@ mod shipment;
 mod shipment_line;
 mod stock_line;
 mod stocktake;
+mod stocktake_line;
 #[cfg(test)]
 pub mod test_data;
 
@@ -38,6 +40,7 @@ pub enum IntegrationUpsertRecord {
     Shipment(InvoiceRow),
     ShipmentLine(InvoiceLineRow),
     Stocktake(StocktakeRow),
+    StocktakeLine(StocktakeLineRow),
 }
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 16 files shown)


server/src/sync/translation_remote/mod.rs
```diff
@@ -185,3 +204,10 @@ pub fn zero_date_as_option<'de, D: Deserializer<'de>>(d: D) -> Result<Option<Nai
     Ok(s.filter(|s| s != "0000-00-00")
         .and_then(|s| NaiveDate::parse_from_str(&s, "%Y-%m-%d").ok()))
 }
+
+pub fn data_and_time_to_datatime(date: NaiveDate, seconds: i64) -> NaiveDateTime {
```
andreievg, line 208: `data` -> `date`

server/src/sync/translation_remote/stocktake.rs
```diff
@@ -0,0 +1,94 @@
...
+        Ok(Some(IntegrationRecord::from_upsert(
+            IntegrationUpsertRecord::Stocktake(StocktakeRow {
+                id: data.ID,
+                store_id: data.store_ID,
+                stocktake_number: data.serial_number,
+                comment: data.comment,
+                description: data.Description,
+                status: stocktake_status(&data.status),
+                created_datetime: data_and_time_to_datatime(data.stock_take_created_date, 0),
+                // TODO finalise doesn't exist in mSupply?
```
andreievg, line 80: hmm, not sure, might need to ask mSuppliers, but either way I think new text field would be added

server/src/sync/translation_remote/stocktake_line.rs
```diff
@@ -0,0 +1,88 @@
...
+    location_id: Option<String>,
+    #[serde(deserialize_with = "empty_str_as_option")]
+    comment: Option<String>,
+    snapshot_qty: i32,
+    snapshot_packsize: i32,
+    stock_take_qty: i32,
+    is_edited: bool,
+    // TODO is this optional?
+    #[serde(deserialize_with = "empty_str_as_option")]
+    item_line_ID: Option<String>,
```
andreievg, line 31: Hmm, you can create stocktake in mSupply via `Item` tab and can check database via Record Browser. It looks like for new lines (where stock_line does not exist, item_line_id is generated but item_line is not made until finalisation). But I think this is the only line that appears with snapshot quantity = 0, so we can use that and set stock_line_id = Null rather the item_line_id

server/src/sync/translation_remote/stocktake_line.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        let data = serde_json::from_str::<LegacyStocktakeLineRow>(&sync_record.data).map_err(
+            |source| SyncTranslationError {
+                table_name,
+                source: source.into(),
+                record: sync_record.data.clone(),
+            },
+        )?;
+
+        // TODO is this correct?
```
andreievg, line 62: yep, just checked with record browser, it looks like it

server/src/sync/translation_remote/stocktake_line.rs
```diff
@@ -0,0 +1,88 @@
...
+                stock_line_id: data.item_line_ID,
+                location_id: data.location_id,
+                comment: data.comment,
+                snapshot_number_of_packs: data.snapshot_qty,
+                counted_number_of_packs,
+                item_id: data.item_ID,
+                batch: data.Batch,
+                expiry_date: data.expiry,
+                // TODO: correct?
+                pack_size: Some(data.snapshot_packsize),
```
andreievg, line 81: looks like pack_size is not editable in mSupply, so yeah. I think we need to restrict pack_size editability on Update/InsertStocktakeLineInput to only allow it for new stock_lines

andreievg (approved):
Are inventory adjustments syncing ?

server/src/sync/translation_remote/stocktake_line.rs
clemens-msupply: I put it as a todo in the issue for now will discuss with you later

Issue #781, opened 2022-02-17
Split graphql into crates


andreievg:
Will outline a proposal of structure, reference PR: https://github.com/openmsupply/remote-server/pull/701

### Tasks

- [x] ~~Add dependencies crate~~  macros don't work if async_graphql is re-exported ?
- [x] https://github.com/openmsupply/remote-server/pull/815
- [x] https://github.com/openmsupply/remote-server/pull/818
- [x] https://github.com/openmsupply/remote-server/pull/819

Structure describe below does not reflect final implementation, see https://github.com/openmsupply/remote-server/pull/819

Below living in `graphql/`

#### graphql_core
* standard error
* context
* auth helper
* generic filter inputs and pagination input
* generic simple errors (that don't rely on other types)

#### graphql_common
* all types (they are circular, and if they are not yet circular they are likely to be circular in the future)
* complex generic errors (that may return types)

#### graphql

used by server

* remainder of lib.rs

#### graphql_outbound_shipment_line
etc..

Below living in top level

#### graphql_external

~~When graphql crates are split we would have external crate dependencies in each of the tomls, for now going toimport and re-export them all in graphql_core, but considering doing an overall `external` crate that import and re-export all external crates~~ macros don't work if async_graphql is re-exported ?

PR #815, opened 2022-02-22
781 Remove external deps from graphql types

### LINKED ISSUE #802: StocktakeLineNode item not optional

andreievg:
https://github.com/openmsupply/remote-server/issues/781
closes #802 (this was also done here since from was removed for item and required a change in resolvers)

Aftermath from removing types in types/mod.rs, i still kept NodeError for now. And most areas that I touched i tried to use latest pattern (even if it's out of scope of this issue, but will be done as part of the epic)

Also haven't yet relocated things from types to queries (like some sorts/filters, etc)

Added map_filter macro to help

### DIFF
--- a/graphql/src/schema/types/sort_filter_types.rs
+++ b/graphql/src/schema/types/sort_filter_types.rs
@@ -1,60 +1,17 @@
-use crate::schema::queries::{ItemSortFieldInput, NameSortFieldInput, StocktakeSortFieldInput};
-
 use super::{
-    InvoiceNodeStatus, InvoiceNodeType, InvoiceSortFieldInput, LocationSortFieldInput,
-    RequisitionNodeStatus, RequisitionNodeType, StocktakeNodeStatus,
+    InvoiceNodeStatus, InvoiceNodeType, RequisitionNodeStatus, RequisitionNodeType,
+    StocktakeNodeStatus,
 };
 
 use domain::{
     invoice::{InvoiceStatus, InvoiceType},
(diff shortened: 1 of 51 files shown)


graphql/src/schema/types/sort_filter_types.rs
```diff
@@ -112,55 +52,36 @@ impl<I: InputType> EqualFilterInput<I> {
...
-            not_equal_to,
+            equal_to: $from.equal_to.map($f),
+            not_equal_to: $from.not_equal_to.map($f),
+            equal_any: $from
+                .equal_any
+                .map(|inputs| inputs.into_iter().map($f).collect()),
             not_equal_all: None,
         }
-    }
+    }};
```
clemens-msupply, line 65: Does it need to be a macro?

clemens-msupply (approved):
Looks good thanks

graphql/src/schema/types/sort_filter_types.rs
andreievg: Hmm could do without it and just duplicate the logic for each conversion from `type`. Previously could do a generic method but got rid of generic filter struct (since it would require dependencies from queries). 
I thought the macro was super simple and reduces number of changes when new end point with new type to filter by, but not feeling very strong towards it

clemens-msupply: Ok I see thanks, missed that EqualFilter is generic...

PR #818, opened 2022-02-23
781 remove nested unions for single record resolvers

andreievg:
#781 

Even though it's not quite in scope for the issue, it would reduce f/e changes.


clemens-msupply (approved):
Yes that looks simpler now :)

PR #819, opened 2022-02-23
781 split graphql crate

### LINKED ISSUE #781: Split graphql into crates

andreievg:
closes #781 

How it went down:

* Started with core crate, just kept context and auth stuff in lib.rs, kept standard graphql error and loaders. Relocated generic filters, pagination and generic simple error (that doesn't require any types)

* Added types, which is just from schema/types, and added generic graphql errors (the ones that require types), moved some errors to generic simple error in core crate
* Added invoice, invoice_line, location, stocktake, stocktake_line, requisition and requisition_line crates (both mutation and queries)
* Added general_queries and general_mutation crates
* Added batched mutation crate (for all batched mutations)

Did some relocation of errors to either generic simple error (core crate) or generic error (types crate).

Removed everything from graphql parent crate, imported all crates into it, it also houses queries and mutations and graphql entry point.

Ended up duplicating map_filter macro (exporting macros from crates is a bit more involved, so did a lazy way, probably better to just remove it ?

I think the only change I did to existing logic is one error that returned InvoiceResponse had the response removed.

Overall happy with result but still needs some adjustment in the future, should be easy to do now, and seems quick to recompile 1-2 seconds.


clemens-msupply (approved):
Very cool! I didn't go through all changes but just had a look at a checkout. Some comment/ideas:

- Split the main query and mutation objects. I had it in my draft PR and it wasn't completely smooth but it would have the advantage that the full endpoint definitions are located in one crate and you don't have to edit the core crate when doing smaller endpoint changes.
- Move batch mutations into their "main" crate? or did you already plan to have one overall batch mutation?
- Similar, move loaders into the crates they are associated with? e.g. requisition loader into requisition crate... Not sure if that good idea.

andreievg:
> Similar, move loaders into the crates they are associated with? e.g. requisition loader into requisition crate... Not sure if that good idea.

They are used in types, so need to be a level below or in types, I thought core would be a good place

> Split the main query and mutation objects. I had it in my draft PR and it wasn't completely smooth but it would have the advantage that the full endpoint definitions are located in one crate and you don't have to edit the core crate when doing smaller endpoint changes.

hmm i am. not sure I understand, core crate shouldn't be touched unless you are adding some generic simple errors or adding something to standard error or context or loaders. 

> Split the main query and mutation objects

Do you mean the general_queries and general should be combined ?

> Move batch mutations into their "main" crate? or did you already plan to have one overall batch mutation?

I was trying to find a balance (theoretically) of compile speed vs partitioning that makes sense.  I thought that since  both main and batch import quite a lot of types and probably generate a lot of code through macros it might be good to split them. 

But once again, it's theoretical, can try fine tuning things, but I already spend a lot of time on this. Would be great if we can leave it for now and see what needs to be changed as we start using the new structure (unless there is something obviously messy atm)

Issue #822, opened 2022-02-24
Use composition domain object (remove domain crate)


andreievg:
* [x] https://github.com/openmsupply/remote-server/pull/824
* [x] https://github.com/openmsupply/remote-server/pull/825
* [x] https://github.com/openmsupply/remote-server/pull/828
* [x] https://github.com/openmsupply/remote-server/pull/829
* [x] https://github.com/openmsupply/remote-server/pull/831
* [x] https://github.com/openmsupply/remote-server/pull/831
* [x] https://github.com/openmsupply/remote-server/pull/830
* [x] https://github.com/openmsupply/remote-server/pull/827

PR #824, opened 2022-02-24
822 Use composite domain object for invoice line

andreievg:
https://github.com/openmsupply/remote-server/issues/822

I actually stared re alignment with this, but then had to change a few graphql things so thought it was worth doing the split first. And i think it was, it's very fast now, thanks for the idea : )

Also changed logic in post sync processor, even though a service would most likely be used instead of generating them manually, needed for fix in compilation errors.


clemens-msupply (approved):
Looks all good to me

PR #842, opened 2022-02-25
836 Unblock batch delete

### LINKED ISSUE #836: Unblock batch deletes
For now just delete lines (use service) and make an [issue](https://github.com/openmsupply/remote-server/issues/839) and link in code comment to do it 'better'

andreievg:
closes #836 

* Remove check for lines existing
* Add removal of lines when record is deleted
* Add TODO's with link to issue
* Had to butcher transactions in shipment delete (in sqlite can't have two concurrent transactions and line services take connection manager)

Didn't add test but did a manual test

Used internal error for now in older pattern graphql mapping.

### DIFF
--- a/graphql/graphql_schema/schema.graphql
+++ b/graphql/graphql_schema/schema.graphql
@@ -115,13 +115,13 @@ type CannotDeleteInvoiceWithLines implements DeleteErrorInterface & DeleteInboun
 type CannotDeleteRequisitionWithLines implements DeleteRequestRequisitionErrorInterface {
 	description: String!
 }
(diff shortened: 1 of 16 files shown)


clemens-msupply (approved):
Regarding tx, you mean deleting lines are done in a new connection? this is only temp right? once the service takes a &StorageConnection this is not a problem?

andreievg:
> Regarding tx, you mean deleting lines are done in a new connection? this is only temp right? once the service takes a &StorageConnection this is not a problem?

Yep, and also we would remove that delete of lines soon.

PR #843, opened 2022-02-25
714 push core

clemens-msupply:
Some initial (untested) push implementation

- After the initial pull the push cursor is set to the end of the changelog so that push starts from this point.
- Push translation for Numbers
- Some push tests that build upon existing pull test data. Pull and push test data could be merged at a later stage but I didn't want to edit all existing test data definitions in one go.

#714

### DIFF
--- a/server/src/sync/remote_data_synchroniser.rs
+++ b/server/src/sync/remote_data_synchroniser.rs
@@ -1,6 +1,6 @@
 use log::info;
 use repository::{
-    schema::{KeyValueType, RemoteSyncBufferAction, RemoteSyncBufferRow},
+    schema::{ChangelogRow, KeyValueType, RemoteSyncBufferAction, RemoteSyncBufferRow},
     ChangelogRepository, KeyValueStoreRepository, RemoteSyncBufferRepository, RepositoryError,
     StorageConnection,
 };
@@ -183,33 +183,7 @@ impl RemoteDataSynchroniser {
                 "Remote push: Translate {} changelogs to push records...",
                 changelogs.len()
             );
-            let mut out_records: Vec<PushRecord> = Vec::new();
-            for changelog in changelogs {
-                translate_changelog(connection, &changelog, &mut out_records)?;
-            }
-
-            info!("Remote push: Send records to central server...");
-            let records: Vec<RemotePostRecordV3> = out_records
-                .into_iter()
-                .map(|record| match record {
-                    PushRecord::Upsert(record) => RemotePostRecordV3 {
-                        sync_id: format!("{}", record.sync_id),
-                        sync_type: SyncTypeV3::Update,
-                        store_id: record.store_id,
-                        record_type: record.table_name.to_string(),
-                        record_id: record.record_id.clone(),
-                        data: Some(record.data),
-                    },
-                    PushRecord::Delete(record) => RemotePostRecordV3 {
-                        sync_id: format!("{}", record.sync_id),
-                        sync_type: SyncTypeV3::Delete,
-                        store_id: None,
-                        record_type: record.table_name.to_string(),
-                        record_id: record.record_id.clone(),
-                        data: None,
-                    },
-                })
-                .collect();
+            let records = translate_changelogs_to_push_records(connection, changelogs)?;
 
             self.sync_api_v3
                 .post_queued_records(self.site_id, self.central_server_site_id, &records)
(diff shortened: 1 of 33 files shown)


server/src/sync/remote_data_synchroniser.rs
```diff
@@ -76,6 +88,14 @@ impl RemoteDataSynchroniser {
 
         state.set_initial_remote_data_synced()?;
 
+        // Update push cursor after initial sync, i.e. set it to the end of the just received data
+        // so we only push new data to the central server
+        let cursor = ChangelogRepository::new(connection)
```
andreievg, line 93: Personally i would prefer this under 'state', maybe stat.advance_push_cursor

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -147,6 +167,74 @@ impl RemoteDataSynchroniser {
...
+                "Remote push: Translate {} changelogs to push records...",
+                changelogs.len()
+            );
+            let records = translate_changelogs_to_push_records(connection, changelogs)?;
+
+            self.sync_api_v3
+                .post_queued_records(self.site_id, self.central_server_site_id, &records)
+                .await?;
+
+            state.update_push_cursor(cursor + 1)?;
```
andreievg, line 195: + 1 ?

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -147,6 +167,74 @@ impl RemoteDataSynchroniser {
...
+    }
+}
+
+pub fn translate_changelogs_to_push_records(
+    connection: &StorageConnection,
+    changelogs: Vec<ChangelogRow>,
+) -> Result<Vec<RemotePostRecordV3>, anyhow::Error> {
+    let mut out_records: Vec<PushRecord> = Vec::new();
+    for changelog in changelogs {
+        translate_changelog(connection, &changelog, &mut out_records)?;
```
andreievg, line 212: I think i've seen you use it, and I've also started using it, iterators can return errors, I haven't figured out how to do it inline but can: 

let out_record_result: Result<Vec<..>, SyncTranslationError> = change_logs.into_iter().map(......).collect();
let our_records = out_record_result?;

server/src/sync/remote_data_synchroniser.rs
```diff
@@ -212,21 +300,38 @@ impl<'a> RemoteSyncState<'a> {
         self.key_value_store
             .set_bool(KeyValueType::RemoteSyncInitilisationFinished, Some(true))
     }
+
+    pub fn get_push_cursor(&self) -> Result<u32, RepositoryError> {
+        let value = self
+            .key_value_store
+            .get_i32(KeyValueType::RemoteSyncPushCursor)?;
```
andreievg, line 307: should it be get_u32 and set_u32 ?

server/src/sync/synchroniser.rs
```diff
@@ -63,6 +68,8 @@ impl Synchroniser {
         // check if there is new data on the central server
         self.central_data.pull(&connection).await?;
 
+        // TODO pull changes from remote server
```
andreievg, line 71: wasn't sure about the todo ?

Also we do push and then pull in existing sync

server/src/sync/translation_remote/number.rs
```diff
@@ -56,6 +58,52 @@ impl RemotePullTranslation for NumberTranslation {
...
+        } = NumberRowRepository::new(connection)
+            .find_one_by_id(&changelog.row_id)
+            .map_err(|err| to_push_translation_error(table_name, err.into(), changelog))?
+            .ok_or(to_push_translation_error(
+                table_name,
+                anyhow::Error::msg("Number row not found"),
+                changelog,
+            ))?;
+
+        let name = match make_number_name(&r#type, &store_id) {
```
andreievg, line 86: maybe from_number_name and to_number_name ?

server/src/sync/translation_remote/test_data/mod.rs
```diff
@@ -32,6 +36,18 @@ pub struct TestSyncRecord {
     pub remote_sync_buffer_row: RemoteSyncBufferRow,
 }
 
+/// To be used in combination with TestSyncRecord.
+/// I.e. first run and integrate a row from TestSyncRecord and then try to push this record out
+#[allow(dead_code)]
```
andreievg, line 41: i think test flag would do it

andreievg (approved):
Looks good, thanks

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: Its a bit of a special case for the initial pull because it looks at the ChangelogRepository to get the current cursor. For the normal case the cursor is fetched from the state store...

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: thanks fixed in [ae84e5b](https://github.com/openmsupply/remote-server/pull/843/commits/ae84e5b3573d2a513669b12357934fdcbcf7ce1d)

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: Think something like this would work:
```
let our_records = changelogs
        .into_iter()
        .map(|changelog| translate_changelog(connection, &changelog, &mut out_records))
        .collect::<Result<Vec<()>, SyncTranslationError>>()?;
```
However, I choose to base the out_records as a &mut to translate_changelog so that it can be filled there. This avoids some extra Vec allocations and refilling but mainly I found it easier...

server/src/sync/synchroniser.rs
clemens-msupply: Think TODO is resolved in next PR will note this point for our next meetings

server/src/sync/remote_data_synchroniser.rs
clemens-msupply: Not sure, key_value_store is quite close to the DB which is i32 and then I left the type cast to the state helper...

server/src/sync/translation_remote/number.rs
clemens-msupply: thanks done

server/src/sync/translation_remote/test_data/mod.rs
clemens-msupply: Yes I already added the mod as a test module so I think all these can be removed. Will do later.

Issue #699, opened 2022-01-24
Move graphql tests into graphql crate


clemens-msupply:
There is not real reason why the graphql tests need to be in the main server crate. Having them in the graphql crate should be a bit more natural.

andreievg:
- [ ] https://github.com/openmsupply/remote-server/pull/853
- [ ] https://github.com/openmsupply/remote-server/pull/854
- [ ] https://github.com/openmsupply/remote-server/pull/855
- [ ] https://github.com/openmsupply/remote-server/pull/856
- [ ] https://github.com/openmsupply/remote-server/pull/857
- [ ] https://github.com/openmsupply/remote-server/pull/858
- [ ] https://github.com/openmsupply/remote-server/pull/859
- [ ] https://github.com/openmsupply/remote-server/pull/860

andreievg:
Should be in dev now

PR #853, opened 2022-02-28
699 invoices relocate tests to graphql crate

andreievg:
https://github.com/openmsupply/remote-server/issues/699

* Relocated map_filter to core (out of scope of PR but was relocating test macros to core so thought to do it)
* Add `test_heplers` to core, and made run_test_gql_query runnable with different combination of Schema/Mutation
* Invoice crate now exports queries and mutations and root lib.rs uses them (this was needed for tests to live close to implementation)
* Adde setsup_graphql_test helper
* Still keeping graphql_client test commented out, will make an issue for that to be done seperately (but relocated)
* All invoice graphql tests are relocated


graphql/invoice/Cargo.toml
```diff
@@ -25,6 +25,10 @@ serde = "1.0.126"
 serde_json = "1.0.66"
 thiserror = "1.0.30"
 
+[dev-dependencies]
+actix-rt = "1.1.1" # for Tokio 0.2
+assert-json-diff = "2.0.1"
```
clemens-msupply, line 30: does this needs to be added to all crates? or can it be put into the root graphql crate later on?

clemens-msupply (approved):
Cool looks good!

Compiling and running the tests still takes about ~10s on my machine but checking the modified crate is much more responsive now.

graphql/invoice/Cargo.toml
andreievg: it's needed for tests in most graphql crates, i think it needs to be in all crates.

originally i wanted core to do import and re-export all deps, but had issue with macros, keen to revisit that idea. I actually wanted just one dep crate (project wide dep import and export), that all other can use. Be keen to chat about this next time

andreievg:
Will be merged in https://github.com/openmsupply/remote-server/pull/863

PR #860, opened 2022-02-28
699 requisition and line relocate tests to graphql crate

andreievg:
https://github.com/openmsupply/remote-server/issues/699

Most of it is as per other 699 PRs, but also:

* Removed all graphql tests from server
* Root lib.rs now lives in ./lib.rs rather than src
* Removed queries and mutations definitions from root


graphql/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2018"
 
 [lib]
-path = "src/lib.rs"
+path = "./lib.rs"
```
clemens-msupply, line 7: Is that a common Rust thing to not have it in src, do you know why?

clemens-msupply (approved):

graphql/Cargo.toml
clemens-msupply: I see, no leave it please

Issue #850, opened 2022-02-28
Use service for batch mutations


andreievg:
Also a way to ignore rollback (continueOnError)

- [x] https://github.com/openmsupply/remote-server/pull/861
- [x] https://github.com/openmsupply/remote-server/pull/862

PR #861, opened 2022-02-28
850 batch mutation as service for inbound shipment

andreievg:
https://github.com/openmsupply/remote-server/issues/850

* Added batch service for invoice
* Redone tests (for mapping just wanted to make sure all end points map correctly and are expose in gql, for service wanted to test rollback on/of)
* To use common mapping in individual gql's use `map_response`
* Added a way to roll back transaction with Ok result

Added `inline_init` which i am very happy with

### DIFF
--- a/repository/src/db_diesel/storage_connection.rs
+++ b/repository/src/db_diesel/storage_connection.rs
@@ -34,6 +34,11 @@ pub enum TransactionError<E> {
     Inner(E),
 }
 
+pub enum OkWithRollback<T> {
+    Ok(T),
+    OkWithRollback(T),
+}
+
 impl<E> TransactionError<E> {
     pub fn to_inner_error(self) -> E
     where
@@ -129,6 +134,18 @@ impl StorageConnection {
         self.transaction_sync_etc(f, true)
     }
 
+    /// Executes operations in transaction. A new transaction is only started if not already in a
+    /// transaction and can be manually rolled back
+    pub fn transaction_sync_with_rollback<'a, T, E, F>(
+        &'a self,
+        f: F,
+    ) -> Result<T, TransactionError<E>>
+    where
+        F: FnOnce(&'a StorageConnection) -> Result<OkWithRollback<T>, E>,
+    {
+        self.transaction_sync_etc_with_rollback(f, true)
+    }
+
     /// # Arguments
     /// * `reuse_tx` - if true and the connection is currently in a transaction no new nested
     /// transaction is started.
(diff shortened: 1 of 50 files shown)


repository/src/db_diesel/storage_connection.rs
```diff
@@ -139,19 +156,42 @@ impl StorageConnection {
...
+    /// * `reuse_tx` - if true and the connection is currently in a transaction no new nested
+    /// transaction is started.
+    pub fn transaction_sync_etc_with_rollback<'a, T, E, F>(
+        &'a self,
+        f: F,
+        reuse_tx: bool,
+    ) -> Result<T, TransactionError<E>>
+    where
+        F: FnOnce(&'a StorageConnection) -> Result<OkWithRollback<T>, E>,
     {
```
clemens-msupply, line 179: I think in general it would be good if transaction_sync_etc_with_rollback returns a special Ok Result type that indicates if there was a rollback or not, e.g. pseudo code:

`Result<struct { result: T, rollback: bool }, TransactionError<E>> `

repository/src/db_diesel/storage_connection.rs
```diff
@@ -173,6 +213,18 @@ impl StorageConnection {
...
+                    .map_err(|err| {
+                        error!("Failed to rollback tx: {:?}", err);
+                        TransactionError::Transaction {
+                            msg: "Failed to rollback tx".to_string(),
+                            level: current_level + 1,
+                        }
+                    })?;
+                Ok(value)
+            }
             Err(e) => {
```
clemens-msupply, line 228: This is where I implemented a similar logic:
`graphql/src/schema/mutations/stocktake/batch.rs`
but agree having a helper like this on the connection is nicer!

However, think it should be possible to reuse the existing StorageConnection methods similar to what I did. This should make it possible to have nested transaction_sync_etc_with_rollback within other transactions without needing to worry about transaction_levels...

repository/src/db_diesel/storage_connection.rs
```diff
@@ -34,6 +34,11 @@ pub enum TransactionError<E> {
     Inner(E),
 }
 
+pub enum OkWithRollback<T> {
+    Ok(T),
+    OkWithRollback(T),
+}
+
```
clemens-msupply, line 41: To avoid OkWithRollback::OkWithRollback(...), maybe something like

```rust
pub enum CancelableTxResult<T> {
   /// Transaction will be committed and Ok(T) is returned
   Ok(T),
   /// Transaction will be rolled backed and Ok(T) is returned
   Rollback(T),
}
```

util/src/inline_init.rs
```diff
@@ -0,0 +1,26 @@
...
+/// ```
+pub fn inline_init<T, F>(mut f: F) -> T
+where
+    T: Default,
+    F: FnMut(&mut T) -> (),
+{
+    let mut t = T::default();
+    f(&mut t);
+    t
+}
```
clemens-msupply, line 26: Nice, thought you implemented it with a macro like

```rust
inline_init!(Check, {
  id: "id1".to_string()
})
```
probably more advanced macro? since I don't like macros much I am fine to leave it as it is though.

clemens-msupply (approved):
Look good. See comment about rollback transaction.

repository/src/db_diesel/storage_connection.rs
```diff
@@ -161,9 +201,9 @@ impl StorageConnection {
         self.transaction_level.set(current_level + 1);
         let result = f(self);
         self.transaction_level.set(current_level);
-
+        println!("finished {}", current_level);
```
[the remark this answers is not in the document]
andreievg: Thanks, it's been removed in downstream PR somewhere

repository/src/db_diesel/storage_connection.rs
andreievg: I couldn't see how this can be easily done if transaction is called within the service. Similar to batch_stocktake, I think we can do it by returning result of error as an enum of either database error or the same type as ok (which I think would require some extra `froms` for this new error result. I think this might be tidier and i can just revert changes to transaction: d3fcfe0594fc0aba347d20c761e38b269ce37388

Thanks for suggestion imo look much nicer now

repository/src/db_diesel/storage_connection.rs
andreievg: reverted changes to storage conneciton completely: https://github.com/openmsupply/remote-server/commit/d3fcfe0594fc0aba347d20c761e38b269ce37388

andreievg:
Will be merged in https://github.com/openmsupply/remote-server/pull/863

PR #866, opened 2022-03-01
799 stocktake is locked

### LINKED ISSUE #799: Stocktake locked
Set stocktake as locked (not editable at all until unlocked). Can be added to update stocktake mutation.

andreievg:
closes #799 

Being still on the re-alignment vibe and wanting to improve future dev experience i decided to use inline_init in code that I touch (I will make a PR with all test data using inline_init), to make it possible without excessive time consumption made a .js script for it.

Probably best to view this by commit, last one being the actual locked changed.

Also I've added another helper inline_edit, simplifies new stocktake update tests


util/src/to_inline_init.js
```diff
@@ -0,0 +1,106 @@
...
+        if (!endsWithDefault) {
+            changedResult.push(changed_line);
+        }
+        first_line = false;
+    }
+
+    return { replacements: [firstLine], next_index: start_index };
+}
+// console.log(process().join('\n'))
+fs.writeFileSync(file_path, process().join('\n'))
```
clemens-msupply, line 106: Does it needs to be in the repo? isn't it just a one off script?

util/src/inline_init.rs
```diff
@@ -15,12 +17,51 @@
...
+/// );
+pub fn inline_edit<T, F>(record: &T, f: F) -> T
+where
+    T: Clone,
+    F: FnOnce(&mut T) -> (),
+{
+    let mut t = record.clone();
+    f(&mut t);
+    t
+}
```
clemens-msupply, line 59: nice

clemens-msupply (approved):

util/src/to_inline_init.js
andreievg: I've been using it quite a lot, would remove when it's not in use

PR #868, opened 2022-03-02
800 and 801 Stocktake dates

### LINKED ISSUE #800: Stocktake stocktakeDatetime
Add to stocktake insert and update mutation.

### LINKED ISSUE #801: Stocktake createDatetime
Should be set when stocktake is created like other end points. (not specified in parameter)

andreievg:
closes #800 
closes #801 

I think stocktake_date should be a date not datetime as described in the issue.

Also added some changes to test and made inline_edit a bit more ergonomic, previously, with references, it was hard to do defaults say for unwraping options with default values coming from the record itself)


graphql/stocktake/src/mutations/insert.rs
```diff
@@ -17,8 +17,8 @@ pub struct InsertStocktakeInput {
     pub id: String,
     pub comment: Option<String>,
     pub description: Option<String>,
-    pub created_datetime: NaiveDateTime,
     pub is_locked: Option<bool>,
+    pub stocktake_date: Option<NaiveDate>,
```
clemens-msupply, line 21: I couldn't find the motivation in the ticket? why not the more precise datetime? Isn't it a frontend matter in what resolution it should be recorded/displayed? Having the time in it seems to be useful to me, e.g. if there are two stocktakes on the same day...

repository/migrations/postgres/2021-12-15T10-00_stocktake_table/up.sql
```diff
@@ -11,6 +11,7 @@ CREATE TABLE stocktake (
     description TEXT,
     status stocktake_status NOT NULL,
     created_datetime TIMESTAMP NOT NULL,
+    stocktake_date DATE,
```
clemens-msupply, line 14: for future proof reasons, shouldn't it be a TIMESTAMP (even if the endpoints only use dates)

clemens-msupply (approved):
See comments about FE date vs BE datetime

repository/migrations/postgres/2021-12-15T10-00_stocktake_table/up.sql
andreievg: Can't do it for expiry, I think we've discussed this earlier, can talk about it in our next chat.

andreievg: I think we've discussed this earlier for expiry: If granularity is days and we use timestamp we need to set correct timestamp (based on server time ?), it's much simpler to just use date

graphql/stocktake/src/mutations/insert.rs
andreievg: I thought it was just the date, i assumed the input on f/e would just be date, added to things to talk about with f/e: https://github.com/openmsupply/remote-server/issues/879

PR #873, opened 2022-03-02
872 Change insert service line to allow for default service item

### LINKED ISSUE #872: Default service item
Insert service line mutation to allow for default service item

andreievg:
closes #872

### DIFF
--- a/service/src/invoice_line/outbound_shipment_service_line/insert/generate.rs
+++ b/service/src/invoice_line/outbound_shipment_service_line/insert/generate.rs
@@ -4,19 +4,28 @@ use util::inline_init;
 use super::{InsertOutboundShipmentServiceLine, InsertOutboundShipmentServiceLineError};
 
 pub fn generate(
-    input: InsertOutboundShipmentServiceLine,
+    InsertOutboundShipmentServiceLine {
+        id,
+        invoice_id,
+        item_id: _,
+        name,
+        total_before_tax,
+        total_after_tax,
+        tax,
+        note,
+    }: InsertOutboundShipmentServiceLine,
     item: ItemRow,
 ) -> Result<InvoiceLineRow, InsertOutboundShipmentServiceLineError> {
     let new_line = inline_init(|r: &mut InvoiceLineRow| {
-        r.id = input.id;
-        r.invoice_id = input.invoice_id;
+        r.id = id;
+        r.invoice_id = invoice_id;
         r.item_id = item.id;
         r.item_code = item.code;
-        r.item_name = input.name.unwrap_or(item.name);
-        r.total_before_tax = input.total_before_tax;
-        r.total_after_tax = input.total_after_tax;
-        r.tax = input.tax;
-        r.note = input.note;
+        r.item_name = name.unwrap_or(item.name);
+        r.total_before_tax = total_before_tax;
+        r.total_after_tax = total_after_tax;
+        r.tax = tax;
+        r.note = note;
         r.r#type = InvoiceLineRowType::Service;
     });
 
--- a/service/src/invoice_line/outbound_shipment_service_line/insert/generate.rs
+++ b/service/src/invoice_line/outbound_shipment_service_line/insert/generate.rs
@@ -1,4 +1,5 @@
 use repository::schema::{InvoiceLineRow, InvoiceLineRowType, ItemRow};
+use util::inline_init;
 
 use super::{InsertOutboundShipmentServiceLine, InsertOutboundShipmentServiceLineError};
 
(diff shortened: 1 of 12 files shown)


service/src/invoice_line/outbound_shipment_service_line/insert/generate.rs
```diff
@@ -1,34 +1,24 @@
...
+        r.invoice_id = input.invoice_id;
+        r.item_id = item.id;
+        r.item_code = item.code;
+        r.item_name = input.name.unwrap_or(item.name);
+        r.total_before_tax = input.total_before_tax;
+        r.total_after_tax = input.total_after_tax;
+        r.tax = input.tax;
+        r.note = input.note;
+        r.r#type = InvoiceLineRowType::Service;
+    });
```
clemens-msupply, line 21: Sorry probably jumped over it before. Do we want to use inline_init() here as well? Saw it to be mainly useful for test data.

The previous version was nice because it showed the exact line that is generated without hiding anything. At least de-structure the input structs?

util/src/constants.rs
```diff
@@ -2,3 +2,5 @@
 pub const INVENTORY_ADJUSTMENT_NAME_CODE: &str = "invad";
 /// Number of days in a month (used in AMC calculation)
 pub const NUMBER_OF_DAYS_IN_A_MONTH: f64 = 30.0;
+/// For use when service item is not specified in service invoice line
+pub const DEFAULT_SERVICE_ITEM_CODE: &str = "service";
```
clemens-msupply, line 6: Would it make sense to prefix default items, e.g. `omsupply_internal#service` or `___service___`

clemens-msupply (approved):

service/src/invoice_line/outbound_shipment_service_line/insert/generate.rs
andreievg: > Saw it to be mainly useful for test data

Hmm i thought it's nice as a general utility method. Like the other variant inline_edit i think is useful for update mutations. Didn't really need to use it for inserts, but I thought it's simpler and requires less changes when new field is added (I guess in case of insert it's likely newly added field would be part of insert mutation, but i still think it's nicer as long as we destructure)

> The previous version was nice because it showed the exact line that is generated without hiding anything

Would be great to clarify again in which cases you think abstraction can be used, and the differences between indirection and abstraction, I thought that abstraction and indirection is fine if it's a common used thing, like `from` for RepositoryError. Anyways wanted to ask this because even though inline_init is new, I see it potentially being used quite a lot.

> At least de-structure the input structs

For sure, thanks (sorry, this could have lead to hard to trace bugs down the line, if tests were also not added): 7ebf8287a8e4c0503dd3bb482abe123206673312

util/src/constants.rs
andreievg: Hmm yeah, although would need to do also in translations, added issue: https://github.com/openmsupply/remote-server/issues/880

andreievg:
Merging but made a note to talk about `inline_int` and `inline_edit` use in our next meeting

service/src/invoice_line/outbound_shipment_service_line/insert/generate.rs
clemens-msupply: Indirection usually (always?) comes when using some abstraction. I would say the abstraction has to outweigh possible disadvantages of indirection, e.g. more jumping around to read the code. However, if devs are familiar with a certain abstraction, code could even become easier to read and understand, e.g. inline_init reduces the boilerplate in tests and  you can focus more on what you want test.

For this shipment_line insert/update I think its different because these are the main mutation methods and reducing boileplate is not really needed here. Having the full insert/update struct in one place without needing to lookup what else is implicitly inserted/updated makes it IMHO easier to understand. For example, the old code had some comments on why some fields are not updated, this information is missing now and a new dev might get confused about it. Its also easier to verify if there is a default value that should be updated but is accidentally ignored.

andreievg: Ok, i can see a benefit of seeing full record with default values for insert mutations, not for updates thought (values are coming from existing record, so no need to look up any defaults), I've noted it down, will make an issue and revert, thanks for pointing it out.

Hmm i don't think abstraction = indirection, i think indirection could be a result of complicated abstraction. If abstraction = indirection then machine code is the only 'direct' code, which no one can read thus abstraction layers are added in a more human readable/understandable code.

> more jumping around to read the code

I would argue this, we don't jump around to `println!` implementation or other standard library implementations, we trust it works, similar to things like to_domain, there is not need to jump to it when reading the context in which it's called. I think the only time you need to jump to implementation is if the abstraction is named incorrectly, or when it has side effects or indirection in form of error conversions with `?` etc (like we've had in original service validation, which I agree now, it's bad)

PR #889, opened 2022-03-06
714 push fixes

clemens-msupply:


andreievg (approved):
Looks good, I still think we still want to disable push ?

Btw for `full_circle` test I was thinking:

* Full circle test are under a special flag (they don't run with normal tests)
* Start with a base data file (can commit it to repo)
* Load that data file to 4d and start server
* Test initilises, then inserts some new records
* Then tries to initilise (should be able to craete a new db and new connection and sync actor with existing helper method)

clemens-msupply:
thanks, push is already disabled in develop...

PR #900, opened 2022-03-08
886 printing endpoint stub

clemens-msupply:
Adds:
- printReport graphql endpoint that only takes a report id and the id of the entity to be printed, e.g. the invoice id, and returns a file id
- /files?id={id} endpoint to fetch the printed report

The files endpoint already contains some proof of concept logic which needs to be update and put into a service:
- It always creates a dummy txt file for a given id in the static_files directory
- Returns this dummy txt file with a proper file name (not the file id). The "proper" filename is currently encode like `{file_id}_filename.pdf`
- Cleans up old static files that are too old

#886

### DIFF
--- a/graphql/printing/src/lib.rs
+++ b/graphql/printing/src/lib.rs
@@ -1,0 +1,69 @@
+use async_graphql::*;
+use graphql_core::simple_generic_errors::RecordNotFound;
+
+pub struct InvalidReport;
+#[Object]
+impl InvalidReport {
+    pub async fn description(&self) -> &'static str {
+        "Report exist but is invalid"
+    }
+}
+
+pub struct FailedToFetchReportData;
+#[Object]
+impl FailedToFetchReportData {
+    pub async fn description(&self) -> &'static str {
+        "Failed to query data required for the report"
+    }
+}
+
+#[derive(Interface)]
+#[graphql(field(name = "description", type = "String"))]
+pub enum PrintReportErrorInterface {
+    /// No report found with the specified report id
+    ReportNotFound(RecordNotFound),
+    InvalidReport(InvalidReport),
+    FailedToFetchReportData(FailedToFetchReportData),
+}
+
+#[derive(SimpleObject)]
+pub struct PrintReportError {
+    pub error: PrintReportErrorInterface,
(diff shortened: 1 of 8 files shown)


graphql/printing/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[derive(PartialEq, Debug)]
+pub struct PrintReportNode {}
+
+#[Object]
+impl PrintReportNode {
+    /// Return the file id of the printed report.
+    /// The file can be fetched using the /files?id={id} endpoint
+    pub async fn file_id(&self) -> &str {
+        "demofile0123456789"
```
andreievg, line 42: would it be be easier for f/e and easier to change, to have the full url (excluding host) ?

server/src/static_files.rs
```diff
@@ -0,0 +1,115 @@
...
+        "Static file not found",
+    ))?;
+    let original_file_name = parse_original_file_name(&query.id, &file_path).ok_or(
+        InternalError::new("Invalid file name", StatusCode::INTERNAL_SERVER_ERROR),
+    )?;
+    // clean up the static file directory
+    let max_lifetime_sec = 60 * 60; // 1 hours
+    delete_old_files(file_dir, max_lifetime_sec)?;
+
+    fs::NamedFile::open(file_path)?
```
andreievg, line 47: Does this pipe into a response ? 

If not, was wondering if we should just delete the file straight after it's been put into a response

andreievg (approved):
Looks good thanks

server/src/static_files.rs
clemens-msupply: Just had a look, it does stream the content directly from the file so can't delete immediately...

graphql/printing/src/lib.rs
clemens-msupply: My first version was actually:

``` rust
pub async fn file_path(&self) -> &str {
        "/files?id=demofile0123456789&name=test.pdf"
}
```

but I found it a bit odd to use. Returning the whole url including the host would be nicer but difficult to implement. So I decided to only return an id.

PR #902, opened 2022-03-09
Improve/fix total_before_tax and total_after_tax calculation

clemens-msupply:
I had a chat with Craig the other day and he basically said what you said before.


andreievg (approved):
Looks great, thanks

PR #905, opened 2022-03-09
904 add user id to requistion, invoice and stocktake

### LINKED ISSUE #904: Add user_id to requisition, invoice, stocktake

andreievg:
closes #904 

Made user_id optional with non enforced reference for now, as per telegram convo (users may not exist).

A lot of changes with init_inline being used.

As per https://github.com/sussol/msupply/blob/ac39d958d89b845af60fcd2c9c4929105a16b67a/Project/Sources/Methods/transaction_create_linked_SI.4dm#L141, transfer record generated in central dont' have user associated until they are edited, thus made in 'None' in sync processor

Also i've revered using inline_init in insert mutations (i think all of them)

I ran cargo fix and fmt at some point, is causing some extra changes sorry


server/src/sync/translation_remote/invoice.rs
```diff
@@ -90,7 +90,9 @@ pub struct LegacyTransactRow {
     #[serde(rename = "type")]
     pub _type: LegacyTransactType,
     pub status: LegacyTransactStatus,
-
+    #[serde(deserialize_with = "empty_str_as_option")]
+    #[serde(rename = "user_ID")]
+    pub user_id: Option<String>,
```
clemens-msupply, line 95: agree it gives a nice summary in one place for how fields are mapped. There are some non-obvious mappings and having the rename here makes it more fail-proof.

In general I would say its good to keep the struct close to the underlying data but in this case the struct is really just used for sync mapping and having mapping straight there is better.

clemens-msupply (approved):
Somewhat related: there was an article to this week regarding postgres audit logs:
https://supabase.com/blog/2022/03/08/audit

However, this wouldn't easily work with sqlite since it can't just serialize a row to json. But if we find a similar solution we wouldn't need to store the user information in the tables but could have a general edit log...

server/src/sync/translation_remote/invoice.rs
andreievg: Yeah i was thinking it's in_between mapper between simple and complex, and was hoping serde can do as much as possible without obfuscating in indirection (with rename and the simple helper methods i think achieve)

PR #908, opened 2022-03-09
901 PricingNode

### LINKED ISSUE #901: Missing translation fields in mSupply
Below is a list of required column at the central server DB to backup new omSupply columns.

Columns annotated with Null means they are optional in omSupply (don't know if it makes a different in mSupply)
NaiveDateTime: is a UTC date+time variable. omSupply would send it as an iso string (e.g. "2011-10-05T14:48:00.000Z") in the push json object.

## trans_line:
- [ ] om_item_code: String
- [ ] om_tax: Null Float
- [ ] om_total_before_tax: Float
- [ ] om_total_after_tax: Float

## transact:
- [x] ~~om_name_store_id: Null String~~ (can actually be deduced)
- [ ] om_created_datetime: NaiveDateTime (maps to entry date and time)
- [ ] om_allocated_datetime: Null NaiveDateTime
- [ ] om_picked_datetime: Null NaiveDateTime
- [ ] om_shipped_datetime: Null NaiveDateTime
- [ ] om_delivered_datetime: Null NaiveDateTime
- [ ] om_verified_datetime: Null NaiveDateTime
- [ ] om_status field: String (currently not a one to one mapping)
- [ ] om_type field: String (inventory adjustment is missing) OR add a type inventory adjustment to mSupply
- [ ] om_colour: String

## requisition:
- [ ] om_created_datetime: NaiveDateTime
- [ ] om_sent_datetime:  Null NaiveDateTime
- [ ] om_finalised_datetime: Null NaiveDateTime
- [ ] om_max_months_of_stock: Float (exiting daysToSupply integer value might loose precision if fractions of a month are specified)
- [ ] om_status; String (not a one to one mapping)
- [ ] om_colour: String

## requisition_line:
- [x] ~~ om_average_monthly_consumption: Float~~ use daily_usage

## stocktake:
- [ ] om_created_datetime: NaiveDateTime
- [ ] om_finalised_datetime:  Null NaiveDateTime

## stocktake_line:
- [ ] om_note: Null String

## name_store_join:
- [ ] om_name_is_customer: bool
- [ ] om_name_is_supplier: bool

andreievg:
closes #901 

Changes as pre commit description:

* Rename InvoicePricingNode to PricingNode
* Rename InvoiceStatsRow to PricingRow
* Add avg tax to PricingRow
* Add taxPercentage to InvoicePricingNode
* Add test for InvoiceLineNode (including pricing)

I like the test of types in the type file (including testing the loaders)


repository/migrations/sqlite/2021-10-05T10-00_create_invoice_stats_view/up.sql
```diff
@@ -1,8 +1,9 @@
 CREATE VIEW invoice_stats AS
 SELECT
 	invoice_line.invoice_id,
-  SUM(invoice_line.total_before_tax) AS total_before_tax,
+    SUM(invoice_line.total_before_tax) AS total_before_tax,
 	SUM(invoice_line.total_after_tax) AS total_after_tax,
+	AVG(invoice_line.tax) AS tax_percentage,
```
clemens-msupply, line 6: I might have missed the discussion but is the avg of a tax meaningful? for example 10 lines with tax 20 and 1 line with tax 10 and an average tax of 15 doesn't look right

graphql/types/src/types/invoice_line.rs
```diff
@@ -198,3 +203,229 @@ impl InvoiceLineNode {
...
+                        location_row_option: Some(inline_init(|r: &mut LocationRow| {
+                            r.name = "line_location_name".to_string();
+                        })),
+                    },
+                }
+            }
+        }
+
+        let expected = json!({
+            "testQuery": {
```
clemens-msupply, line 264: Yes nice!

clemens-msupply (approved):
Please see comment about avg tax

repository/migrations/sqlite/2021-10-05T10-00_create_invoice_stats_view/up.sql
joshxg: No - effective tax rate is though. Thinking to show in the UI the effective tax rate for stock, service and maybe the full invoice.

andreievg: I've added test to invoice pricing: ccff5a2e9608ce714b330674810d8507daacd46b

>  for example 10 lines with tax 20 and 1 line with tax 10 and an average tax of 15 doesn't look right

This would be 10.9, sum / count, i think test confirms this.

> I might have missed the discussion but is the avg of a tax meaningful?

I think total tax for the invoice is useful, but not 100% about it, wanted to use common type, thus did the calculation in view.

> effective tax rate is though.

What would be the difference between effective and avg accross lines?

>  Thinking to show in the UI the effective tax rate for stock, service and maybe the full invoice

So the tax should also be broken down by type ?

(also added ability to initilise mock data on test start up, together with inline init makes it easy to add some extra data for the test, I was pushing against that at the start, but it seems very easy now and for tests like this I think it's usefull)

(also made invoice_stats_view filter by invoice_line.type rather then item.type)

andreievg:
Merging, will keep an eye on further comments

repository/migrations/sqlite/2021-10-05T10-00_create_invoice_stats_view/up.sql
joshxg: Either a couple of late night deadline mistakes here - or long day mistakes about to happen from me 😁 

> What would be the difference between effective and avg accross lines?

Line1: tax 10%, subtotal: $100, total: $110
line2: tax 0%, subtotal: $50, total $50
line3: tax 5%, subtotal: $100, total $105

subtotal: 250
total: 265

average: 5%  (sum of tax percentages / number of lines - 15 / 3) - 250*1.05 = 262.5 ❌ 
effective: 6%  (total amount of tax paid / total amount on invoice  - (265-250) / 265) -  250*1.06 = 265 ✅ 

> So the tax should also be broken down by type ?

Happy to just calculate it in the UI - but the tax percentage right now is actually probably harmful if anything

andreievg: Thank you, yeah my bad, just didn't really think about it.

PR #912, opened 2022-03-10
757 stocktake batch mutation as service

### LINKED ISSUE #757: Create stocktake batch mutations service
Move batch implementation from graphql to service.

Other changes:
- Give standard error details of which operation failed
- Add multiple stocktake inserts, updates, deletes to the batch

andreievg:
Closes #757 

Sorry for so many changes, it's mainly all just mapping. I did this PR now since i wanted to do a common pattern for passing user_id to mutations (which also includes stocktake batch mutations), and seems like front end needs to have a way to add a bunch of stock lines even when those stock lines are present (continue_on_error)

* Started of with simplifying request requisition batch, i wanted to copy it to use for stocktake batch [Simplify batch request requisition](https://github.com/openmsupply/remote-server/pull/912/commits/e25dedb50415795f34d928df6d91ba909f54921a)
* After copying request requisition batch and doing a global replace, realised that I need to align stocktake graphql and service with pattern that I used, which ended up being quite a big job, sorry I didn't want to change the style with which stocktake was implemented, but ended up changing so that stocktake batch becomes easy to do
* Added overall test just incase i've introduced conflicting graphql types


clemens-msupply (approved):
Looks good to me

PR #914, opened 2022-03-10
913 Associate user with records

### LINKED ISSUE #913: Associate users with records on edit/create

andreievg:
closes #913 

* Change from EditRequistion to MutateRequisition resource
* Added MutateInboundShipment and OutboundShipment resource
* Had to add do_mutation_with_user_id to batch mutations (I think we should have just one common parameter to services and store_id and user_id should be within ServiceContext (quite a lot of changes otherwise if we want to add a common parameter to services)

Outbound Shipment, Request Requisition and Stocktake are only creatable by user (not auto creating), so just added user on insert.

Inbound Shipment and Response Requisition however can be created by system, so need to add user update on all of the mutations (didn't include create requisition shipment, since quantity needs to be set on requisition via other mutations that set user id, before creating requisition shipment)


clemens-msupply (approved):
Looks good.

I think ServiceContext is too general to store user_id and store_id. For example, there could be requests without user and without store and then you have to store them as optional. But we could have another "context" object that contains the user and store id? or something like:
```rust 
struct UserRequest {
   service_context: ServiceContext,
   store_id: String,
   user: User,
}
```

and pass this one around where applicable

andreievg:
Hmm, I was thinking about this, could also just make the whole service input a struct.

UserRequest sounds pretty good though

PR #915, opened 2022-03-11
Add support to sync inventory adjustments

clemens-msupply:
Some of this will be refactored once we have missing fields in mSupply.

Not sure about the invad status mapping (was just guessing but as far as I remembered they where only new and finalised?)


andreievg (approved):
Not sure about the invad status mapping (was just guessing but as far as I remembered they where only new and finalised?)

Yeah i think it's just finalised for invad

PR #927, opened 2022-03-13
714 pull fixes

clemens-msupply:
Just some smaller fixes:
- return when remote pull translation succeeded (don't print warning that it didn't)
- some API fixes for insert and delete records


andreievg (approved):
thanks

PR #934, opened 2022-03-15
591 outbound shipment transport reference

### LINKED ISSUE #591: Outbound shipment: transport reference
Not sure which bucket you'd like to put this - an entirely new free text field for outbound shipments, which is to store `Transport Reference`

We've been asked to add alongside the `Shipping Method` selection.
This would require
- new field adding to the `invoice` table ( though only applicable for outbound - how do we handle type-specific fields 🤔  )
- accept a value as part of the update mutation
- return the value as part of the invoice query

andreievg:
closes #591 

Ended up adding a test to invoice translation, tests were failing and it was really hard to figure out why.

Also added test to outbound shipment service (rather then adding to existing graphql test), i made it so that. new fields added to update service would need to be added to the test for compilation.

Also added destructuring to outbound shipment update


server/src/sync/translation_remote/invoice.rs
```diff
@@ -432,3 +439,28 @@ fn legacy_invoice_status(
...
+
+        let translator = InvoiceTranslation {};
+        for record in get_test_transact_records() {
+            let translation_result = translator
+                .try_translate_pull(&connection, &record.remote_sync_buffer_row)
+                .unwrap();
+
+            assert_eq!(translation_result, record.translated_record);
+        }
+    }
```
clemens-msupply, line 465: Yes I noticed this as well that it gets sometimes tricky to find out what when wrong in a test. Since I added translation on the fly it wasn't a big issue but when doing unrelated changes it becomes more difficult. I can add fine grain tests like this for all the translations.

clemens-msupply (approved):
Looks good to me

server/src/sync/translation_remote/invoice.rs
andreievg: I think we just add them on the go, when needed

PR #946, opened 2022-03-17
Add ignore_user_in_auth setting

### LINKED ISSUE #945: Add setting for enabling user validation in auth
Looks like `debug_no_access_control` is set to true with a TODO to enable it when f/e supports auth, this mean we always consider user as dummy user. 

Add server setting `ingore_user_in_auth`, this will be the value for debug_no_access_control

andreievg:
closes #945


clemens-msupply (commented):
This only disables access control, right? or is there another use for it? I think the naming could be better.

I used `debug_` but in reqwest I have seen a `danger_` prefix for dangerous options which I liked as well. This option is just for dev isn't it?

andreievg:
> This only disables access control, right?

Yeah, that what it looks like, i think `debug_no_access_control` could be the name, can use that in the yaml, comment it out, and have a default in settings = false

andreievg:
changes as per last comment c5f280343fe8e3d6c83857243bec138b30ece1ae

Issue #951, opened 2022-03-18
Add allocate line to batch mutation


andreievg:
- [x] Align existing outbound/inbound shipment batch mutations with latest requisition and stocktake batch mutations [PR](https://github.com/openmsupply/remote-server/pull/952)
- [x] Add allocate line mutation to batch mutations [PR](https://github.com/openmsupply/remote-server/pull/953)

PR #952, opened 2022-03-18
951 1 Allign existing batch mutations

andreievg:
Fixes buletpoint 1 in #951 

Wanted to do this before adding new mutation to batch outbound shipment, to simplify future tasks of adding to batch mutations (and to simplify batch mutations)


clemens-msupply (approved):
Looks good to me

PR #959, opened 2022-03-21
958 requisition mock chart changes

### LINKED ISSUE #958: Order by date, isHistoric in evolution and max and min
Update mock requisition charts as per: https://github.com/openmsupply/remote-server/issues/809#issuecomment-1073484472
* order nodes by date (asc)
* add isHistoric to consumptionHistory
* add isHistoric to stockEvolution (with single stockOnHand value)
* add mix and max to stockEvolution nodes

andreievg:
closes #958

### DIFF
--- a/graphql/types/src/types/item_chart_data.rs
+++ b/graphql/types/src/types/item_chart_data.rs
@@ -5,6 +5,7 @@ use chrono::NaiveDate;
 pub struct ConsumptionHistoryNode {
     pub consumption: u32,
     pub amc: f64,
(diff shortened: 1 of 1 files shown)


clemens-msupply (approved):

mark-prins:
@andreievg  have just noticed that for the 'StockEvolution' data, the `isHistoric` is true for the later dates, and false for earlier dates. Probs no need to revisit the mock data, but worth checking when the real data comes through.

Issue #919, opened 2022-03-11
Translation issues


andreievg:
Ended up doing a PR: 

I went as far as:
* Making LegacyStockLineRow available and quantity f64, and casting with `as` to underlying data type
* Adding rename for invoice type 'Sr' for 'Sr', and new variant Sr2 for 'sr'
* Decided to just gracefully fail translation for now

I think in dev stages we should ignore records that are not translating, log them though and encourage remote server users to provide us with a log file (and data file)

To investigate can use data file (thanks @Josh):
[Harry_potter_demo_data_20181212.zip](https://github.com/openmsupply/remote-server/files/8231356/Harry_potter_demo_data_20181212.zip)
Or provided sqlite file (thanks @Josh), as I did and do this change then cargo run
[omsupply-database.sqlite.zip](https://github.com/openmsupply/remote-server/files/8231368/omsupply-database.sqlite.zip)

PR #968, opened 2022-03-23
961 remote sync missing fields

### LINKED ISSUE #961: Remote Sync: make use of new om_* fields
## trans_line:
- [x] om_item_code: String
- [x] om_tax: Null Float
- [x] om_total_before_tax: Float
- [x] om_total_after_tax: Float

## transact:
- [x] ~~om_name_store_id: Null String~~ (can actually be deduced)
- [x] om_created_datetime: NaiveDateTime (maps to entry date and time)
- [x] om_allocated_datetime: Null NaiveDateTime
- [x] om_picked_datetime: Null NaiveDateTime
- [x] om_shipped_datetime: Null NaiveDateTime
- [x] om_delivered_datetime: Null NaiveDateTime
- [x] om_verified_datetime: Null NaiveDateTime
- [x] om_status field: String (currently not a one to one mapping)
- [x] om_type field: String (inventory adjustment is missing) OR add a type inventory adjustment to mSupply
- [x] om_colour: String

## requisition:
- [x] om_created_datetime: NaiveDateTime
- [x] om_sent_datetime:  Null NaiveDateTime
- [x] om_finalised_datetime: Null NaiveDateTime
- [x] om_expected_delivery_datetime
- [x] om_max_months_of_stock: Float (exiting daysToSupply integer value might loose precision if fractions of a month are specified)
- [x] om_status; String (not a one to one mapping)
- [x] om_colour: String

## requisition_line:
- [x] ~~ om_average_monthly_consumption: Float~~ use daily_usage
- [x] om_calculated_datetime

## stocktake:
- [x] om_created_datetime: NaiveDateTime
- [x] om_finalised_datetime:  Null NaiveDateTime

## stocktake_line:
- [x] om_note: Null String

## name_store_join:
- [x] om_name_is_customer: bool
- [x] om_name_is_supplier: bool

#714

clemens-msupply:
- Make all integration tests pass
- Enables pushing remote data
- For invoice line a None tax value is synced as -1 to the central server since the central server doesn't seem to have null but sends back an empty string or 0. For tax there a difference between None and 0 though.

Closes #961


andreievg:
> For invoice line a None tax value is synced as -1 to the central server since the central server doesn't seem to have null but sends back an empty string or 0. 

I am just checking this now, all om_fields should be nullable, and central server does send a null value, was the problem in sending nulls to the central server ?

server/src/sync/translation_remote/invoice_line.rs
```diff
@@ -80,45 +91,65 @@ impl RemotePullTranslation for InvoiceLineTranslation {
...
                 cost_price_per_pack: data.cost_price,
                 sell_price_per_pack: data.sell_price,
-                total_before_tax: total,
-                total_after_tax: total,
-                tax: None,
+                total_before_tax,
+                total_after_tax,
+                tax,
                 r#type: line_type,
                 number_of_packs: data.quantity / data.pack_size,
```
andreievg, line 154: quantity is number_of_packs in mSupply (don't need to multiply by pack_size)

server/src/sync/translation_remote/invoice_line.rs
```diff
@@ -80,45 +91,65 @@ impl RemotePullTranslation for InvoiceLineTranslation {
...
-                    table_name,
-                    source: anyhow::Error::msg(format!("Failed to get item: {}", data.item_ID)),
-                    record: sync_record.data.clone(),
-                })
-            }
-        };
         let line_type = to_invoice_line_type(&data._type).ok_or(SyncTranslationError {
             table_name,
             source: anyhow::Error::msg(format!("Unsupported trans_line type: {:?}", data._type)),
             record: sync_record.data.clone(),
```
andreievg, line 97: :nit 
we could simplify error return by just returning the source, the caller of try_translate_pull can add table_name and sync_record info, i think it would simplify translations slightly

server/src/sync/translation_remote/invoice.rs
```diff
@@ -282,7 +409,6 @@ fn invoice_status(
             let delivered = data.arrival_date_actual.is_some();
             match data.status {
                 LegacyTransactStatus::Nw => InvoiceRowStatus::New,
-                // TODO could also mean Allocated
                 LegacyTransactStatus::Sg => InvoiceRowStatus::New,
```
andreievg, line 412: I think we should do a very basic translation from legacy rows:

if om_status then concrete status
else

if Outbound

sg -> New
cn -> Picked
fn -> Shipped

if inbound

nw -> New
cn -> Delivered
fn -> Verified

this will simplify and protect against edge case conditions, don't really want to think about them all but one is:
* Picked Inbound shipment is not editable in API, if legacy row has ship_date is not specified (it's optional in legacy), then record would be not editable

server/src/sync/translation_remote/invoice.rs
```diff
@@ -225,34 +326,60 @@ fn invoice_type(_type: &LegacyTransactType, name: &NameRow) -> Option<InvoiceRow
...
+    let mut mapping = LegacyMapping {
+        created_datetime: NaiveDateTime::new(data.entry_date, data.entry_time),
+        picked_datetime: None,
+        delivered_datetime: None,
+        allocated_datetime: None,
+        shipped_datetime: data
+            .ship_date
+            .map(|ship_date| date_and_time_to_datatime(ship_date, 0)),
+        verified_datetime: None,
+        colour: Some(format!("#{:06X}", data.Colour)),
```
andreievg, line 362: I don't think we need to convert colour, I think we shoud just use om field

server/src/sync/translation_remote/invoice.rs
```diff
@@ -225,34 +326,60 @@ fn invoice_type(_type: &LegacyTransactType, name: &NameRow) -> Option<InvoiceRow
...
+    verified_datetime: Option<NaiveDateTime>,
+    colour: Option<String>,
 }
-fn map_legacy_confirm_time(
-    invoice_type: &InvoiceRowType,
-    data: &LegacyTransactRow,
-) -> ConfirmTimeMapping {
+/// Either make use of om_* fields, if present, or do a best afford mapping
+fn map_legacy(invoice_type: &InvoiceRowType, data: &LegacyTransactRow) -> LegacyMapping {
+    if let Some(created_datetime) = data.created_datetime {
```
andreievg, line 341: Would be great to have a comment, something like if created_datetime exists then the record was created in omSupply, thus can use omsupply fields

server/src/sync/translation_remote/invoice.rs
```diff
@@ -225,34 +326,60 @@ fn invoice_type(_type: &LegacyTransactType, name: &NameRow) -> Option<InvoiceRow
...
+        };
+    }
+
+    let mut mapping = LegacyMapping {
+        created_datetime: NaiveDateTime::new(data.entry_date, data.entry_time),
+        picked_datetime: None,
+        delivered_datetime: None,
+        allocated_datetime: None,
+        shipped_datetime: data
+            .ship_date
```
andreievg, line 359: hmm didn't really conider using ship_date from legacy row, not sure what affect it would have (i.e. is it possible to have ship date set but record not finalised/shipped), maybe it's best to avoid mapping from that field for now (can map to it in push though)

server/src/sync/translation_remote/invoice.rs
```diff
@@ -225,34 +326,60 @@ fn invoice_type(_type: &LegacyTransactType, name: &NameRow) -> Option<InvoiceRow
...
+            .map(|ship_date| date_and_time_to_datatime(ship_date, 0)),
+        verified_datetime: None,
+        colour: Some(format!("#{:06X}", data.Colour)),
+    };
+
     let confirm_datetime = data
```
andreievg, line 365: Since we onlyl really have two dates that are guaranteed to be set in legacy system (entry and confirm), we would need to populate all relevant om supply date/time fields based on current status.

i.e. if it's a conversion from legacy row:

Outbound

om.created = legacy.entry

cn -> om.allocate && om.picked = legacy.confirm
fn -> om.allocated && om.picked && om.shipped = legacy.confirm

Inbound

om.created = legacy.entry

cn -> om.delivered = legacy.confirm
fn -> om.delivered && om.verified = legacy.confirm

Inventory Adjustment

nw -> Ideally we just ingore this record, otherwise om.created = legacy.entry
cn -> om.verified = legacy.confirm (as per status translation comment)
fn -> om.verified = legacy.confirm

server/src/sync/translation_remote/invoice.rs
```diff
@@ -225,34 +326,60 @@ fn invoice_type(_type: &LegacyTransactType, name: &NameRow) -> Option<InvoiceRow
...
+        InvoiceRowType::InventoryAdjustment => {
+            mapping.delivered_datetime = confirm_datetime;
+        }
+    };
+    mapping
 }
+
 fn to_legacy_confirm_time(
     invoice_type: &InvoiceRowType,
     picked_datetime: Option<NaiveDateTime>,
```
andreievg, line 385: Can't quite comment on below, do we ever set delivered_datetime on Inventory Adjustment (i thought it was only verified ?)
Also I think time can also be an optional result ?

server/src/sync/translation_remote/invoice.rs
```diff
@@ -79,6 +79,65 @@ pub enum TransactMode {
     Dispensary,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum InvoiceStatus {
```
andreievg, line 84: Why do we need this one ?

andreievg:
"request_requisition" => NumberRowType::RequestRequisition,
"response_requisition" => NumberRowType::ResponseRequisition,

I am not really sure what to do with this one, I remember making a comment about this, but not sure where (either a pr or telegram). Anyways, mSupply only has one type for both, and mobile has completely different types for all number rows. mSupply it's just `requisition`, would be great if we can have seperate ones in omSupply, but translations become a bit tricky, i think we should just use combined type in v1. Beyond v1, we can do:

* requisition maps to -> response_requisition
* a query is made when requisition type is received to see if there is request_requisition type already, if not then add request_requisition type being = number of requisition number being received, otherwise ignore

server/src/sync/translation_remote/requisition_line.rs
```diff
@@ -34,6 +35,11 @@ pub struct LegacyRequisitionLineRow {
```
andreievg, line 35: do we still need om_average_monthly_consumption ? I think we should be ok with daily_usage f64

server/src/sync/translation_remote/requisition_line.rs
```diff
@@ -34,6 +35,11 @@ pub struct LegacyRequisitionLineRow {
 
     #[serde(deserialize_with = "empty_str_as_option")]
     pub comment: Option<String>,
+
+    #[serde(rename = "om_snapshot_datetime")]
+    #[serde(default)]
+    #[serde(deserialize_with = "empty_date_time_as_option")]
```
andreievg, line 41: this should map directly, it's nullable

server/src/sync/translation_remote/invoice_line.rs
```diff
@@ -57,6 +57,17 @@ pub struct LegacyTransLineRow {
     pub quantity: i32,
     #[serde(deserialize_with = "empty_str_as_option")]
     pub note: Option<String>,
+
+    #[serde(rename = "om_item_code")]
+    #[serde(deserialize_with = "empty_str_as_option")]
```
andreievg, line 62: this should map directly, it's nullable

server/src/sync/translation_remote/invoice.rs
```diff
@@ -130,6 +190,43 @@ pub struct LegacyTransactRow {
     pub confirm_time: NaiveTime,
 
     pub mode: TransactMode,
+
+    #[serde(rename = "om_created_datetime")]
+    pub created_datetime: Option<NaiveDateTime>,
+
+    #[serde(rename = "om_allocated_datetime")]
+    #[serde(default)]
+    #[serde(deserialize_with = "empty_date_time_as_option")]
```
andreievg, line 199: this should map directly, it's nullable, also for other om_fields in invoice

server/src/sync/translation_remote/requisition.rs
```diff
@@ -79,25 +108,46 @@ pub struct LegacyRequisitionRow {
...
+    pub sent_datetime: Option<NaiveDateTime>,
+
+    #[serde(rename = "om_finalised_datetime")]
+    #[serde(default)]
+    #[serde(deserialize_with = "empty_date_time_as_option")]
+    pub finalised_datetime: Option<NaiveDateTime>,
+
+    #[serde(rename = "om_expected_delivery_date")]
+    #[serde(default)]
+    #[serde(deserialize_with = "zero_date_as_option")]
```
andreievg, line 140: this should map directly, it's nullable, also for other om_field in requisition

server/src/sync/translation_remote/requisition.rs
```diff
@@ -79,25 +108,46 @@ pub struct LegacyRequisitionRow {
...
+    #[serde(rename = "om_created_datetime")]
+    pub created_datetime: Option<NaiveDateTime>,
+
+    #[serde(rename = "om_sent_datetime")]
+    #[serde(default)]
+    #[serde(deserialize_with = "empty_date_time_as_option")]
+    pub sent_datetime: Option<NaiveDateTime>,
+
+    #[serde(rename = "om_finalised_datetime")]
+    #[serde(default)]
```
andreievg, line 134: do we need serde(default) on om fields ?

server/src/sync/translation_remote/requisition.rs
```diff
@@ -64,6 +64,35 @@ pub enum LegacyRequisitionStatus {
...
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum RequisitionStatus {
+    Draft,
+    New,
+    Sent,
+    Finalised,
+}
+
+impl RequisitionStatus {
```
andreievg, line 76: Do we need this ?

andreievg:
I've made this comment a few times now, sorry I'll keep on making it:
https://github.com/openmsupply/remote-server/blob/85b5b7cdeadd12d9dfed30d899c5c70be12886ab/server/src/sync/translation_remote/requisition.rs#L23-L47

Why do we need to be concerned with other types of legacy types ? I think it creates clutter, we may not ever need to implement this functionality (it would be up to stakeholders to decide if we need them)

server/src/sync/translation_remote/requisition.rs
```diff
@@ -122,19 +172,54 @@ impl RemotePullTranslation for RequisitionTranslation {
...
+                data.daysToSupply as f64 / NUMBER_OF_DAYS_IN_A_MONTH,
+                from_legacy_status(&data.status).ok_or(SyncTranslationError {
+                    table_name,
+                    source: anyhow::Error::msg(format!(
+                        "Unsupported requisition status: {:?}",
+                        data.status
+                    )),
+                    record: sync_record.data.clone(),
+                })?,
+                data.colour.and_then(|colour| req_colour_to_hex(colour)),
```
andreievg, line 213: I don't think we need to be converting legacy row colours

andreievg:
https://github.com/openmsupply/remote-server/blob/85b5b7cdeadd12d9dfed30d899c5c70be12886ab/server/src/sync/translation_remote/requisition.rs#L339

`&` can be removed

server/src/sync/translation_remote/requisition.rs
```diff
@@ -122,19 +172,54 @@ impl RemotePullTranslation for RequisitionTranslation {
...
+            ),
+        };
+
         let t = from_legacy_type(&data.r#type).ok_or(SyncTranslationError {
```
andreievg, line 217: would prefer r#type rather then t

andreievg:
https://github.com/openmsupply/remote-server/blob/85b5b7cdeadd12d9dfed30d899c5c70be12886ab/server/src/sync/translation_remote/stocktake_line.rs#L30-L33

forgot to reply to todo in earlier PRs, in mSupply it's set to some uuid that does not point to any stock_line. I think what you have as `empty_str_as_option` is fine

andreievg:
https://github.com/openmsupply/remote-server/blob/85b5b7cdeadd12d9dfed30d899c5c70be12886ab/server/src/sync/translation_remote/stocktake_line.rs#L71-L75

looks good

server/src/sync/translation_remote/stocktake_line.rs
```diff
@@ -82,7 +87,7 @@ impl RemotePullTranslation for StocktakeLineTranslation {
             pack_size: Some(data.snapshot_packsize),
             cost_price_per_pack: Some(data.cost_price),
             sell_price_per_pack: Some(data.sell_price),
-            note: None,
+            note: data.note,
         };
         // the following lines are from stock_line, manually unset when stock_line is None
         // (because they are otherwise set to Some(0.0)
```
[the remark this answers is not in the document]
andreievg: Also saw this:

https://github.com/openmsupply/remote-server/blob/85b5b7cdeadd12d9dfed30d899c5c70be12886ab/server/src/sync/translation_remote/stocktake_line.rs#L161-L183

and just looked at stocktake service, sorry I've missed this earlier.

Shouldn't we populate all stocktake fields = stock_line fields, and when stocktake is finalised, if we see a change between stock-line and stocktake fields we update them ? It seems like the easiest way to do it, otherwise front end would need to do quite a bit more (and it's seems more prone to error)

server/src/sync/translation_remote/stocktake.rs
```diff
@@ -51,8 +52,17 @@ pub struct LegacyStocktakeRow {
     pub stock_take_created_date: NaiveDate,
     #[serde(rename = "stock_take_date")]
```
andreievg, line 53: there is also stock_take_time

andreievg:
I think i've looked at all of the translations now, sorry quite a lot of comments, I wanted to review it all at once when all new fields are added. I'll try to see which changes are most important tomorrow

server/src/sync/translation_remote/invoice_line.rs
clemens-msupply: thanks done in [7224574](https://github.com/openmsupply/remote-server/pull/968/commits/7224574a99eb6abffafa7316f7c8b3ff40f279fc)

server/src/sync/translation_remote/invoice_line.rs
clemens-msupply: Yes that annoyed me as well and I simplified it: [c35bcbe](https://github.com/openmsupply/remote-server/pull/968/commits/c35bcbe63cca7d87a340b8cd33ae47f164c68cc0)

However, while doing it I remembered why I didn't do it earlier. It required some change to the SyncTranslationError which in turn required some more work in the central translation. On the plus side the central translation are using a translation trait as well now:
[a688238](https://github.com/openmsupply/remote-server/pull/968/commits/a6882386c33196b7565d0071eee0e066b9665817)

server/src/sync/translation_remote/invoice.rs
clemens-msupply: thanks done, I always mapped sg to New, right? [70687aa](https://github.com/openmsupply/remote-server/pull/968/commits/70687aafe97f559b0e092875bef908d697d9c294)

server/src/sync/translation_remote/invoice.rs
clemens-msupply: does it hurt to keep it in? can remove

server/src/sync/translation_remote/invoice.rs
clemens-msupply: Thanks done in [bd124f3](https://github.com/openmsupply/remote-server/pull/968/commits/bd124f3393ae16856dc8b2d65154dc3e66e584ac)

I m not sure anymore if serde_name is the way to go here, having it called om_created_datetime would make it a bit clearer in the first place. Had this a couple of times now where I was not sure anymore it its an om field or not. Staying close to the original data structure is probably the better option here.

server/src/sync/translation_remote/invoice.rs
clemens-msupply: The other option was to add the serde annotations to the DB row but found it weird to define something needed for translation in our DB layer. Since we always use SCREAMING_SNAKE_CASE for enums it its not that bad though. Found it cleaner to just copy it. 50/50 on it, whats your opinion?

server/src/sync/translation_remote/invoice.rs
clemens-msupply: thanks done in [456a086](https://github.com/openmsupply/remote-server/pull/968/commits/456a086265d2eed2a503dc2347bc67bdafa1f83b)

server/src/sync/translation_remote/invoice.rs
clemens-msupply: I made I None for InventoryAdjustment:

[5e01db5](https://github.com/openmsupply/remote-server/pull/968/commits/5e01db549328f606ad498903ead84b6ce49ef5d4)

The time was always set in mSupply (might work to push it as optional though)...

clemens-msupply:
> 

I created a ticket but left it as is (like to clarify first...):
https://github.com/openmsupply/remote-server/issues/990

server/src/sync/translation_remote/requisition_line.rs
clemens-msupply: think its safer to have keep it?, e.g. 28, 30 or 31 days in a month?

server/src/sync/translation_remote/invoice_line.rs
andreievg: thanks looks great

server/src/sync/translation_remote/invoice.rs
andreievg: yep, thanks

server/src/sync/translation_remote/invoice.rs
andreievg: There is no test to verify that this conversion is correct (that the actual colour mSupply displays is exactly the same that omSupply displays), for that reason I think it should be removed, without the test we need extra comments to confirm that colours are translated correctly (i think both will be a bit of extra work that we don't need to do)

server/src/sync/translation_remote/invoice.rs
andreievg: hmm, i am pretty binary in this for consistency (I think consistency is super important), if we keep to the original data shape then we should do also keep to original data type, would mean remove all of the serde methods including serde(deserialise_with and serialise_with), default etc.. this would mean we also need to make sure Option<T> is only on the fields that are nullable in LegacyRows, and we would need to add conversions from Option<T> to non optional fields (i.e. reverse of empty_str_as_option)

I still think using serde renaming and manual methods for se/de is good though, i do agree that it's not that clear that date.created_datetime is a new om_supply field. Can flag this for a discussion

server/src/sync/translation_remote/invoice.rs
andreievg: Hmm, we've been removing intermediary types, I think it's an ok compromise to add Deserialise and Serialise to db type for statuses and types etc.., it's quite a common derive (we don't need to specify rename_all there, just derive()), ohh man it would be soo good if rust allowed derives outside of struct definition.

server/src/sync/translation_remote/requisition_line.rs
andreievg: sorry I think i got confused with another one we added for om_max_months_of_stock, this one was ok

server/src/sync/translation_remote/requisition_line.rs
clemens-msupply: I tried but I am getting Nulls back as empty strings, maybe something in the json serialization layer in mSupply?

server/src/sync/translation_remote/stocktake.rs
clemens-msupply: but its not needed is it?

server/src/sync/translation_remote/requisition_line.rs
andreievg: oki, let's revisit this

clemens-msupply:
> I've made this comment a few times now, sorry I'll keep on making it:
> 
> https://github.com/openmsupply/remote-server/blob/85b5b7cdeadd12d9dfed30d899c5c70be12886ab/server/src/sync/translation_remote/requisition.rs#L23-L47
> 
> Why do we need to be concerned with other types of legacy types ? I think it creates clutter, we may not ever need to implement this functionality (it would be up to stakeholders to decide if we need them).
> 
> This also goes for other types and statuses not just requisition

Because they are part of the legacy enum. I could create an Other variant but I had to lookup the types anyway so why not document it? and its easy to ignore them in a match...

server/src/sync/translation_remote/stocktake_line.rs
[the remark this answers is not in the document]
clemens-msupply: I removed the code from your first comment [aae63e0](https://github.com/openmsupply/remote-server/pull/968/commits/aae63e02cbdac6f26150c2d99efb2f26aca29869). 

Regarding the stocktake service. The stock_line data in the stocktake_line is currently only used to create a new stock_line if it is a new item. Or should it also be used to update the stock_line? But I remember there where some discussion how this is logged...

Anyway for the push code we shouldn't fall back to the stock_line at all should we and only use push the field that where set in omSupply, right?

server/src/sync/translation_remote/requisition.rs
clemens-msupply: done

server/src/sync/translation_remote/requisition.rs
```diff
@@ -122,19 +172,54 @@ impl RemotePullTranslation for RequisitionTranslation {
                 }
             })?;
 
+        let (
+            created_datetime,
+            sent_datetime,
+            finalised_datetime,
```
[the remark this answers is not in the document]
clemens-msupply: Agree, we do in the future. Use to do i a lot in TS but in TS its also a lot easier to do than in Rust

clemens-msupply:
> https://github.com/openmsupply/remote-server/blob/85b5b7cdeadd12d9dfed30d899c5c70be12886ab/server/src/sync/translation_remote/requisition.rs#L336-L354
> 
> Hmm, i don't think[ this issue ](https://github.com/openmsupply/remote-server/issues/703) gave enough info (although it does suggest status translation is based on type).
> 
> I think we would need o ignore `sg` type ResponseRequisition and make `cn` type RequestRequisition = Sent:
> 
> From legacy
> 
> Request
> 
> `sg` -> draft `cn` -> sent `fn` -> sent
> 
> Response
> 
> `sg` -> ignore record `cn` -> new `fn` -> finalised
> 
> To legacy
> 
> Request
> 
> `draft` -> `sg` `sent` -> `fn` `finalised` -> `fn`
> 
> Response
> 
> `new` -> `cn` `finalised` -> `fn`

Done: [53c9883](https://github.com/openmsupply/remote-server/pull/968/commits/53c9883df18e895b59a6fc8e299340453f4a7bbc)

server/src/sync/translation_remote/stocktake_line.rs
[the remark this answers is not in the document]
andreievg: i think so, if we just take the values form legacy row it should be ok for now

server/src/sync/translation_remote/stocktake.rs
andreievg: this is really wierd, i've just checked what happens in mSupply:

stock_take_create_date, is set when stocktake is created automatically
stock_take_time is set when stocktake is created (so can be used to make created time)
stock_take_date is settable, so that's omSupply stocktake_date

andreievg:
> Because they are part of the legacy enum. I could create an Other variant but I had to lookup the types anyway so why not document it? and its easy to ignore them in a match...

The reason i make this comment:

* With "other" it's easy to see what we are actually integrating 
* Having more translations then we are integrating adds extra code (and it's always better to have less code)
* It's very likely that we are not going to be implementing all of mSupply logic in omSupply, 
* mSupply has a huge features set, i think it's time consuming to accomodate all of existing features, and associated types and translations.
* There are some field types/statuses that are not mentioned in mSupply documentation, and would require looking through all of mSupply code base to find them, if those are not mentioned we get a serde error, which is harder to digest if we are logging it (this was already a case before we starting skipping translations:https://github.com/openmsupply/remote-server/issues/919 )
* When the time comes to implement the new feature, we would have it in full context (and understanding), and it's easy to add the extra type/status etc.. at that stage 

There are a few other points, but those are the main ones

andreievg:
thanks for all of the changes

andreievg (approved):

server/src/sync/translation_remote/invoice.rs
clemens-msupply: yes lets discuss, renaming and type conversion are two different things...

server/src/sync/translation_remote/invoice.rs
clemens-msupply: Removed in [4a58540](https://github.com/openmsupply/remote-server/pull/968/commits/4a58540da9e3cc80c76e950e8c2661824c9307c3)

server/src/sync/translation_remote/requisition.rs
clemens-msupply: [4a58540](https://github.com/openmsupply/remote-server/pull/968/commits/4a58540da9e3cc80c76e950e8c2661824c9307c3)

server/src/sync/translation_remote/requisition.rs
```diff
@@ -122,19 +172,54 @@ impl RemotePullTranslation for RequisitionTranslation {
...
+                        source: anyhow::Error::msg(
+                            "Invalid data: om_created_datetime set but om_status missing",
+                        ),
+                        record: sync_record.data.clone(),
+                    })?,
+                data.om_colour,
+            ),
+            None => (
+                date_and_time_to_datatime(data.date_entered, 0),
+                None,
```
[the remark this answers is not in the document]
clemens-msupply: Using lastModifiedAt now:
[018ce61](https://github.com/openmsupply/remote-server/pull/968/commits/018ce61911ab4820a400be6377b6a75866afd969)

server/src/sync/translation_remote/stocktake_line.rs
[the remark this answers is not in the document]
clemens-msupply: Ok thanks I don't try to auto fill value but just use the original StocktakeLineRow:
[7fec7fc](https://github.com/openmsupply/remote-server/pull/968/commits/7fec7fce9cceff7b8c2d8373eba375e412eb0384)

server/src/sync/translation_remote/stocktake.rs
clemens-msupply: Ok I use stock_take_time as stocktake_created_time now: [26a9599](https://github.com/openmsupply/remote-server/pull/968/commits/26a9599b6818dfd543db75bd98ef982bbd9ee151)

server/src/sync/translation_remote/invoice.rs
andreievg: > Staying close to the original data structure is probably the better option here

I would assume staying close to original data structure implies types too

clemens-msupply:
> 

Found it quite use full to have the full picture when doing the translation. I put the unused fields in a comment.
I also added an Other variant to all enums. [abab866](https://github.com/openmsupply/remote-server/pull/968/commits/abab8663ede94792fb7c2ed45601033c5e975620)

server/src/sync/translation_remote/requisition.rs
clemens-msupply: Just looked into it a bit more. We only need it for fields having a "deserialize_with". Don't know why they are not working because the deserialize_with implementations use Option::deserialize(d) so it should work with missing fields but for some reason it doesn't...

I removed the default for some of the fields though: [c5dd65b](https://github.com/openmsupply/remote-server/pull/968/commits/c5dd65bb2ebcd029d802748983febab051b74368)

server/src/sync/translation_remote/requisition.rs
clemens-msupply: removed in [85963a4](https://github.com/openmsupply/remote-server/pull/968/commits/85963a4c428e71a89fa31913c1f684b3d9dbe576)

server/src/sync/translation_remote/invoice.rs
clemens-msupply: Thanks done in [85963a4](https://github.com/openmsupply/remote-server/pull/968/commits/85963a4c428e71a89fa31913c1f684b3d9dbe576)

PR #974, opened 2022-03-27
973 consumption and stock on hand repositories

### LINKED ISSUE #973: Consumption and stock on hand repositories

andreievg:
closes #973 

Requisition charts need to use consumption repository. Also a bit of logic in repositories (which i think should just hold the data access since we have services for logical bits).

Changes as per comments in PR


graphql/core/src/loader/item_stats.rs
```diff
@@ -1,17 +1,18 @@
...
+    service_provider::ServiceProvider,
+};
 use std::collections::HashMap;
 
 pub struct ItemsStatsForItemLoader {
     pub service_provider: Data<ServiceProvider>,
 }
 
-pub type ItemStatsLoaderInputPayload = Option<NaiveDateTime>;
+pub type ItemStatsLoaderInputPayload = u32; // amc_lookback_months
```
andreievg, line 15: Changed graphql input parameter

graphql/general/src/queries/tests/item_stats.rs
```diff
@@ -1,18 +1,19 @@
...
 
     #[actix_rt::test]
     async fn test_graphql_item_stats_loader() {
-        let (_, _, _, settings) = setup_graphl_test(
+        let (_, _, _, settings) = setup_graphl_test_with_data(
             GeneralQueries,
             EmptyMutation,
             "test_graphql_item_stats_loader",
             MockDataInserts::all(),
+            test_item_stats::mock_item_stats(),
```
andreievg, line 16: Test data was hard to decipher, so refactored using the new technique of defining test data

graphql/requisition/src/query_tests/requisition_loaders.rs
```diff
@@ -155,10 +155,12 @@ mod test {
                         supplyQuantity
                         suggestedQuantity
                         comment
-                        itemStats {
-                            averageMonthlyConsumption
-                            availableStockOnHand
-                            availableMonthsOfStockOnHand
+                        linkedRequisitionLine {
```
andreievg, line 158: When response requisition is generated, not taking snapshot of supplying store stats, i.e. response itemStats = live item stats but request item stats are snapshot

graphql/types/src/types/item.rs
```diff
@@ -58,14 +57,15 @@ impl ItemNode {
         &self,
         ctx: &Context<'_>,
         store_id: String,
-        look_back_datetime: Option<NaiveDateTime>,
+        #[graphql(default_with = "util::constants::default_amc_look_back_months()")]
```
andreievg, line 60: I kind of like the defaults in graphql like this, but I think need to back track on that idea, since i think defaults should be defined in service (so that service can be used in other areas with the same default)

repository/src/mock/mod.rs
```diff
@@ -302,10 +301,6 @@ fn all_mock_data() -> MockDataCollection {
         "mock_test_requisition_line_repository",
         mock_test_requisition_line_repository(),
     );
-    data.insert(
```
andreievg, line 305: can specify extra mock data now with test inits `_with_data`, and not having extra test data inserted with `all` mock data means id's can be simplified without being conflicted (makes test creation easier)

repository/src/mock/mod.rs
```diff
@@ -465,3 +460,49 @@ pub async fn insert_mock_data(
 
     mock_data
 }
+
+impl MockData {
+    pub fn join(mut self, other: MockData) -> MockData {
```
andreievg, line 465: see `mock_item_stats`

repository/src/mock/test_requisition_service.rs
```diff
@@ -253,45 +248,6 @@ pub fn mock_test_not_store_a_master_list() -> FullMockMasterList {
     }
 }
 
-pub fn mock_test_add_from_master_list() -> FullMockMasterList {
```
andreievg, line 256: relocated to the underlying test, since

repository/src/tests.rs
```diff
@@ -1146,72 +1146,6 @@ mod repository_test {
         );
     }
 
-    #[actix_rt::test]
-    async fn test_item_stats_repository() {
```
andreievg, line 1150: Relocated to new item stats service

service/src/requisition/request_requisition/generate.rs
```diff
@@ -35,24 +39,26 @@ pub fn generate_suggested_quantity(
...
     item_ids: Vec<String>,
 ) -> Result<Vec<RequisitionLineRow>, RepositoryError> {
-    let repository = ItemStatsRepository::new(&connection);
-
-    let filter = ItemStatsFilter::new().item_id(EqualFilter::equal_any(item_ids));
-    let item_stats_rows = repository.query(store_id, None, Some(filter))?;
+    let item_stats_rows = get_item_stats(
+        ctx,
+        store_id,
+        DEFAULT_AMC_LOOK_BACK_MONTHS,
```
andreievg, line 50: As per earlier comment, maybe service should apply default

service/src/sync_processor/requisition/common.rs
```diff
@@ -111,40 +110,22 @@ fn generate_linked_requisition_lines(
...
-
-        let new_row = inline_init(|r: &mut RequisitionLineRow| {
-            r.id = uuid();
-            r.requisition_id = linked_requisition.id.clone();
-            r.item_id = source_line.requisition_line_row.item_id;
-            r.requested_quantity = source_line.requisition_line_row.requested_quantity;
-            r.suggested_quantity = source_line.requisition_line_row.suggested_quantity;
-            r.available_stock_on_hand = item_stats.available_stock_on_hand();
-            r.average_monthly_consumption = item_stats.average_monthly_consumption();
+        new_lines.push(RequisitionLineRow {
```
andreievg, line 113: shouldn't have used inline_init when generating new row (to reduce chance of missing newly added fields

service/src/sync_processor/requisition/common.rs
```diff
@@ -111,40 +110,22 @@ fn generate_linked_requisition_lines(
...
+            comment: None,
         });
-
-        new_lines.push(new_row);
     }
 
     Ok(new_lines)
 }
-
-fn get_item_stats(
```
andreievg, line 133: As per earlier comment about not snapshotting response requisition stats

util/src/date_operations.rs
```diff
@@ -1,10 +1,83 @@
...
+use chrono::{Datelike, Duration, NaiveDate, Utc};
 
-pub fn date_now_with_offset(off_set: Duration) -> NaiveDate {
-    let now = Utc::now().naive_local();
-    now.checked_add_signed(off_set).unwrap_or(now).date()
+pub fn date_now_with_offset(off_set: Duration, is_addition: bool) -> NaiveDate {
+    date_with_offset(&date_now(), off_set, is_addition)
+}
+
+pub fn date_with_offset(date: &NaiveDate, off_set: Duration, is_addition: bool) -> NaiveDate {
```
clemens-msupply, line 7: Duration also supports negative values. Not well documented but I tried it here:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=%23!%5Ballow(unused)%5D%0A%2F%2F%20Hello%20%7Barg%200%20(%22x%22)%7D%20is%20%7Barg%201%20(0.01)%20with%20precision%20specified%20inline%20(5)%7D%0Afn%20main()%20%7B%0A%20let%20duration%20%3D%20chrono%3A%3ADuration%3A%3Ananoseconds(500%20*%201000000000%20%2B%20500)%3B%0A%20let%20d2%20%3D%20chrono%3A%3ADuration%3A%3Ananoseconds(-(100%20*%201000000000%20%2B%20300))%3B%0A%20println!(%22%7B%3A%3F%7D%22%2C%20d2)%3B%0A%20println!(%22%7B%3A%3F%7D%22%2C%20duration.checked_add(%26d2))%0A%7D

util/src/constants.rs
```diff
@@ -16,3 +16,8 @@ pub fn stock_line_expiring_soon_offset() -> Duration {
 pub fn expected_delivery_date_offset() -> Duration {
     Duration::weeks(2)
 }
+/// Default AMC calculation lookback
+pub const DEFAULT_AMC_LOOK_BACK_MONTHS: u32 = 3;
+pub fn default_amc_look_back_months() -> u32 {
+    DEFAULT_AMC_LOOK_BACK_MONTHS
+}
```
clemens-msupply, line 23: Why not use the const directly? or remove pub from the const value?

service/src/requisition/request_requisition/generate.rs
clemens-msupply: Nice thing about the current version is that it gives you a clearer indication that get_item_stats is calculated based on a look back parameter (compared to a None parameter).

repository/src/db_diesel/consumption.rs
```diff
@@ -0,0 +1,80 @@
...
+use diesel::{QueryDsl, RunQueryDsl};
+
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct ConsumptionFilter {
+    pub item_id: Option<EqualFilter<String>>,
+    pub store_id: Option<EqualFilter<String>>,
+    pub date: Option<DateFilter>,
+}
+
+pub struct ConsumptionRepository<'a> {
```
clemens-msupply, line 16: This is all about amc, right? maybe perfix repo, table and row with it to make it clearer?

clemens-msupply (approved):
Looks very good thanks

graphql/types/src/types/item.rs
andreievg: > Probably depends on the default if it should go into gql or service. For example, I had some trouble before with the pagination option in the service not allowing me to always fetch all items in one query. In this example I think its actually better to have the pagination defaults in the gql layer.

good point, i've actually changed some defaults to services in the last PR, it's just amc_look_back (service is used with the default) and expected_delivery_date (only used in graphql, i made this one service default at first, but after  reading your comment changed to graphql) https://github.com/openmsupply/remote-server/compare/3fbcdd3a46ef43e1840529d7dc2270f7d0e270b2...983-Requisition-line-chart-graphql-mapping

repository/src/mock/mod.rs
andreievg: thanks

service/src/requisition/request_requisition/generate.rs
andreievg: Yep, although, i wonder if it's relevant here (can assume None is default for something like get_item_stats), i think eventually this will come from settings, they might live in ctx or db query

graphql/types/src/types/item_stats.rs
```diff
@@ -1,41 +1,30 @@
...
-        self.available_stock_on_hand
+    pub async fn available_stock_on_hand(&self) -> u32 {
+        self.item_stats.available_stock_on_hand
     }
 
     pub async fn available_months_of_stock_on_hand(&self) -> f64 {
-        if self.average_monthly_consumption == 0 {
-            return self.available_stock_on_hand as f64;
+        if self.item_stats.average_monthly_consumption == 0.0 {
+            return self.item_stats.available_stock_on_hand as f64;
```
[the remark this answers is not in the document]
andreievg: great, thanks: finally get to use then on boolean c924647c8098ad6100536cd973b5bae61d5fb7bb

graphql/types/src/types/requisition_line.rs
```diff
@@ -127,13 +127,38 @@ impl RequisitionLineNode {
...
 
-    /// Snapshot Stats (when requisition was created)
-    pub async fn item_stats(&self) -> ItemStatsNode {
-        ItemStatsNode {
-            average_monthly_consumption: self.row().average_monthly_consumption,
-            available_stock_on_hand: self.row().available_stock_on_hand,
-            snapshot_datetime: self.row().snapshot_datetime.clone(),
+    /// For request requisition: snapshot stats (when requisition was created)
+    /// For response requisition current item stats
+    pub async fn item_stats(
```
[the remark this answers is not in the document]
andreievg: just when new line is added (i'll make an issue for requisition left over sometimes, when it was discussed last we thought of also adding, re-calculate or re-fresh suggested, which will update this)

repository/src/db_diesel/consumption.rs
andreievg: Not just amc, but also consumption, not sure what you mean about prefix

util/src/constants.rs
andreievg: first one was used, then the second, and then reverted back to 1: e5d2c48334a58166bb87ab406964b68909d2a6f6

util/src/date_operations.rs
andreievg: Thanks ! : dfe91f7e78a793d0e8dd280e944bb8b8fcc316fe

repository/src/db_diesel/consumption.rs
clemens-msupply: Was unsure about what kind of consumption but just had another look and it seems clear to me now

PR #976, opened 2022-03-27
975 add requisition line chart service (consumption)

### LINKED ISSUE #975: Requisition line chart (consumption) service

andreievg:
closes #975


service/src/requisition_line/chart/historic_consumption.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+pub fn get_historic_consumption_for_item(
+    connection: &StorageConnection,
+    store_id: &str,
+    item_id: &str,
+    reference_date: NaiveDate,
+    options: ConsumptionHistoryOptions,
+) -> Result<Vec<ConsumptionHistory>, RepositoryError> {
+    // Initilise series
```
clemens-msupply, line 33: spelling

service/src/requisition_line/chart/mod.rs
```diff
@@ -0,0 +1,354 @@
...
+        &requisition_line.requisition_line_row.item_id,
+        requisition_line_datetime.date(),
+        consumption_history_options,
+    )?;
+
+    // Replace last consumption_history element with requisition line AMC (current AMC)
+    let requsition_line_row = &requisition_line.requisition_line_row;
+    if let Some(last) = consumption_history.last_mut() {
+        last.consumption = requsition_line_row.average_monthly_consumption as u32;
+        last.average_monthly_consumption = requsition_line_row.average_monthly_consumption as f64;
```
clemens-msupply, line 51: Why doesn't the last consumption_history element contains the current AMC? whats the difference in the calculation?

service/src/requisition_line/chart/historic_consumption.rs
```diff
@@ -0,0 +1,274 @@
+use std::ops::Neg;
```
clemens-msupply, line 1: Think this file is not requisition specific but could go into its own service, e.g. historic_avg_monthly_item_consumption.rs

Some naming nitpicking: in the service layer its not really about "charts" but more about "statistic data" and this data can then be used to produce charts by the frontend, but it could also be used to create other kind of reports.

clemens-msupply (approved):
Looks good

service/src/requisition_line/chart/historic_consumption.rs
andreievg: 994356a3175722bdb19bb53524b1db5a436c13a7, thanks

service/src/requisition_line/chart/mod.rs
andreievg: AMC in the last bar should match AMC in requisition line, which is 'snapshot' amc not quite historic monthly amc, it's calculated from the datetime of insertion of line. Historic amc and consumption is calculated on monthly bases (taking months as a whole)

telegram link to convo https://t.me/c/1673216748/602

service/src/requisition_line/chart/historic_consumption.rs
andreievg: Yep, agree, and i've renamed the result to ItemCharts, was going to relocate it to a more common place when need arrises (when it needs to be used in other services)

andreievg:
changes were done in https://github.com/openmsupply/remote-server/pull/984, closing

PR #978, opened 2022-03-27
977 add requisition line chart service (stock evolution)

### LINKED ISSUE #977: Requisition line chart (stock evolution) service

andreievg:
closes #977


repository/migrations/postgres/2021-12-25T10-00_item_stats/up.sql
```diff
@@ -1,26 +1,70 @@
...
+	store_id,
+	verified_datetime as datetime
+FROM invoice_line 
+JOIN invoice
+	ON invoice_line.invoice_id = invoice.id
+WHERE invoice.type = 'INVENTORY_ADJUSTMENT' 
+	AND verified_datetime IS NOT NULL
+	AND invoice_line.number_of_packs > 0;
+		
+CREATE VIEW stock_movement AS
```
andreievg, line 50: this could be done in one view (without the union, and with `case` statements, but thought breaking down by individual types seems tidier)

service/src/requisition_line/chart/stock_evolution.rs
```diff
@@ -0,0 +1,320 @@
...
+            reference_date,
+            quantity: running_stock_on_hand,
+        })
+    }
+
+    // Reverse
+    result.into_iter().rev().collect()
+}
+
+fn calculate_projected_stock_evolution(
```
clemens-msupply, line 162: Nice how its projected including the expected delivery data. Maybe some high level overview how it is projected:

``` rust
/// Projects the daily stock level using two factors:
/// 1) stock is decreased based on the average_daily_consumption
/// 2) stock is increased based the expected delivery date of an expected delivery
```

projected_points must be consecutive days, right? maybe use a range instead so that it can be enforced? or just add a comment

service/src/requisition_line/chart/stock_evolution.rs
```diff
@@ -0,0 +1,320 @@
...
+    reference_stock_on_hand: u32,
+    average_monthly_consumption: f64,
+    suggested_quantity: u32,
+    expected_delivery_date: NaiveDate,
```
clemens-msupply, line 166: Think this could be more independent from the requisition use-case by renaming `suggested_quantity` to `expected_delivery_quantity`.

This could even more generalized (if needed) to `expected_deliveries: Vec<{quantity, delivery_date}>`

service/src/requisition_line/chart/stock_evolution.rs
```diff
@@ -0,0 +1,320 @@
...
+pub fn get_stock_evolution_for_item(
+    connection: &StorageConnection,
+    store_id: &str,
+    item_id: &str,
+    reference_datetime: NaiveDateTime,
+    reference_stock_on_hand: u32,
+    expected_delivery_date: NaiveDate,
+    suggested_quantity: u32,
+    average_monthly_consumption: f64,
+    options: StockEvolutionOptions,
```
clemens-msupply, line 42: As mentioned in the previous PR think this file is not requisition specific but could go into its own "stock stats" service.

util/src/date_operations.rs
```diff
@@ -13,10 +13,37 @@ pub fn date_with_offset(date: &NaiveDate, off_set: Duration, is_addition: bool)
...
+}
+
 pub fn date_now() -> NaiveDate {
     Utc::now().naive_local().date()
 }
 
+pub fn datetime_with_offset(
+    date: &NaiveDateTime,
+    off_set: Duration,
+    is_addition: bool,
```
clemens-msupply, line 28: use negative Duration instead

repository/migrations/postgres/2021-12-25T10-00_item_stats/up.sql
```diff
@@ -1,26 +1,70 @@
...
+    number_of_packs * pack_size * -1 as quantity,
+	item_id,
+	store_id,
+	picked_datetime as datetime
+FROM invoice_line 
+JOIN invoice
+	ON invoice_line.invoice_id = invoice.id
+WHERE invoice.type = 'OUTBOUND_SHIPMENT' 
+	AND picked_datetime IS NOT NULL
+	AND invoice_line.number_of_packs > 0
```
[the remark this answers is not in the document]
andreievg: yep just checking not 0

service/src/requisition_line/chart/stock_evolution.rs
andreievg: Great idea, i'll change it when doing further charts work (when adding ItemCharts for item end point)

service/src/requisition_line/chart/stock_evolution.rs
andreievg: Agree, i've made a comment, will change when expanding to use this functionality in other areas

util/src/date_operations.rs
andreievg: https://github.com/openmsupply/remote-server/commit/dfe91f7e78a793d0e8dd280e944bb8b8fcc316fe

repository/migrations/postgres/2021-12-25T10-00_item_stats/up.sql
```diff
@@ -1,26 +1,70 @@
...
+-- TODO sc type ?
 CREATE VIEW consumption AS
 SELECT 
     'n/a' as id,
     items_and_stores.item_id AS item_id, 
     items_and_stores.store_id AS store_id,
-	COALESCE(consumption.quantity, 0) AS quantity,
-	consumption.date::date AS date
+	-- @ = absolute value
+	@ COALESCE(consumption.quantity, 0) AS quantity,
```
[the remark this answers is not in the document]
andreievg: it's using `outbound_shipment_stock_movement` now, which would be negative quantity, i'll rename: 1b247d473313ecc40c39c342fd5cdc0ef1f93959

service/src/requisition_line/chart/stock_evolution.rs
andreievg: I'll make sure to comment each of those statistic methods when they are changed to be more common (when the data is also used in item node -> charts)

> maybe use a range instead so that it can be enforced?

it's a private method used only by `get_stock_evolution_for_item` which uses another private method generate_evolution_series. It's much nicer to test that way, can test series generation from 'range' independently of historic/projected calculation method. I think the exposed method will be something like get_projected_stock_on_hand, which will take a range.

andreievg:
changes were done inhttps://github.com/openmsupply/remote-server/pull/984, closing

andreievg:
oops merged parent branch into this one, so will merge this into dev

PR #986, opened 2022-03-30
960 central login

### LINKED ISSUE #960: User login via Central Server
In our last sync discussion, we've talked about user management and decided that the quickest way to do it would be to use existing mobile mechanism of logging in:

* Remote Server tries to use Central Server user end point:
  * If it can access central server then user information is pulled and saved in db
  * If it can't access it then local db lookup is made

Remote Server is multi sore and Mobile is single store, i think we have a slight problem: 
* Need to know what stores user can log in when api request to central is made

`/sync/v3/user` accepts storeId, is it required field ? Can it be made optional ? Can it return user stores ? 

We've also talked about permissions being returned by user store, I don't think we would have time to do permissions for V1, but just wanted to flag it

clemens-msupply:
Every login we now try to update the user data from the central server. If user data has been retrieved successfully existing data is replaced with the new data. From then on the login process is as before.

The user stores and the default store are available in the gql /me endpoint.

Other notes:
- I used a 10sec connection timeout to reach the central server (same as in mobile).
- Best effort mapping of central server permissions to our permissions.
- There where some async hickups using the service context and I pass in the ServiceProvider to the LoginService (see comment in code)
- In my central server test file the Admin user had some invalid stores, I ignore them in a sub transaction.

Sorry for the large PR got a bit bigger than expected...

Edit: I put the `table!` macros for the new tables in the same file as the row schema definitions. I also put the `joinable!` and `allow_tables_to_appear_in_same_query!` there. We probably should split up the existing `allow_tables_to_appear_in_same_query!` since it generates code for any pair in the argument list which sounds dangerous for compilation...

Closes #960


graphql/types/src/types/user.rs
```diff
@@ -1,29 +1,85 @@
...
+impl UserStoreNode {
+    pub async fn id(&self) -> &str {
+        &self.user_store.store_row.id
+    }
+
+    pub async fn code(&self) -> &str {
+        &self.user_store.store_row.code
+    }
+
+    pub async fn name(&self, ctx: &Context<'_>) -> Result<String> {
```
andreievg, line 21: If we are just exposting name string, why not just join name in UserStore ?

service/src/login.rs
```diff
@@ -0,0 +1,364 @@
...
+    InternalError(String),
+    DatabaseError(RepositoryError),
+}
+
+impl LoginService {
+    /// Note, this service takes a ServiceProvider instead of a ServiceContext. The reason is that a
+    /// ServiceContext can't be used across async calls (because of the containing thread bound
+    /// SqliteConnection). Since we need an async api call to the remote server to fetch user data
+    /// we need to create the service context after the call where the compiler can deduce that we are
+    /// not passing it to another thread.
```
andreievg, line 51: Hmm, form my understanding each api request is already running on a new thread (i.e. one call to api is not going to block another call to api). Can we not use blocking reqwest ?

service/src/login.rs
```diff
@@ -0,0 +1,364 @@
...
+        let max_age_refresh = chrono::Duration::hours(6).num_seconds() as usize;
+        let pair = match token_service.jwt_token(&user_account.id, max_age_token, max_age_refresh) {
+            Ok(pair) => pair,
+            Err(err) => return Err(LoginError::FailedToGenerateToken(err)),
+        };
+        Ok(pair)
+    }
+
+    /// # Arguments
+    /// * `client` client to be used to do the central server login request (mainly for testing)
```
andreievg, line 103: Can client be another service method that we can override for testing ?

service/src/login.rs
```diff
@@ -0,0 +1,364 @@
...
+    async fn fetch_user_from_central(
+        username: &str,
+        password: &str,
+        central_server_url: &str,
+        client: Option<Client>,
```
andreievg, line 108: Can we have client provided by another service method ? We can override it for testing, rather then passing optional

graphql/general/src/queries/login.rs
```diff
@@ -22,107 +20,56 @@ impl AuthToken {
...
-    let mut token_service = TokenService::new(
-        &auth_data.token_bucket,
-        auth_data.auth_token_secret.as_bytes(),
-    );
-    let max_age_token = chrono::Duration::minutes(60).num_seconds() as usize;
-    let max_age_refresh = chrono::Duration::hours(6).num_seconds() as usize;
-    let pair = match token_service.jwt_token(&user_account.id, max_age_token, max_age_refresh) {
+    let sync_settings = ctx.get_sync_settings();
+    let central_server_url = &sync_settings.url;
+    let pair = match LoginService::login(
```
andreievg, line 33: Would prefer a struct input, my recent experience of adding user_id input to services required me to change a lot of method input shapes. i.e. if we want to add timeout from settings it would be much easier to add it to a struct

graphql/general/src/queries/login.rs
```diff
@@ -22,107 +20,56 @@ impl AuthToken {
...
+            return Err(graphql_error.extend());
         }
     };
 
-    set_refresh_token_cookie(ctx, &pair.refresh, max_age_refresh, auth_data.debug_no_ssl);
+    let now = Utc::now().timestamp() as usize;
+    set_refresh_token_cookie(
+        &ctx,
+        &pair.refresh,
+        pair.refresh_expiry_date - now,
```
andreievg, line 68: I couldn't see in description why this was changed, was wondering what is the benefit of doing it inside login service (i would assume login service would just deal with checking user credentials etc..)

graphql/general/src/queries/login.rs
```diff
@@ -22,107 +20,56 @@ impl AuthToken {
...
-                            "Lock error".to_string(),
-                        ))
-                    }
-                },
-            })
+        Err(error) => {
+            let formatted_error = format!("{:#?}", error);
+            let graphql_error = match error {
+                LoginError::LoginFailure => {
+                    StandardGraphqlError::Unauthenticated("Invalid credentials".to_string())
```
andreievg, line 48: I think it's best to make api ergonommic for consumers, when there is an expected error that can be dealt with it should be a union result

service/src/login.rs
```diff
@@ -0,0 +1,364 @@
...
+            })
+            .await
+        {
+            Ok(result) => {
+                if result.status == LoginStatusV4::Error {
+                    return Err(FetchUserError::ConnectionError(
+                        "Failed to fetch user from central server".to_string(),
+                    ));
+                }
+                result
```
andreievg, line 134: It's kind of hard to see that result user_data is coming from here, also why do you check ::Error here but != ::Status and user_info below ?

service/src/login.rs
```diff
@@ -0,0 +1,364 @@
...
+            }
+        };
+
+        // convert user_info to internal format
+        let user = UserAccountRow {
+            id: user_info.user.id,
+            username: username.to_string(),
+            hashed_password: UserAccountService::hash_password(password)
+                .map_err(|err| FetchUserError::InternalError(format!("{:?}", err)))?,
+            email: match user_info.user.e_mail.as_str() {
```
andreievg, line 164: is_empty here seem more readable, can even do `is_empty().then(|| user_info.user.e_mail.to_string())`, for readabilty i like `let = email = user_info.user.e_mail` first

service/src/apis/permissions.rs
```diff
@@ -0,0 +1,409 @@
...
+        (505, Permissions::HISDischargePatients),
+        (506, Permissions::HISCreateWards),
+        (507, Permissions::HISCreateBeds),
+        (508, Permissions::HISEditWardDetail),
+        (509, Permissions::HISEditBedDetail),
+        (510, Permissions::HISCreateICDCodes),
+        (511, Permissions::HISEditICDCodes),
+        (512, Permissions::HISMergePatients),
+        (513, Permissions::HISAddEncounterDiseases),
+        (514, Permissions::HISEditEncounterDiseases),
```
andreievg, line 392: Wow this looks like a lot of work, are you sure we are going to be using all of these permissions ? I've made a suggestion a few time now to maybe just translate what is required for a given task, i'll keep on making that suggestion

service/src/apis/permissions.rs
```diff
@@ -0,0 +1,409 @@
...
+    CreateSupplierInvoices,
+    ViewSupplierInvoices,
+    CreateNewItems,
+    CreateEditBackorders,
+    CreateCustomerInvoices,
+    ViewCustomerInvoices,
+    AddEditGoodsReceived,
+    ManageTenders,
+    AddPatients,
+    /// not used
```
andreievg, line 50: Is this comment correct ?

repository/src/schema/user_permission.rs
```diff
@@ -0,0 +1,38 @@
...
+    Stocktake,
+    InboundShipment,
+    OutboundShipment,
+    Requisition,
+}
+
+#[derive(DbEnum, Debug, Clone, PartialEq, Eq)]
+#[DbValueStyle = "SCREAMING_SNAKE_CASE"]
+pub enum Permission {
+    Query,
```
andreievg, line 25: You've asked for my opinion about breaking down Permissions into query/mutate, i've suggested we analyse permissions as a whole to see which ones will be used and how before we do this, I think we should have kept the existing structure for the time being, i don't see how a lot of the permissions fit into the two categories, just having a breif look at permission list, it looks like a lot of them will not user query or mutate. I think we are defining a rigid structure too early

service/src/login.rs
```diff
@@ -0,0 +1,364 @@
...
+    let mut output = HashMap::new();
+    for per in permissions {
+        match per {
+            // stocktake
+            // TODO: correct
+            Permissions::ViewStock => {
+                update_permission(&mut output, Resource::Stocktake, Permission::Query)
+            }
+            Permissions::CreateStocktake => {
+                update_permission(&mut output, Resource::Stocktake, Permission::Mutate)
```
andreievg, line 243: Going back to another comment I've made about permission being classified to early, i think it would have been a lot simpler to use previous structure (just add resource to a vec)

service/src/apis/login_v4.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub phone2: String,
+    //date_created: "2017-10-11",
+    //date_left: "0000-00-00",
+    pub job_title: String,
+    pub responsible_officer: bool,
+    #[serde(rename = "Language")]
+    pub language: i32,
+    pub use_ldap: bool,
+    pub ldap_login_string: String,
+    pub receives_sms_errors: bool,
```
andreievg, line 64: I think we should only be translating field that we are using

service/src/token.rs
```diff
@@ -136,10 +136,12 @@ impl<'a> TokenService<'a> {
         refresh_token: &str,
         valid_for: usize,
         refresh_token_valid_for: usize,
+        leeway_sec: Option<u64>,
```
andreievg, line 139: Why was leeway_sec added ?

graphql/types/src/types/user.rs
```diff
@@ -1,29 +1,85 @@
...
+        &self.user.user_row.email
     }
 
     pub async fn username(&self) -> &str {
-        &self.user.username
+        &self.user.user_row.username
+    }
+
+    pub async fn default_store(&self) -> Option<UserStoreNode> {
+        let default_store = self
```
andreievg, line 60: I think this is a very good use case for a struct method

graphql/types/src/types/user.rs
```diff
@@ -1,29 +1,85 @@
...
+            .user
+            .stores
+            .iter()
+            .find(|s| s.user_store_join.is_default);
+        default_store.map(|user_store| UserStoreNode {
+            user_store: user_store.clone(),
+        })
+    }
+
+    pub async fn stores(&self) -> Vec<UserStoreNode> {
```
andreievg, line 70: We've used connectors so far for pretty much all store node collections, it's probably an overkill for this use case, what do you think ?

repository/src/db_diesel/user.rs
```diff
@@ -0,0 +1,191 @@
...
+    LeftJoin<LeftJoin<user_account::table, OnUserStoreJoinToUserJoin>, OnStoreJoinToUserStoreJoin>,
+    DBType,
+>;
+
+fn create_filtered_query(filter: Option<UserFilter>) -> BoxedUserQuery {
+    let mut query = user_dsl::user_account
+        .left_join(
+            user_store_join_dsl::user_store_join.on(user_store_join_dsl::user_id.eq(user_dsl::id)),
+        )
+        .left_join(store_dsl::store.on(store_dsl::id.eq(user_store_join_dsl::store_id)))
```
andreievg, line 154: Do we need to have these as manual joins ?

repository/src/db_diesel/user.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+#[derive(PartialEq, Debug, Clone, Default)]
+pub struct User {
+    pub user_row: UserAccountRow,
+    pub stores: Vec<UserStore>,
+}
+
+#[derive(PartialEq, Debug, Clone, Default)]
+pub struct UserStorePermissions {
```
andreievg, line 38: Is this used anywhere ?

repository/src/db_diesel/user.rs
```diff
@@ -0,0 +1,191 @@
...
+            stores: vec![],
+        });
+        if let (Some(user_store_join), Some(store_row)) = (join.1, join.2) {
+            entry.stores.push(UserStore {
+                store_row,
+                user_store_join,
+            })
+        }
+    }
+    let users = user_map.into_iter().map(|(_, user)| user).collect();
```
andreievg, line 130: into_values().collect()

repository/src/db_diesel/user.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let result = final_query.load::<UserAndUserStoreJoin>(&self.connection.connection)?;
+        Ok(to_domain(result))
+    }
+}
+
+fn to_domain(results: Vec<UserAndUserStoreJoin>) -> Vec<User> {
+    // collect all joins for a user
+    let mut user_map = HashMap::<String, User>::new();
+    for join in results {
```
andreievg, line 118: I think it's a good idea to destructure join here, so it's clear what 0, 1, 2 mean

repository/src/db_diesel/user.rs
```diff
@@ -0,0 +1,191 @@
...
+        //     "{}",
+        //     diesel::debug_query::<DBType, _>(&final_query).to_string()
+        // );
+
+        let result = final_query.load::<UserAndUserStoreJoin>(&self.connection.connection)?;
+        Ok(to_domain(result))
+    }
+}
+
+fn to_domain(results: Vec<UserAndUserStoreJoin>) -> Vec<User> {
```
andreievg, line 115: What is the benefit of getting User with user store joins here, vs using another repo for UserStoreJoin and loaders ? Btw default store id could be a field on user

repository/src/db_diesel/user_permission.rs
```diff
@@ -0,0 +1,130 @@
...
+    },
+};
+use crate::{EqualFilter, Pagination, Sort};
+
+use diesel::{dsl::IntoBoxed, prelude::*};
+
+pub type UserPermission = UserPermissionRow;
+
+#[derive(Clone, Default)]
+pub struct UserPermissionFilter {
```
andreievg, line 16: I don't really see UserPermissions being used anywhere, so would assume you added it preemptively, i think it needs store_id

andreievg: I wonder if we need user_store_join. If we added another permission, `can_login` this would eliminate need to for user_store_join, I would assume that in permission check, one of the first things to check if user can access store (which would require query to user_store_join, and after that a permission check would be needed), it seems simpler and maybe faster to do a query for user_id = user, store_id = store, permission in (can_login, permission_in_question)

repository/src/schema/user_store_join.rs
```diff
@@ -0,0 +1,25 @@
...
+      store_id -> Text,
+      is_default -> Bool,
+  }
+}
+
+joinable!(user_store_join -> user_account (user_id));
+joinable!(user_store_join -> store (store_id));
+
+allow_tables_to_appear_in_same_query!(user_store_join, user_account);
+allow_tables_to_appear_in_same_query!(user_store_join, store);
```
andreievg, line 16: as per your comment about macros generating a bit of code, I think it's worth checking the difference in code generated with cargo expand. I think it's possible to have a diesel error that is very hard to decipher, resulting from missing tables_appear_in_same_query (if there is not much code generated by this I would suggest we just have one place with this macro for all tables)

service/src/sync_settings.rs
```diff
@@ -0,0 +1,10 @@
+#[derive(serde::Deserialize, Clone)]
+pub struct SyncSettings {
```
andreievg, line 2: cool, i think most things can be in services tbh

service/src/user_account.rs
```diff
@@ -46,30 +56,84 @@ impl<'a> UserAccountService<'a> {
...
+                // remove existing user (if exists)
+                permission_repo.delete_by_user_id(&user.id)?;
+                user_store_repo.delete_by_user_id(&user.id)?;
+                user_repo.delete_by_id(&user.id)?;
+                // insert user
+                user_repo.insert_one(&user)?;
+                for store in stores_permissions {
+                    // The list may contain stores we don't know about; try to insert the store
+                    // in a sub-transaction and ignore the store when there is an error
+                    let sub_result = con.transaction_sync_etc(
```
andreievg, line 81: Personally i would have this inner block in a new method, I have a phobia of deep nesting

andreievg: would this work instead : https://github.com/openmsupply/remote-server/blob/897b22f404d743708acfe6c5d41e4783dbeda395/service/src/user_account.rs#L77-L92 (i ran the test they passed, but not sure if you testing this sub transaction in them, I haven't reviewed them). I would assume it works (since error skipping like this with sub transaction is typically used when you don't know or can't predict the possible error, in this case it's predictable).

andreievg (approved):
I've made quite a few comments. I think the only thing that is stopping immediate merge (if front end needs it), it's this one: https://github.com/openmsupply/remote-server/pull/986#discussion_r839437440

Yep, the PR is ver big, but it's very well written (i didn't review any tests and didn't understand what the changes in refresh token are). 

I think changes were modular enough to be broken down into smaller PRs (that's definitely one benefit of breaking down issue into smaller bullet points/tasks/issues it given an idea of how it can be PR'ed in smaller chunks)

I tried to give it a thorough review, but I wouldn't be surprised if i missed a few things (it's really hard to think about any possible logic problems with so many logic changes)

`very sorry` I felt like i wasn't being polite with some of my comments (including some things in this summary), i just tried to focus on review and forgot the diplomacy.

service/src/login.rs
andreievg: I did some benching, will publish a repository, can use blocking wit tokio::task::spawn_blocking (which is a bit hacky), I'll have to do a bit more testing with diesel and sync in API before conclusions.

anyways, for this though, we can break it down into two services, one to use async reqwest and another for sync db stuff  (a bit more logic in graphql though). I think we would need to run all of the diesel operation in spawn_blocking btw (can chat about my finding in our next meeting)

service/src/login.rs
clemens-msupply: Yes could do if we need clients in more services. But usually we don't pass in the service provider to the service so we would need to pass the "HTTPClientService" to the service...

graphql/general/src/queries/login.rs
clemens-msupply: The cookie is set to the gql context as part of the request response, so doesn't really fit into the service...

graphql/types/src/types/user.rs
clemens-msupply: Yes can do, but I defer it for now because i just noticed that the diesel type for that query is already quite complex :)

service/src/apis/permissions.rs
clemens-msupply: The whole file is more or less automatically converted from https://github.com/sussol/msupply/blob/499233688a61f283c7cbcaea06c45f2ddaf2b1ac/Project/Sources/Methods/user_permission_check.4dm#L402-L784 and I just took the comment over...

service/src/login.rs
clemens-msupply: Yeah Resource and permissions are just an enums though.

service/src/apis/login_v4.rs
clemens-msupply: Its not a translation is just a type def of what the api is giving me.

service/src/token.rs
clemens-msupply: I updated the jwt lib and they added a default leeway. This broke some tests and I needed to configure the leeway to fix the tests...

repository/src/db_diesel/user.rs
clemens-msupply: Plan to use it when enforcing the permissions

repository/src/db_diesel/user.rs
clemens-msupply: would work as well was just trying to follow other "similar" repos where we join things in the repo..

repository/src/db_diesel/user_permission.rs
clemens-msupply: yes don't think we need user_store_join for permission checking

graphql/general/src/queries/login.rs
clemens-msupply: [1ae3e70](https://github.com/openmsupply/remote-server/pull/986/commits/1ae3e7040a91e9ffee7b07efcf5d4e8ced85c4f3)

this actually removed the Option<Client> parameter because it wasn't needed for testing (the server is mocked not the client)

service/src/apis/permissions.rs
clemens-msupply: I had a script to extract them from 4d code... I was quite happy to have them all in one file when looking for matching permissions.

service/src/login.rs
clemens-msupply: thanks using map_err now hope that looks better now:
[686227a](https://github.com/openmsupply/remote-server/pull/986/commits/686227aa1894af28cced1c3fefd5555f07977f12)

graphql/general/src/queries/login.rs
clemens-msupply: I made it a structured error: [0feace3](https://github.com/openmsupply/remote-server/pull/986/commits/0feace3fb598048020b6538f5f06e6e26f94d90d)

repository/src/db_diesel/user.rs
clemens-msupply: I think so, at least I haven't managed to do it without them...

repository/src/schema/user_store_join.rs
```diff
@@ -0,0 +1,25 @@
+use super::diesel_schema::{store, user_account};
```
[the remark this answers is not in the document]
clemens-msupply: you mean because of the join macros? think the joins belong to the schema definition...

graphql/types/src/types/user.rs
clemens-msupply: Yes thanks [9d4d7d4](https://github.com/openmsupply/remote-server/pull/986/commits/9d4d7d4dddfb74cec67e360cd5f0df26248fedd4)

graphql/types/src/types/user.rs
clemens-msupply: done [22ad6f8](https://github.com/openmsupply/remote-server/pull/986/commits/22ad6f8a04eacbb9297a265bee44322f550045aa)

repository/src/db_diesel/user.rs
clemens-msupply: thanks, done in one of the commits

repository/src/db_diesel/user.rs
clemens-msupply: thanks [722aea3](https://github.com/openmsupply/remote-server/pull/986/commits/722aea3dd9c96d3452922ec25bb51e6d8a2063f3)

Rust should have anonymous structs instead of tuples :)

service/src/user_account.rs
clemens-msupply: This doesn't work for Postgres because PG aborts the whole transaction when there is an error, will add a comment about it.

service/src/login.rs
andreievg: ok, you don't think having an optional parameter used only for test is a bit hacky ?

repository/src/schema/user_store_join.rs
clemens-msupply: lets discuss tomorrow...

service/src/apis/login_v4.rs
andreievg: oki, i mean do we need those fields if we are not using them ? or does deserialisation break if those fields are missing ?

repository/src/db_diesel/user.rs
andreievg: yeah ! there is an enum variant like that, why not anonymous structs.

repository/src/schema/user_store_join.rs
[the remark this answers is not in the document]
andreievg: > you mean because of the join macros? think the joins belong to the schema definition...

To keep partitioning based on functional area rather the characteristic.

service/src/user_account.rs
andreievg: there is no error thrown, i might be missing something, i'll check

repository/src/db_diesel/user_permission.rs
andreievg: do we need it for anything else ?

clemens-msupply:
thanks for the review! I will merge now because I have to leave in a bit. Will reply to your latest comments tomorrow

service/src/user_account.rs
andreievg: ok, i remember now. 

What do you, this is predictable enough to query stores ? Or too much overhead for login action ?

clemens-msupply: That there are invalid stores seems to be a 4d problem, isn't it? I think doing it here where we handling the 4d input is ok.

service/src/apis/login_v4.rs
clemens-msupply: no, e.g. some blob objects are not deserialized. This api file has to be seen as a mini lib, i.e. its a stand alone file..

PR #989, opened 2022-03-31
NameByIdLoader: Fetch names for every store_id

clemens-msupply:
Prevents potential bug where names from different stores are requested.

I did this because I thought I ran into this bug but that wasn't the case. Thought the change would be useful anyway.

### DIFF
--- a/graphql/core/src/loader/name.rs
+++ b/graphql/core/src/loader/name.rs
@@ -37,26 +37,31 @@ impl Loader<NameByIdLoaderInput> for NameByIdLoader {
     ) -> Result<HashMap<NameByIdLoaderInput, Self::Value>, Self::Error> {
         let service_context = self.service_provider.context()?;
 
-        let store_id = if let Some(item_and_store_ids) = ids_with_store_id.first() {
-            &item_and_store_ids.primary_id
-        } else {
-            return Ok(HashMap::new());
-        };
-
-        let filter = NameFilter::new().id(EqualFilter::equal_any(IdPair::get_all_secondary_ids(
-            &ids_with_store_id,
-        )));
-
-        let names = self
-            .service_provider
-            .general_service
-            .get_names(&service_context, store_id, None, Some(filter), None)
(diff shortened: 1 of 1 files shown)


graphql/core/src/loader/name.rs
```diff
@@ -37,26 +37,31 @@ impl Loader<NameByIdLoaderInput> for NameByIdLoader {
...
-            .map_err(StandardGraphqlError::from_list_error)?;
-
-        Ok(names
-            .rows
-            .into_iter()
-            .map(|name| (NameByIdLoaderInput::new(store_id, &name.name_row.id), name))
-            .collect())
+        // store_id -> Vec of name_id
+        let mut store_name_map = HashMap::<String, Vec<String>>::new();
+        for item in ids_with_store_id {
```
andreievg, line 42: IdPair is pretty ambiguous data type, i tried to make it a bit cleaner with implementations, i think it's a good idea to add store_id and name_id getters to NameByIdLoaderInput, and use then here

graphql/core/src/loader/name.rs
```diff
@@ -37,26 +37,31 @@ impl Loader<NameByIdLoaderInput> for NameByIdLoader {
...
+        }
+        let mut output = HashMap::<NameByIdLoaderInput, Self::Value>::new();
+        for (store_id, names) in store_name_map {
+            let names = self
+                .service_provider
+                .general_service
+                .get_names(
+                    &service_context,
+                    &store_id,
+                    None, // TODO this needs to be ALL without limit
```
andreievg, line 56: yep, agree with another comment you've made in default context, I think we need a way to not use pagination in services getters. Either with another enum variant for service pagination input or do pagination validation in graphql

andreievg (approved):
Thanks, looks good

graphql/core/src/loader/name.rs
clemens-msupply: You mean to get the `HashMap<store_id, Vec<name_id>>`? yes can to in next iteration

PR #1007, opened 2022-04-05
Add some high level docs for the TokenService

clemens-msupply:
Thought I write down the motivating thoughts we had on why use the current auth token setup.

### DIFF
--- a/service/src/token.rs
+++ b/service/src/token.rs
@@ -104,8 +104,8 @@ pub struct TokenPair {
 /// Q: Still why an auth token? the long lived refresh token could be invalidated on the server
 /// site to revoke access.
 /// A: This requires a reliable mechanism to do the token invalidation. This can be hard to do. With
-/// the current solution the auth token simply expires when we stop issuing tokens to the 3rd party,
-/// e.g. by deactivated a plugin.
+/// the current solution the auth token simply expires when we stop issuing tokens to the 3rd party
+/// (which doesn't has the refresh token), e.g. by deactivated a plugin.
 pub struct TokenService<'a> {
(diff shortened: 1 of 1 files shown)


andreievg:
Thanks. I am wondering about our other overall docs, and how they fit with rust docs, would imagine we just link rust docs from overall docs.

andreievg (approved):

service/src/token.rs
```diff
@@ -79,6 +79,33 @@ pub struct TokenPair {
...
+/// attack can be active.
+///
+/// Q: Why an auth token? and why not just having one long lived session cookie?
+/// A: Having short lived auth tokens makes it possible to issue these tokens to a 3rd party to act
+/// on behave of a user, e.g. a plugin or an external service could use an auth token to gain
+/// temporary access to an API but is not able to refresh the token indefinitely.
+///
+/// Q: Still why an auth token? the long lived refresh token could be invalidated on the server
+/// site to revoke access.
+/// A: This requires a reliable mechanism to do the token invalidation. This can be hard to do. With
```
andreievg, line 106: Hmm not sure if i follow this. 
Refresh token would still need to be invalidated, and also the third party will still need to be given refresh token

clemens-msupply: I assumed the 3rd party wouldn't get the refresh token, e.g. a 3rd party would periodically get a new token from the "admin" instead...

clemens-msupply:
> Thanks. I am wondering about our other overall docs, and how they fit with rust docs, would imagine we just link rust docs from overall docs.

Yes was wondering this as well and this would probably should go in some other kind of docs. But I wanted to get it out as long as I was in a writing mood ;)

PR #1010, opened 2022-04-06
1009 add mock permission restriction for requisition mutation

### LINKED ISSUE #1009: Add mock permission restriction
Add a permission restriction for MutateRequisition that uses live data from central server

andreievg:
closes #1009 

Fixed permission mapping (it's a good idea to test central server UI configurations propagate correctly):
Screen Shot 2022-04-06 at 11 38 13 AM

Added a temp restriction for requisition mutate.

I did the following test:

* Log In, try mutate requisition with API (parameters don't matter, as long as strore_id is correct), you should get record not found error
* Update user permissions for the store_id you were using in central UI (see image above)
* Login again and try to mutate, should see permission error

I am not sure what front end would be showing as an error and if we need to present the error in a better format, would assume for V1 beta we would just show Unauthorised in UI

I think the error should be structure error, in the future when we sync users, user permissions could update without front end really knowing, and the plan was to return current user permission on authorisation error (so front end can update them).

Lastly, would assume permissions validations would be simplified quite a bit, i found it hard differentiation between PermissionServicTrait permissions and permissions in validation service. (I am sure there was a good reason to have them, but I don't think it would be needed to satisfy current requirements)

### DIFF
--- a/repository/src/schema/user_permission.rs
+++ b/repository/src/schema/user_permission.rs
@@ -1,7 +1,5 @@
 use diesel_derive_enum::DbEnum;
 
-use crate::EqualFilter;
-
 table! {
   user_permission (id) {
       id -> Text,
@@ -37,26 +35,4 @@ pub struct UserPermissionRow {
     pub store_id: Option<String>,
     pub resource: Resource,
     pub permission: Permission,
-}
-
-impl Resource {
-    pub fn equal_to(&self) -> EqualFilter<Resource> {
-        EqualFilter {
-            equal_to: Some(self.clone()),
-            not_equal_to: None,
-            equal_any: None,
-            not_equal_all: None,
-        }
-    }
-}
-
-impl Permission {
-    pub fn equal_to(&self) -> EqualFilter<Permission> {
-        EqualFilter {
-            equal_to: Some(self.clone()),
-            not_equal_to: None,
-            equal_any: None,
-            not_equal_all: None,
-        }
-    }
 }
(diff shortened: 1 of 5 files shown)


repository/src/schema/user_permission.rs
```diff
@@ -36,3 +38,25 @@ pub struct UserPermissionRow {
...
+
+impl Permission {
+    pub fn equal_to(&self) -> EqualFilter<Permission> {
+        EqualFilter {
+            equal_to: Some(self.clone()),
+            not_equal_to: None,
+            equal_any: None,
+            not_equal_all: None,
+        }
+    }
```
clemens-msupply, line 61: Should these filter related impl go into the query repository where the filter is used?

service/src/apis/permissions.rs
```diff
@@ -401,7 +401,7 @@ pub fn map_api_permissions(permissions: Vec<bool>) -> Vec<Permissions> {
         if !has_per {
             continue;
         }
-        if let Some(permission) = mapping.get(&(i as i16)) {
+        if let Some(permission) = mapping.get(&((i + 1) as i16)) {
```
clemens-msupply, line 404: thanks! good catch

clemens-msupply (approved):
Looks good!

PermissionServiceTrait might be useful in /me endpoint to retrieve user permissions but direct repo access might work as well. (No strong reason to hang on to it)

repository/src/schema/user_permission.rs
andreievg: yeah sorry, I was in the quick hack mode: e45bb31f6649780e8112645a8ad3b2b6bbaaa9f9

Issue #1006, opened 2022-04-05
Login speed


mark-prins:
The `authToken` request is pretty slow, though only if the username is correct. This gives a little bit of security information away - as an attacker can determine if a username is valid by looking at how slow the query responds. As an example:

https://user-images.githubusercontent.com/9192912/161855347-a5be2ec5-8d95-4047-b561-f327228e6df8.mp4

Would it be possible to
- speed up the `authToken` query, or
- slow down the `authToken` when there's an invalid username? or is that too perverse and unnecessary?

Also, can you remove the `registerUser` mutation? that's no longer useful, as we cannot login with a registered user, they won't have any stores assigned.

jmbrunskill:
Love the security conscious thinking @mark-prins !

clemens-msupply:
thanks good point, to mitigate this I can delay the login response if there is an error, i.e. always takes ~5sec regardless if wrong username or wrong password

mark-prins:
if you know what the slow part is - perhaps just run that even if the username is wrong? that way the response if always (roughly) the same elapsed time

clemens-msupply:
> if you know what the slow part is - perhaps just run that even if the username is wrong? that way the response if always (roughly) the same elapsed time

could get tricky because the password hash cost parameter is encoded in the stored hashed_password and there could be other factors that affect response time... but yes would be an option

andreievg:
if it's the password hash (not the token signature), that's slow, can we not just hash password all the time and query for user from user table with both password and username ?

clemens-msupply:
> if it's the password hash (not the token signature), that's slow, can we not just hash password all the time and query for user from user table with both password and username ?

In practice probably yes. However, one thing that could happen is that passwords could be hashed with different cost factors so if you guess a correct username that uses a different cost factor than the default cost factor you would observe a time difference in the response time...

andreievg:
> In practice probably yes. However, one thing that could happen is that passwords could be hashed with different cost factors so if you guess a correct username that uses a different cost factor than the default cost factor you would observe a time difference in the response time...

Hmm, good point, and calculating hash all the time (even if username is incorrect), seems less optimal. I think the cost factor should be constant (if we ever decide to change it, we can re-save new password hash during successful login). If cost is the same and salt is the same length I think we should be able to calculate aprox calculation time on server startup and use it as the delay

PR #1011, opened 2022-04-06
Use fix login response time in case of a login error

clemens-msupply:
If there is a login error I wait till (the randomly chosen time of) 6s for the total response has been passed.

#1006 

Needed to add tokio as a dependency for an async sleep...


andreievg:
https://github.com/openmsupply/remote-server/issues/1006

thanks

andreievg (approved):

PR #1012, opened 2022-04-06
Remove register_user endpoint

### LINKED ISSUE #1006: Login speed

clemens-msupply:
Endpoint was unprotected... Service method is still in place.

Closes #1006


andreievg (approved):
Looks good thanks, i think just removing the GeneralMutations might have been enough

PR #1019, opened 2022-04-11
Add feature to run tests in sqlite memory mode

clemens-msupply:
Adds a new feature to run test "in memory"

Some results for the test runtime:
postgres: 1m39.591s, 1m52.946s
sqlite: 1m28.498s, 1m23.752s
memory: 18.924s, 18.539s

### DIFF
--- a/repository/src/database_settings.rs
+++ b/repository/src/database_settings.rs
@@ -12,6 +12,7 @@ pub struct DatabaseSettings {
     pub database_name: String,
 }
 
+// feature postgres
 #[cfg(feature = "postgres")]
 impl DatabaseSettings {
     pub fn connection_string(&self) -> String {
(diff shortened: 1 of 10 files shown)


repository/src/database_settings.rs
```diff
@@ -29,14 +30,27 @@ impl DatabaseSettings {
     }
 }
 
-#[cfg(not(feature = "postgres"))]
+// feature sqlite
+#[cfg(all(not(feature = "postgres"), not(feature = "memory")))]
```
mark-prins, line 34: is this to account for the default case? Was wondering why it isn't `feature = "sqlite"` and why the implicit assumption that `not(feature = "postgres") === (feature = "sqlite")` 😃

service/src/permission_validation.rs
```diff
@@ -416,11 +416,11 @@ mod permission_validation_test {
         }
 
         fn store() -> StoreRow {
-            StoreRow {
-                id: "store".to_string(),
-                name_id: name().id,
-                code: "n/a".to_string(),
-            }
+            inline_init(|s: &mut StoreRow| {
```
mark-prins, line 419: just checking.. this was in the debug PR too

mark-prins (approved):
👍 that looks good to me. 
I've been running the tests with `--features memory` as well as `--features postgres` and sqlite and they run.. and are a little faster 😁

mark-prins:
Side note: having trouble running postgres tests. Is that to be expected?

`FATAL:  sorry, too many clients already` when running posgres 12

service/src/permission_validation.rs
clemens-msupply: Yes its a duplicated commit. Since its so tiny I did it twice and make the git hub action happy...

clemens-msupply:
> Side note: having trouble running postgres tests. Is that to be expected?
> 
> `FATAL: sorry, too many clients already` when running posgres 12

No sorry didn't ran into this before. Rust test are executed in parallel (probably directly correlated to the number of cpu cores). How many do you have? still strange would have expected postgres would allow "lots of clients" on default.

mark-prins:
> How many do you have? 

8 cores - though I agree, it's a little strange. Perhaps due to the postgres server I'm running 🤷

Issue #887, opened 2022-03-06
Report PDF printing


clemens-msupply:
Goal: Api user should be able to print a selected report as a PDF. In the future report printing should also be possible as an automated task on the server (i.e. data fetching, document generation and PDF printing needs all be done on the server).

Printing templates
- Tera templates to specify a document to be printed
- Tera files are stored in the report table on the central server (editor column is used to specify the report type)
- Other resources like logos, css, could possibly stored in this table as well

Data queries
- default queries: e.g. to get the full invoice data
- custom query: be able to specify a custom GQL query (also stored in the report table)
- the query should be evaluated with the permissions of the requesting user. Doing a self request with the Bearer token of the user is probably the easiest way to achieve this

PDF printing
- Use headless chrome to generate a PDF
- TODO: does it work on Android?

Tasks:
- [ ] pull report central records
- [ ] define new editor type(s) (e.g. json blob that holds all data, containing, tera file, query, images, link to other report data records, 
etc..)
- [ ] service for printing a report
- [ ] endpoint for printing a report

andreievg:
- [ ] https://github.com/openmsupply/remote-server/issues/899

mark-prins:
The mock endpoint (#899) isn't filtering by category ( if you specify a category filter you get all report categories returned ).
Commenting so that we remember to test when this goes into PR

PR #1023, opened 2022-04-12
Remove hard coded template keys and use a template index map instead

clemens-msupply:
Add an index to the report definition to make it clearer what entry in the definition file should be used for which purpose.

#887

### DIFF
--- a/service/src/report/definition.rs
+++ b/service/src/report/definition.rs
@@ -71,13 +71,18 @@ pub enum ReportDefinitionEntry {
     Ref(ReportRef),
 }
 
+/// Specifies which report definition entries are the "main" entries.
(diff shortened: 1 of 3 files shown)


mark-prins (approved):
👍  working nicely.

One concern is that existing reports have to be migrated - is there a way to do that easily? I just deleted the database locally, but that's not so easy on the demo server. Thinking that I'll just copy the templates from my local database. is that the best?

clemens-msupply:
> +1 working nicely.
> 
> One concern is that existing reports have to be migrated - is there a way to do that easily? I just deleted the database locally, but that's not so easy on the demo server. Thinking that I'll just copy the templates from my local database. is that the best?

No sorry no easy way to migrate. Copying from the local db sounds good.

PR #1026, opened 2022-04-13
971 enforce permissions

### LINKED ISSUE #971: Not all queries are checking auth token
Most queries don't require an auth token, and I understood that all queries & mutations should now?

I've only found requisitions & stocktakes which requires the auth token, and the mutation to allocate placeholder stock.
These ones require the token:

- requisition
- requisitionByNumber
- requisitions
- stocktake
- stocktakeByNumber
- stocktakes

haven't tested all the mutations yet.. that's harder than just querying 😉

clemens-msupply:
Add permission checks to all endpoints.

Also rework how permissions are stored on the remote-server, i.e. a user can have a simple list of permissions.

Closes #971


mark-prins (approved):
😮  wow! that's a ton of work - awesome to have in place! thanks heaps.

Have tested a few, and they're working nicely. Didn't quite have the patience to test them all! did read through carefully and didn't spot anything amiss.

One thought - it will require some string parsing to interpret the response and display suggested actions to the user. Mostly we would interpret based on the action taken or location of the error, but I wonder if it would be possible to return a more structured response in some way? An array of permissions required? 🤷 

Generally, the highly structured responses require a fair bit of work to parse.. so I'm a little reluctant to request them! 😆  

```
            "extensions": {
                "details": "Required permissions: And([HasStoreAccess, HasPermission(StockLineQuery)])"
            }
```

clemens-msupply:
thanks for the review, plan to do some automated permission testing (if I don't run out of time)

Yes a structured error would be a bit of a pain on the backend as well. Should be no problem to include the required permission as a json blob with a predefined format though. However, maybe it would be better to provide an endpoint to query the endpoint permissions. This could be used to figure out if a user can do a request in the first place and if there is an error you could figure out the missing permission as well. This could also be useful for the permission UI.

PR #1028, opened 2022-04-18
887 report generation cli

clemens-msupply:
This PR contains:
- `report_builder` cli tool to build a template definition and to print a report (see readme for details)
- An example project for building a template definition
- New endpoint to print a report from a provided template (while testing/designing the template, this avoids the round trip to upload a template definition to the central server and sync the template to the remote server)
- On the remote-server, add all resources to Tera to make them available in a "Tera way"
- Some remote-server template generation refactoring

#887


report_builder/src/print.rs
```diff
@@ -0,0 +1,197 @@
...
+"#;
+
+#[derive(Debug, Serialize, Deserialize)]
+struct Config {
+    url: String,
+    username: String,
+    password: String,
+}
+
+#[allow(dead_code)]
```
mark-prins, line 53: I've removed and built ok - is this required?

report_builder/src/print.rs
```diff
@@ -0,0 +1,197 @@
...
+    Ok(file_id)
+}
+
+fn fetch_file(
+    url: Url,
+    token: &str,
+    file_id: &str,
+    name: &Option<String>,
+) -> anyhow::Result<String> {
+    let result = reqwest::blocking::Client::new()
```
mark-prins, line 125: this still bothers me 😆  every time I see it, I see a typo!

mark-prins (approved):
😁  wow! super useful - nicely documented and easy to use. worked really well for me

would be great to get it in, and the demo server updated, to make it easier to run against the demo data. Perhaps we could also build and share the binary for template developers!

just have some nit picky comments.. basically it all worked fine for me 😄

report_builder/README.md
```diff
@@ -0,0 +1,95 @@
...
+## Usage
+To build the report builder from the Rust source code run the following command in the `report_builder` directory:
+```bash
+> cargo build
+# or 
+> cargo build --release
+```
+Alternatively you can also just run it from the project dir
+```bash
+> cargo run -- {builder args go here}
```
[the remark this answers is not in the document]
clemens-msupply: Thanks I refer to the `--help` argument now so that the readme is less likely to become outdated.

report_builder/src/print.rs
clemens-msupply: The `errors` in the struct are not used directly and then gives a warning... The errors are printed to the command line though which is helpful if there is an error in the query.

so it just to ignore the warnings

clemens-msupply:
Thanks for the review! I applied the suggested changes

PR #1032, opened 2022-04-19
Always filter locations by store_id

### LINKED ISSUE #1027: Locations: not using store_id
The `locations` query is returning locations for all stores, regardless of which `store_id` is supplied.
This results in:

image

when trying to delete.

The list of locations returned should be specific to the store the user is currently in.

clemens-msupply:
Closes #1027


mark-prins (approved):
👍  that works nicely - should prevent the user confusion with editing other store's locations 😁

PR #1041, opened 2022-04-22
Always filter on store id

### LINKED ISSUE #1040: master_lists, items, invoice_counts and stock_count queries don't make use of the store_id
For example, the master_list endpoint takes a store_id parameter but doesn't filter master_list by this store_id. The endpoint should only return lists for the provided store_id. The MasterListFilterInput filter already has a `exists_for_store_id` option that can be used to filter for the desired lists.

GraphQl entry point for master_list: `graphql/general/src/queries/master_list.rs` (https://github.com/openmsupply/remote-server/pull/1032)

Similar changes have been done recently for the location endpoint: `graphql/location/src/lib.rs` ()

- [x] master_list
- [x] invoice_counts
- [x] stock_count

jmbrunskill:
Fixes #1040 
Using existing mock data in tests at this stage.


graphql/general/src/queries/master_list.rs
```diff
@@ -100,19 +100,26 @@ pub fn master_lists(
...
 
     let service_provider = ctx.service_provider();
     let service_context = service_provider.context()?;
 
+    // always filter by store_id
+    let mut query_filter = MasterListFilter::new();
+    if let Some(filter_input) = filter {
+        query_filter = filter_input.to_domain()
+    }
+    query_filter = query_filter.exists_for_store_id(EqualFilter::equal_to(&store_id));
```
jmbrunskill, line 115: @clemens-msupply Is there a risk here that this filter could already be set? Would this be a problem?

graphql/general/src/queries/tests/master_lists.rs
```diff
@@ -236,4 +237,62 @@ mod graphql {
...
+        //     .filter(|v| v.store_id == "store_a")
+        //     .count();
+        // let count_store_b = mock_locations()
+        //     .iter()
+        //     .filter(|v| v.store_id == "store_b")
+        //     .count();
+        // assert!(count_store_a != count_store_b);
+
+        //TODO: Fix mocks
+        let count_store_a = 1;
```
jmbrunskill, line 261: @clemens-msupply Is hard coding some values here based on existing mocks ok, or do you think I need to create new mock datasets?

graphql/general/src/queries/master_list.rs
clemens-msupply: No thats not a problem, `exists_for_store_id` can be called multiple times.

graphql/general/src/queries/tests/master_lists.rs
clemens-msupply: think its better to get them from the mock data, e.g. in case we add more master_lists to the mock data base set. I had this `assert!(count_store_a != count_store_b)` in to verify that the counts we are getting from the mocks are different for different stores.

However, when I left on Fri I realized that by fixing the test that broke we already did all the testing in the gql layer, i.e. make sure that the filter is correctly set. The actual test should better be moved into the master_list service (if it hasn't been already tested there) Sorry, for putting you on the wrong lead with my test example.

mark-prins:
I've added the issue reference in the PR description - that's been the convention, though it appears that GitHub figured out the relationship anyway 😮

jmbrunskill:
> I've added the issue reference in the PR description - that's been the convention we've been following, though it appears that GitHub figured out the relationship anyway 😮
Ok cool. So the text 'fixes ####' in the PR?

mark-prins:
@jmbrunskill yeah, just the `Fixes #[issue number]`

I've seen Andrei using `closes` instead.. and I see that [many options](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) are available! 😁

service/src/dashboard/invoice_count.rs
```diff
@@ -33,6 +33,7 @@ pub trait InvoiceCountServiceTrait: Send + Sync {
         range: &CountTimeRange,
         now: &DateTime<Utc>,
         timezone_offset: &FixedOffset,
+        store_id: &str,
```
clemens-msupply, line 36: We usually have the store_id as one of the first parameters in the other endpoints, i.e. just after ctx...

service/src/dashboard/stock_expiry_count.rs
```diff
@@ -11,8 +11,9 @@ pub trait StockExpiryCountServiceTrait: Send + Sync {
         &self,
         ctx: &ServiceContext,
         date_time: NaiveDate,
+        store_id: &str,
     ) -> Result<i64, RepositoryError> {
-        StockExpiryServiceCount {}.count_expired_stock(ctx, date_time)
+        StockExpiryServiceCount {}.count_expired_stock(ctx, date_time, store_id)
```
clemens-msupply, line 16: move store_id after ctx

service/src/master_list/tests/query.rs
```diff
@@ -45,5 +45,30 @@ mod query {
...
+            .get_master_lists(
+                &context,
+                None,
+                Some(MasterListFilter::new().exists_for_store_id(EqualFilter::equal_to("store_a"))),
+                None,
+            )
+            .unwrap();
+        assert!(result.count >= 1);
+
+        //Test filter for non existant store finds nothing
```
clemens-msupply, line 60: spelling existent

clemens-msupply (approved):
Looks good just minor things, see comments

clemens-msupply:
> @jmbrunskill yeah, just the `Fixes #[issue number]`
> 
> I've seen Andrei using `closes` instead.. and I see that [many options](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) are available! grin

"Closes" closes the issue automatically though which can sometimes be a bit annoying...

mark-prins:
actually, any of the keywords will close the issue. it's the linking which closes on merge and any of those words will link the PR with the issue - though I have had some issues that were linked and they didn't close. Not sure why not, and couldn't see any mention on that page. most often it is when there are multiple issues linked.. it's been such a minor thing that I haven't investigated, just wondered wistfully sometimes... 💭

Issue #1038, opened 2022-04-20
Rename repositories (i.e. when we have [Type]Query to [Type] and [Type]Row


clemens-msupply:
We have two types of Repositories:
1) Raw row access repositories
2) More domain specific repos, e.g. doing joins or special queries

Currently these repositories are named inconsistently. The desired naming is, for example, for name rows:
1) NameRowRepository
2) NameRepository

Repositories are located in:
`repository/src/db_diesel`

PR #1042, opened 2022-04-26
1038 rename repositories (i.e. when we have type]query to type] and type]row

roxy-dao:
#1038 

- Renamed invoice, name, item, and changelog repositories for consistency


clemens-msupply (approved):
Great thanks! that where many more files involved than expected.

PR #1044, opened 2022-04-26
1015 settings service

### LINKED ISSUE #1015: Manage server settings and server bootstrapping
(To be discussed):
Motivation: provide an UI to let the user enter details such as central server url and side id, e.g. for the initial startup of the Android app.

### General settings management (during normal operation)
- [x] Store server settings (apart from DB connection settings) in the DB
- [x] Option to start the server using the current config file settings (in case settings in the DB are invalid and the server becomes unreachable)
- [x] Endpoint to retrieve server settings from the DB (including currently used settings if differ from the settings in the DB). 
- [x] Endpoint to modify server settings in the DB
- [x] Endpoint to restart the server, e.g. to use the new settings

### Create settings on first startup
- [x] If no settings are available on startup (or settings are incomplete), boot the server into a "stage 0" mode where only endpoints to manage the server settings are provided.
- [x] Relax user permission validation in this mode, e.g. if there is no user yet
- [x] Endpoint to fetch server status, e.g. to determine if server is running in "stage 0" mode

clemens-msupply:
This PR allows users to change sync settings at runtime and to restart the server to use the new settings.

Currently the settings from the config file have precedence. However, when no settings are found the server starts into a "stage0" bootstrap mode which only provides endpoints to update the server settings and to restart the server.

Open issues:
- store other settings in the DB as well?
- Add cli option to specify config files and if not specified run server from settings from the DB.

Closes #1015


configuration/base.yaml
```diff
@@ -4,8 +4,8 @@ server:
 sync:
   url: "http://localhost:2048"
   username: "username"
-  password: "password"
-  interval: 300
+  password_sha256: "hashed_password"
+  interval_sec: 300
```
mark-prins, line 8: perhaps `interval_seconds`?
It's interesting to have `interval` in full, and `seconds` abbreviated 🤷

configuration/base.yaml
```diff
@@ -4,8 +4,8 @@ server:
...
   central_server_site_id: 1
```
mark-prins, line 9: I wonder if the whole sync settings should be commented out in the `base.yaml` otherwise it is hard to set the hashed password 😉

mark-prins:
@clemens-msupply I'm getting a little stuck testing this one! How do I update the settings when in bootstrap mode - it requires being logged in ( the mutation requires an auth header ), but when in bootstrap mode there is no option to authenticate!

I've had to enable `debug_no_access_control` to use the mutation. but that leaves the setting enabled, which I do not want

graphql/general/src/queries/server_settings.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    pub async fn central_server_site_id(&self) -> u32 {
+        self.settings.central_server_site_id
+    }
+
+    pub async fn site_id(&self) -> u32 {
+        self.settings.site_id
+    }
+
+    pub async fn site_hardware_id(&self) -> String {
```
mark-prins, line 52: I'm wondering if hardware id should be exposed in this way. the intention is to limit which clients can sync with mSupply and typically the mac address is used. I wonder if it should just be set to the hardware mac address automatically?

graphql/general/src/queries/server_settings.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+#[Object]
+impl ServerSettingsNode {
+    async fn status(&self) -> ServerStatus {
+        self.status
+    }
+
+    /// Currently used sync settings (may differ from what is stored in the DB)
+    async fn sync_settings(&self, ctx: &Context<'_>) -> Option<SyncSettingsNode> {
```
mark-prins, line 70: does db equate to config file? I'm really not clear on the difference between these two sets of settings

graphql/general/src/queries/server_settings.rs
```diff
@@ -0,0 +1,151 @@
...
+    Ok(ServerSettingsResponse::Response(
+        ServerSettingsNode::from_domain(sync_settings, stage0),
+    ))
+}
+
+impl ServerSettingsNode {
+    pub fn from_domain(sync_settings: Option<SyncSettings>, stage0: bool) -> Self {
+        ServerSettingsNode {
+            status: match stage0 {
+                true => ServerStatus::Stage0,
```
mark-prins, line 127: are there any thoughts about extending this and being able to query sync status? as in, to know if the server is synchronising? Useful to know after the stage0 process, as the server takes a little time to sync, and the site will be unavailable until fully synchronised.

Side thought - am now wondering about triggering sync manually, which is possible with mobile currently. We could fire off a request to sync and then monitor progress?

server/src/lib.rs
```diff
@@ -26,34 +37,119 @@ pub mod static_files;
...
-pub async fn start_server(
+async fn run_stage0(
     settings: Settings,
-    mut off_switch: oneshot::Receiver<()>,
-) -> std::io::Result<()> {
-    let cert_type = find_certs();
+    off_switch: Arc<Mutex<oneshot::Receiver<()>>>,
+    connection_manager: StorageConnectionManager,
+) -> std::io::Result<bool> {
+    info!("Start server in bootstrap mode. Please use API to configure the server.");
```
mark-prins, line 45: :nit the phrase `Start server..` sounds like a command.. i.e. that I should start the server 😁 

I wondered about `Starting server in bootstrap..` or `Bootstrap mode. Please...` or `Server started in bootstrap mode...` 🤷

server/src/lib.rs
```diff
@@ -26,34 +37,119 @@ pub mod static_files;
...
+        ServerCertType::None => {
+            if !settings.server.develop {
+                error!("No certificates found");
+                return Err(std::io::Error::new(
+                    ErrorKind::Other,
+                    "Certificate required in production",
+                ));
+            }
+
+            warn!("No certificates found: Run in HTTP development mode");
```
mark-prins, line 98: :nit similar to my pondering on `Start server` - this one could be `Running in HTTP development mode`

server/src/lib.rs
```diff
@@ -117,18 +217,70 @@ pub async fn start_server(
...
+        Ok(_) => info!("DB migrations succeeded"),
+        Err(err) => {
+            let msg = format!("Failed to run DB migrations: {}", err);
+            error!("{}", msg);
+            panic!("{}", msg);
+        }
+    };
+
+    // allow the off_switch to be passed around during multiple server stages
+    let off_switch = Arc::new(Mutex::new(off_switch));
```
mark-prins, line 261: that lead me to wonder what happens when you spam the restart query 😉  Not much as it happens, just a 'failed to fetch' after a bit.

service/src/permission_validation.rs
```diff
@@ -56,12 +56,19 @@ pub enum Resource {
...
     Report,
+    // view/edit server setting
+    ServerAdmin,
 }
 
 fn all_permissions() -> HashMap<Resource, PermissionDSL> {
     let mut map = HashMap::new();
     // me: No permission needed
 
+    map.insert(
```
mark-prins, line 67: oh! nice! 😁

mark-prins (approved):
🎉  squeeee!! that's super exciting!

am super keen to put in a settings section in the Admin area on the site now 😁  Will have a look tomorrow!

Just have some questions around usability with the bootstrap mode, as the authentication is proving to be an issue for me. Doesn't affect configuration file usage though.. so we should be ok with merging this and updating the config file on the demo server.

clemens-msupply:
> @clemens-msupply I'm getting a little stuck testing this one! How do I update the settings when in bootstrap mode - it requires being logged in ( the mutation requires an auth header ), but when in bootstrap mode there is no option to authenticate!
> 
> I've had to enable `debug_no_access_control` to use the mutation. but that leaves the setting enabled, which I do not want.
> 
> ```
> [2022-04-26T05:34:40Z INFO  server] Start server in bootstrap mode. Please use API to configure the server.
> [2022-04-26T05:34:40Z WARN  server] No certificates found: Run in HTTP development mode
> [2022-04-26T05:34:40Z INFO  actix_server::builder] Starting 10 workers
> [2022-04-26T05:34:40Z INFO  actix_server::server] Actix runtime found; starting in Actix runtime
> [2022-04-26T05:34:47Z INFO  async_graphql::extensions::logger] [Execute] query serverRestart { serverRestart { message } }mutation updateServerSettings { updateServerSettings(input: {syncSettings: {url: "http://localhost:2020", username: "remote", password: "pass", intervalSec: 300, centralServerSiteId: 1, siteId: 2, siteHardwareId: ""}}) { ... on ServerSettingsNode { __typename status syncSettings { centralServerSiteId username url siteId siteHardwareId intervalSec } syncSettingsDb { centralServerSiteId intervalSec siteHardwareId siteId username url } } } }
> [2022-04-26T05:34:47Z INFO  graphql] [Execute Response] Some("updateServerSettings")
>     Response { data: Null, extensions: {}, cache_control: CacheControl { public: true, max_age: 0 }, errors: [ServerError { message: "Unauthenticated", locations: [Pos(2:3)], path: [Field("updateServerSettings")], extensions: Some(ErrorExtensionValues({"details": String("NotAuthenticated(\"Missing auth token\")")})) }], http_headers: {} }
> ```

Thanks for testing it! the access control was not configures correctly. [627a7ab](https://github.com/openmsupply/remote-server/pull/1044/commits/627a7abd40172b09ebc71baf09a31e5ca5c2b98a)

configuration/base.yaml
clemens-msupply: good point but think sec is common enough? think seconds becomes a bit to prominent and it reads a little bit like "interval-seconds" instead "interval in seconds" (?)

graphql/general/src/queries/server_settings.rs
clemens-msupply: Makes sense, would this better be managed on the FE site, e.g. would it still make sense to set a "random" hardware id through the API but FE always picks the MAC address?

graphql/general/src/queries/server_settings.rs
clemens-msupply: The idea was to show the settings as they are currently used by the server, e.g. could come from DB or config file. You could change the DB settings at runtime in which case they differ and I thought this might be useful for the FE (?), e.g. to show if a restart is needed.

configuration/base.yaml
clemens-msupply: Yes agree. How the config files are used need some thoughts anyway but think its better to leave it till a later point because it requires some other changes as well, e.g. to have cli args to overwrite settings...

server/src/lib.rs
clemens-msupply: thanks [386624a](https://github.com/openmsupply/remote-server/pull/1044/commits/386624a9a04f532868d0b7dd1446284683ff54db)

server/src/lib.rs
clemens-msupply: The restart switch channel is configure to size 1, i.e. only one message is allowed in the queue. Even if there would be multiple calls I don't think its a problem because the errors should be handled gracefully (untested)...

clemens-msupply:
Thanks for your review will merge now and will look at your issue about invalid settings next.

graphql/general/src/queries/server_settings.rs
mark-prins: the web client isn't able to determine the MAC address of the server though, not even the local machine's MAC address! So the f/e could only ever send a random string which I think isn't very helpful. we want something that will be consistent and persist outside of the scope of either b/e or f/e

in mSupply you can now only reset the hardware address for remote sites which sync to the current instance - previously you could specify an address. [new docs](https://docs.msupply.org.nz/synchronisation:sync_sites?s[]=hardware#hardware_id)

personally, I can't see a reason for setting the id via the f/e.. I might be misunderstanding the intention of the id though

configuration/base.yaml
mark-prins: 🤣  `interval_in_seconds` 
perhaps should be `interval_secs` to be strictly correct.. and helps disambiguate from `interval_secondary` and `interval_security` etc.

server/src/lib.rs
mark-prins: nice! it appeared gracefully handled to me

graphql/general/src/queries/server_settings.rs
clemens-msupply: good idea haven't thought about it yet though, wonder if we could merge it together with api version, etc...

PR #1048, opened 2022-04-26
Falling back to bootstrap mode if the initial sync failed

clemens-msupply:
Instead of terminating when the initial sync failed restart into bootstrap mode.

To make this work the http server need to be bound to an ip address after the initial sync otherwise we get an addr in use error. This makes the PR look bigger than it is.

https://github.com/openmsupply/remote-server/issues/1046


mark-prins (approved):
👍  awesome! that's working nicely

PR #1059, opened 2022-04-28
1058 server settings fixes

### LINKED ISSUE #1058: Server settings issues
- [ ] the token bucket with existing tokens should be kept across restarts otherwise users have to re-authenticate
- [ ] when starting with settings from the config these settings are always preferred over the DB settings and restart doesn't use the new settings -> always use DB settings if present and it was a restart for now

clemens-msupply:
- Users keep logged in after restart
- After restart the DB settings are preferred

Closes #1058

### DIFF
--- a/graphql/core/src/test_helpers.rs
+++ b/graphql/core/src/test_helpers.rs
@@ -1,4 +1,4 @@
-use std::sync::RwLock;
+use std::sync::{Arc, RwLock};
 
 use actix_web::{
     guard,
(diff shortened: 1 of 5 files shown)


mark-prins (approved):
👍  works well! 
Can't comment on the use of `Arc` though I can say that it works. and is very nice to stay logged in 😁

clemens-msupply:
Arc is an Atomic Reference Counter. Its a smart pointer which allows us to have multiple references to the containing data, it only deletes the containing data once the count is down to 0, i.e. when nobody needs it anymore.  The atomic is needed because the token_bucket is passed across threads, i.e. the count needs to be mutated in an atomic operation.

PR #1062, opened 2022-05-02
Generate a UUID based token secret on startup if doesn't exist

### LINKED ISSUE #1047: Init a strong token secret at first server startup
The token_secret is used to sign/verify the Bearer auth tokens.

Currently the token_secret in the AuthSettings comes from the config file. This can be dangerous, i.e. if the default value is not updated during server initialization or a weak secret is chosen.

Solution: remove default secret and generate a strong secret at startup (store it in the key value store)

Make AuthSettings optional in the config file (or remove it?);

jmbrunskill:
Closes #1047


service/src/settings_service.rs
```diff
@@ -99,6 +100,33 @@ pub trait SettingsServiceTrait: Sync + Send {
             .map_err(|err| UpdateSettingsError::RepositoryError(err.to_inner_error()))?;
         Ok(result)
     }
+
+    /// Loads auth settings from the DB
+    fn token_secret(&self, ctx: &ServiceContext) -> Result<String, UpdateSettingsError> {
```
jmbrunskill, line 105: I wondered if this should be called 'get_or_create_token_secret' or similar to be more explicit?

server/src/lib.rs
```diff
@@ -153,8 +153,13 @@ async fn run_server(
     };
 
     let cert_type = find_certs();
+    let token_secret: String = if prefer_config_settings {
```
jmbrunskill, line 156: prefer_config_settings always seems to false on my local. Tested by hard coding to true.
@clemens-msupply is this expected behavour?
Should we just always used token secret from KV and remove it from yaml entirely?

repository/src/schema/key_value_store.rs
```diff
@@ -19,6 +19,7 @@ pub enum KeyValueType {
     SettingsSyncCentralServerSiteId,
     SettingsSyncSideId,
     SettingsSyncSideHardwareId,
+    SettingsTokenSecret,
```
clemens-msupply, line 22: You also need to update the DB constraints in the postgres SQL migration file. We plan to generate some docs from the postgres schema and to reduce dev overhead I didn't add it to sqlite schema (but maybe we should, because it becomes confusing?).

Maybe rename to SettingsAuthTokenSecret because its in the AuthSettings...

server/src/lib.rs
clemens-msupply: Yes prefer_config_settings is always true at startup right now. Only after a restart the DB settings are preferred. This was for the recent sync settings endpoints and probably need to be revised...

Yes I think the AuthSettings can safely be removed with this PR.

service/src/settings_service.rs
```diff
@@ -99,6 +100,31 @@ pub trait SettingsServiceTrait: Sync + Send {
...
+                .connection
+                .transaction_sync(|con| {
+                    let key_value_store = KeyValueStoreRepository::new(con);
+                    key_value_store
+                        .set_string(KeyValueType::SettingsTokenSecret, token_secret.clone())?;
+                    Ok(())
+                })
+                .map_err(|err| UpdateSettingsError::RepositoryError(err.to_inner_error()))?;
+        }
+        Ok(token_secret.unwrap())
```
clemens-msupply, line 126: To avoid the unwrap() you can do:

```rust
let token_secret = match token_secret {
    Some(token_secret) => token_secret,
    None => { 
        let token_secret = uuid();
        // insert...
        token_secret
    }
}

Ok(token_secret)

File: service/src/settings_service.rs [hunk 5]
```diff
@@ -99,6 +100,31 @@ pub trait SettingsServiceTrait: Sync + Send {
...
+            .get_string(KeyValueType::SettingsTokenSecret)
+            .map_err(|err| UpdateSettingsError::RepositoryError(err))?;
+
+        if token_secret.is_none() {
+            //There is no token set, so we want to generate a secure one and store it for next startup
+            token_secret = Some(uuid());
+            //store new token in Key Value Store so it doesn't change on next startup
+            let _result = ctx
+                .connection
+                .transaction_sync(|con| {
```
clemens-msupply, line 118: The transaction is not needed here, you can call the Repo directly with the ctx.connection...

service/src/settings_service.rs
clemens-msupply: get_* prefix are not common in Rust maybe just add it the function docs

service/src/settings_service.rs
jmbrunskill: Yes, that's a much more 'Rust' way to do it. I'll update.

service/src/settings_service.rs
jmbrunskill: Ok right, was copy and pasting but this is a good point. I'll do some more reading to understand and update.

repository/src/schema/key_value_store.rs
jmbrunskill: Ok didn't look at this. What's the value of the check constraints? Do they just add extra dev overhead with little value?
I wonder if we should consider removing constraints from both, it is a KV store after all!

server/src/lib.rs
jmbrunskill: Is this even true one second start up? Does a restart somehow get triggered? I'm slightly confused by the mechanism but doesn't matter for this PR.

Thanks for the great comments!

repository/src/schema/key_value_store.rs
clemens-msupply: I was arguing like this as well but we also wanted to have the schema fully self documented. The compromise was to only add the constraint to one DB file...

server/src/lib.rs
clemens-msupply: agree, yes needs to be revised. My thinking was the use case (which we don't support yet) that you explicitly overwrite the DB settings by providing the settings file, e.g. to fix doggy db settings. From then you may change the settings in the DB and when restarting you want the new settings and not the initial settings from the config...

service/src/settings_service.rs
```diff
@@ -101,29 +101,28 @@ pub trait SettingsServiceTrait: Sync + Send {
...
-        }
-        Ok(token_secret.unwrap())
+        let token_secret = match token_secret {
+            Some(token_secret) => token_secret,
+            None => {
+                let token_secret = uuid();
+                //Result is ignored here, as server can operate with the current secret (will change on next startup though)
+                let _result = key_value_store.set_string(
+                    KeyValueType::SettingsAuthTokenSecret,
+                    Some(token_secret.clone()),
```
clemens-msupply, line 119: I see what you mean now I would do `.map_err(|err| UpdateSettingsError::RepositoryError(err))?;` again. If we can't write to the DB something is wrong...

service/src/settings_service.rs
```diff
@@ -113,11 +113,12 @@ pub trait SettingsServiceTrait: Sync + Send {
             Some(token_secret) => token_secret,
             None => {
                 let token_secret = uuid();
-                //Result is ignored here, as server can operate with the current secret (will change on next startup though)
-                let _result = key_value_store.set_string(
-                    KeyValueType::SettingsAuthTokenSecret,
-                    Some(token_secret.clone()),
-                );
+                let _result = key_value_store
```
clemens-msupply, line 116: Sorry for being pedantic: `let _result =` is not needed anymore when returning the error with `?`

In the method's docs comment we usually start with a Upper case

jmbrunskill: No worries, this is all good learning for me.

repository/src/schema/key_value_store.rs
andreievg: > I wonder if we should consider removing constraints from both, it is a KV store after all!

I think if keys are not known, we don't need to use enum, but if keys are known we should make an effort to document them in database schema (enums seem like a good way to do it, it keeps logic in code in sync with database schema doc).

server/src/lib.rs
```diff
@@ -47,7 +48,7 @@ async fn run_stage0(
 
     let cert_type = find_certs();
     let auth_data = Data::new(AuthData {
-        auth_token_secret: settings.auth.token_secret.to_owned(),
+        auth_token_secret: uuid(), //Temporary uuid secret for bootstrap mode
```
clemens-msupply, line 51: Note, we don't even use it because there is not auth in bootstrap mode

server/src/lib.rs
```diff
@@ -279,12 +281,14 @@ pub async fn start_server(
...
     let mut prefer_config_settings = true;
     let token_bucket = Arc::new(RwLock::new(TokenBucket::new()));
+    let token_secret = uuid();
     loop {
         match run_server(
             config_settings.clone(),
             prefer_config_settings,
             off_switch.clone(),
             token_bucket.clone(),
+            token_secret.clone(),
```
clemens-msupply, line 291: the token_secret doesn't need to be cloned because it is not used afterwards...

jmbrunskill: image
Seems to give me an error because of the loop?

server/src/lib.rs
clemens-msupply: Having another thought, there were some discussion of exposing the auth_token endpoint during bootstrap mode to test the connection. So its better to pass the correct secret to the bootstrap mode as well to avoid problems in the future in case we go this way.

server/src/lib.rs
jmbrunskill: image

clemens-msupply (approved):
Some more smaller change please:
- clean up migrations and remove variant from the key value store enum
- see comments

server/src/lib.rs
clemens-msupply: sorry wrong line I meant the clone in run_server

jmbrunskill: No worries, fixing that one

PR #1065, opened 2022-05-03
Remove permissive from CORS settings, and set Origin in Server Settings

### LINKED ISSUE #1045: Remove permissive CORS options from production
This also requires an `origin` config in the settings

jmbrunskill:
Closes #1045

### DIFF
--- a/android/src/android.rs
+++ b/android/src/android.rs
@@ -90,7 +90,7 @@ pub mod android {
                         develop: false,
                         debug_no_access_control: false,
                         debug_cors_permissive: false,
-                        cors_origin: "http://localhost:3003".to_string(),
+                        cors_origins: vec!["http://localhost:3003".to_string()],
                     },
                     database: DatabaseSettings {
                         username: "n/a".to_string(),
--- a/android/src/android.rs
+++ b/android/src/android.rs
@@ -89,6 +89,7 @@ pub mod android {
                         port,
                         develop: false,
                         debug_no_access_control: false,
+                        cors_origin: "http://localhost:3003".to_string(),
                     },
                     database: DatabaseSettings {
                         username: "n/a".to_string(),
--- a/android/src/android.rs
+++ b/android/src/android.rs
@@ -89,6 +89,7 @@ pub mod android {
                         port,
                         develop: false,
                         debug_no_access_control: false,
+                        debug_cors_permissive: false,
                         cors_origin: "http://localhost:3003".to_string(),
                     },
                     database: DatabaseSettings {
--- a/configuration/base.yaml
+++ b/configuration/base.yaml
@@ -2,7 +2,7 @@ server:
   port: 8000
   # debug_no_access_control: true # enable this if you want to ignore API authorisation (dummy user will be used for all operations)
   # debug_cors_permissive: true # enable this to allow any origin for development purposes
-  cors_origin: http://localhost:3003 # Used to set the allowed Origin in Cross Origin Requst Security
+  cors_origins: [http://localhost:3003, https://demo-open.msupply.org] # Used to set the allowed Origin in Cross Origin Requst Security
 sync:
   url: "http://localhost:2048"
   username: "username"
--- a/configuration/base.yaml
+++ b/configuration/base.yaml
@@ -1,7 +1,7 @@
 server:
   port: 8000
   # debug_no_access_control: true # enable this if you want to ignore API authorisation (dummy user will be used for all operations)
-  cors_origin: http://localhost:3003 #Used 
+  cors_origin: http://localhost:3003 #Used to set the allowed Origin in Cross Origin Requst Security
 sync:
   url: "http://localhost:2048"
   username: "username"
--- a/configuration/base.yaml
+++ b/configuration/base.yaml
@@ -1,7 +1,8 @@
 server:
   port: 8000
   # debug_no_access_control: true # enable this if you want to ignore API authorisation (dummy user will be used for all operations)
-  cors_origin: http://localhost:3003 #Used to set the allowed Origin in Cross Origin Requst Security
+  # debug_cors_permissive: true # enable this to allow any origin for development purposes
+  cors_origin: http://localhost:3003 # Used to set the allowed Origin in Cross Origin Requst Security
 sync:
   url: "http://localhost:2048"
   username: "username"
--- a/configuration/base.yaml
+++ b/configuration/base.yaml
@@ -1,6 +1,7 @@
 server:
   port: 8000
   # debug_no_access_control: true # enable this if you want to ignore API authorisation (dummy user will be used for all operations)
+  cors_origin: http://localhost:3003 #Used 
 sync:
   url: "http://localhost:2048"
   username: "username"
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -67,7 +67,7 @@ async fn run_stage0(
     let closure_settings = settings.clone();
 
     let mut http_server = HttpServer::new(move || {
-        let cors = choose_cors_policy(&closure_settings);
+        let cors = cors_policy(&closure_settings);
         App::new()
             .wrap(logger_middleware())
             .wrap(cors)
@@ -198,7 +198,7 @@ async fn run_server(
 
     let closure_settings = config_settings.clone();
     let mut http_server = HttpServer::new(move || {
-        let cors = choose_cors_policy(&closure_settings);
+        let cors = cors_policy(&closure_settings);
         App::new()
             .wrap(logger_middleware())
             .wrap(cors)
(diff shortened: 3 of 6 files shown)


server/src/lib.rs
```diff
@@ -196,7 +197,18 @@ async fn run_server(
     let mut http_server = HttpServer::new(move || {
         App::new()
             .wrap(logger_middleware())
-            .wrap(Cors::permissive())
+            .wrap(
+                Cors::default()
+                    .allowed_origin(&cors_origin)
+                    .supports_credentials()
+                    .allowed_methods(vec!["GET", "POST", "OPTIONS"])
```
jmbrunskill, line 204: @clemens-msupply any particular requirements here. This seems to work from testing the client for a bit locally...

configuration/base.yaml
```diff
@@ -1,6 +1,7 @@
 server:
   port: 8000
   # debug_no_access_control: true # enable this if you want to ignore API authorisation (dummy user will be used for all operations)
+  cors_origin: http://localhost:3003 #Used to set the allowed Origin in Cross Origin Requst Security
```
clemens-msupply, line 4: :nit: missing space "# Used to set ..." 

Maybe also move or duplicate the comment as a doc comment to the ServerSettings

server/src/lib.rs
clemens-msupply: Should be ok

clemens-msupply (commented):
Looks good thanks!

During development on local host we don't have an origin though. Could you please introduce a new `debug_cors_permissive` server settings option and use it like:

```rust
let cors = if config_settings.server.develop && config_settings.server.debug_cors_permissive {
    Cors::permissive()
} else {
    Cors::...
}
```
In this way you can enable permissive cors options during dev...

jmbrunskill:
> Looks good thanks!
> 
> During development on local host we don't have an origin though. Could you please introduce a new `debug_cors_permissive` server settings option and use it like:
> 
> ```rust
> let cors = if config_settings.server.develop && config_settings.server.debug_cors_permissive {
>     Cors::permissive()
> } else {
>     Cors::...
> }
> ```
> 
> In this way you can enable permissive cors options during dev...

I did consider doing something like this but was concerned that if we did development without the same CORS settings we could end up with unexpected production issues. However, I like it with your idea of adding an extra config flag that would make sure the develop was explicitly disabling this for a purpose. Will update.

jmbrunskill:
@clemens-msupply Do you see a scenario where we want to support more than 1 origin? A single origin seems wee bit restrictive, but I guess can always update if this requirement comes up...

clemens-msupply:
> @clemens-msupply Do you see a scenario where we want to support more than 1 origin? A single origin seems wee bit restrictive, but I guess can always update if this requirement comes up...

good point, easier to do it now?

jmbrunskill:
Testing with this origin config:

```
server:
  port: 8000
  # debug_no_access_control: true # enable this if you want to ignore API authorisation (dummy user will be used for all operations)
  # debug_cors_permissive: true # enable this to allow any origin for development purposes
  cors_origins: [http://localhost:3003, https://demo-open.msupply.org] # Used to set the allowed Origin in Cross Origin Request Security
```

```
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: http://example.com' 2>&1 | grep Origin 
> Origin: http://example.com
Origin is not allowed to make this request
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: http://localhost:3003' 2>&1 | grep Origin
> Origin: http://localhost:3003
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: http://demo-open.msupply.org' 2>&1 | grep Origin
> Origin: http://demo-open.msupply.org
Origin is not allowed to make this request
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: https://demo-open.msupply.org' 2>&1 | grep Origin
> Origin: https://demo-open.msupply.org
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
```

With debug_cors_permissive=true
```
server:
  port: 8000
  # debug_no_access_control: true # enable this if you want to ignore API authorisation (dummy user will be used for all operations)
  debug_cors_permissive: true # enable this to allow any origin for development purposes
  cors_origins: [http://localhost:3003, https://demo-open.msupply.org] # Used to set the allowed origins in Cross Origin Request Security
```

```
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: http://example.com' 2>&1 | grep Origin
> Origin: http://example.com
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: http://localhost:3003' 2>&1 | grep Origin
> Origin: http://localhost:3003
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: http://demo-open.msupply.org' 2>&1 | grep Origin
> Origin: http://demo-open.msupply.org
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
jamesbrunskill@Jamess-MBP ~ % curl -v 'http://localhost:8000/graphql' -X 'OPTIONS' -H 'Access-Control-Request-Headers: authorization,content-type' -H 'Access-Control-Request-Method: POST' -H 'Origin: https://demo-open.msupply.org' 2>&1 | grep Origin
> Origin: https://demo-open.msupply.org
< vary: Origin, Access-Control-Request-Method, Access-Control-Request-Headers
```

android/src/android.rs
```diff
@@ -89,6 +89,8 @@ pub mod android {
                         port,
                         develop: false,
                         debug_no_access_control: false,
+                        debug_cors_permissive: false,
```
clemens-msupply, line 92: Just realized this needs to be permissive on Android because the server runs on localhost... will write an issue for it.

server/src/lib.rs
```diff
@@ -193,10 +194,30 @@ async fn run_server(
...
+                    header::AUTHORIZATION,
+                    header::ACCEPT,
+                    header::CONTENT_TYPE,
+                ])
+                .max_age(3600);
+            for origin in cors_origins.iter() {
+                cors_tmp = cors_tmp.allowed_origin(origin);
+            }
+            cors_tmp
+        };
```
clemens-msupply, line 217: There is another cors permissive in run_stage0(), might be safer to use it there as well, e.g. have this as a helper function.

clemens-msupply (approved):
Looks good.

@mark-prins this requires another settings in dev mode to "disable" cors...

server/src/lib.rs
```diff
@@ -160,7 +162,10 @@ async fn run_server(
         debug_no_access_control: config_settings.server.develop
             && config_settings.server.debug_no_access_control,
     });
+
     let cors_origins = config_settings.server.cors_origins.clone();
+    let cors_permsissive_mode =
+        config_settings.server.develop && config_settings.server.debug_cors_permissive;
```
clemens-msupply, line 168: might as well have &config_settings as a parameter to choose_cors_policy and extract the information there...

server/src/lib.rs
```diff
@@ -336,3 +341,24 @@ fn load_certs(cert_files: SelfSignedCertFiles) -> Result<SslAcceptorBuilder, any
     builder.set_certificate_chain_file(cert_files.public_cert_file)?;
     Ok(builder)
 }
+
+fn choose_cors_policy(config_settings: &Settings) -> Cors {
```
clemens-msupply, line 345: :nit: just cors_policy?

server/src/lib.rs
```diff
@@ -64,10 +64,13 @@ async fn run_stage0(
 
     let restart_switch = Data::new(restart_switch);
 
+    let closure_settings = settings.clone();
```
clemens-msupply, line 67: would it work to create the cors object outside of the closure and then move it into the closure?

Issue #1049, opened 2022-04-27
Location mutation don't use std errors yet


clemens-msupply:
We have two types of errors:
1) std graphql errors as they are used in graphql
2) structured errors if we need to provide some structured information to the caller (graphql doesn't has schemas for the error details)

The location endpoint uses some structured errors that should be std errors:

```rust
// graphql/location/src/mutations/delete.rs
pub enum DeleteLocationErrorInterface {
    LocationNotFound(RecordNotFound), // std BadUserInput
    RecordBelongsToAnotherStore(RecordBelongsToAnotherStore), // std BadUserInput
    LocationInUse(LocationInUse), // keep as structured error
    DatabaseError(DatabaseError), // std InternalError
}

// graphql/location/src/mutations/insert.rs
pub enum InsertLocationErrorInterface {
    LocationAlreadyExists(RecordAlreadyExist), // std BadUserInput
    UniqueValueViolation(UniqueValueViolation), // std BadUserInput
    InternalError(InternalError), // std InternalError
    DatabaseError(DatabaseError), // std InternalError
}

// graphql/location/src/mutations/update.rs
pub enum UpdateLocationErrorInterface {
    LocationNotFound(RecordNotFound), // std BadUserInput
    UniqueValueViolation(UniqueValueViolation), // std BadUserInput
    RecordBelongsToAnotherStore(RecordBelongsToAnotherStore), // std BadUserInput
    InternalError(InternalError), // std InternalError
    DatabaseError(DatabaseError), // std InternalError
}
```

in `graphql/requisition/src/mutations/request_requisition/insert.rs` there is an example how we do the error handling now (`fn map_error`).

PR #1067, opened 2022-05-04
Add std errors to location mutation

roxy-dao:
#1049 

~~- Currently investigating test failure~~


service/src/location/mod.rs
```diff
@@ -1,4 +1,4 @@
-use self::{
+pub use self::{
```
clemens-msupply, line 1: Is this needed?

graphql/location/src/mutations/update.rs
```diff
@@ -42,7 +43,9 @@ pub fn update_location(
         Ok(location) => Ok(UpdateLocationResponse::Response(LocationNode::from_domain(
             location,
         ))),
-        Err(error) => Ok(UpdateLocationResponse::Error(error.into())),
```
clemens-msupply, line 45: You probably can remove our From implementation for the error... (the one that .into() uses)

clemens-msupply (approved):
Looks good thanks! just have a look if you can remove the error From implementations they should be unused now...

service/src/location/mod.rs
roxy-dao: Need it to import the Enums... I can move just the enums out to public

roxy-dao: Will change back~

clemens-msupply:
Some tests for the errors are failing now, should be straight forward to update (have a look for other tests which test std err)...

PR #1069, opened 2022-05-05
#1061 Fix user login when a user has no login permission for site's default store

### LINKED ISSUE #1061: Login allowed when central server details incorrect
Raised as https://github.com/openmsupply/openmsupply-client/issues/1197 
The steps to recreate are:
- use a new database for the server
- configure central server settings
- ...and sync, but do not log in using the `authToken` endpoint!
- modify the sync settings, so that the remote-server no longer can access the central server
- login using the authToken endpoint

Result: the user is authenticated - however, there is no user account in `[user_account]` and no connection to the central server. How is it possible to authenticate?

The result from the web client pov is that the user is authenticated, but no store available and permission denied errors pop up all over the place.

raviSussol:
Fixes #1061

### DIFF
--- a/repository/src/db_diesel/user_row.rs
+++ b/repository/src/db_diesel/user_row.rs
@@ -1,4 +1,8 @@
-use super::{user_row::user_account::dsl as user_account_dsl, StorageConnection, User};
+use super::{
+    user_row::user_account::dsl as user_account_dsl,
+    user_store_join_row::user_store_join::dsl as user_store_join_dsl,
+    StorageConnection, User
+};
 
 use crate::repository_error::RepositoryError;
 
@@ -11,6 +15,15 @@ table! {
         hashed_password -> Text,
         email -> Nullable<Text>,
     }
+}
+
+table! {
+  user_store_join (id) {
+      id -> Text,
+      user_id -> Text,
+      store_id -> Text,
+      is_default -> Bool,
+  }
 }
 
 #[derive(Clone, Queryable, Insertable, Debug, PartialEq, Eq, Default)]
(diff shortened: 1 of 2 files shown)


repository/src/db_diesel/user_row.rs
```diff
@@ -13,6 +17,15 @@ table! {
...
 
+table! {
+  user_store_join (id) {
+      id -> Text,
+      user_id -> Text,
+      store_id -> Text,
+      is_default -> Bool,
+  }
+}
+
```
clemens-msupply, line 28: This macro is already in `repository/src/db_diesel/user_store_join_row.rs` so you could just import it from there...

repository/src/db_diesel/user_row.rs
```diff
@@ -70,6 +83,29 @@ impl<'a> UserAccountRowRepository<'a> {
...
+            .first(&self.connection.connection);
+        match result {
+            Ok(row) => Ok(Some(row)),
+            Err(err) => match err {
+                diesel::result::Error::NotFound => Ok(None),
+                _ => Err(RepositoryError::from(err)),
+            },
+        }
+    }
+
```
clemens-msupply, line 108: As mentioned in the [ticket](https://github.com/openmsupply/remote-server/issues/1061) for the remote-server you can login even if you don't have permissions to any store. For example, you can still call endpoints which don't need a store such as the /me or /logout endpoint.

andreievg:
@raviSussol, sorry going to close this, as per: https://github.com/openmsupply/remote-server/issues/1061#issuecomment-1123004333

raviSussol:
> @raviSussol, sorry going to close this, as per: [#1061 (comment)](https://github.com/openmsupply/remote-server/issues/1061#issuecomment-1123004333)

Yea no worries... I had thought a user must have store permission check in order to login to the openmsupply (like in the mSupply desktop) but as per https://github.com/openmsupply/remote-server/issues/1061#issuecomment-1123178824 this change is no required now.

Issue #1054, opened 2022-04-27
Logout endpoint doesn't use std errors


clemens-msupply:
`graphql/general/src/queries/logout.rs`

They should be all std errors.

There are some structs in this files (including AccessDenied) please check if they are used and if not please remove.

PR #1072, opened 2022-05-06
#1054 Logout endpoint doesn't use std errors

roxy-dao:
#1054

### DIFF
--- a/graphql/graphql_schema/schema.graphql
+++ b/graphql/graphql_schema/schema.graphql
@@ -1,158 +1,160 @@
-type AccessDenied implements LogoutErrorInterface {
-	description: String!
-	fullError: String!
-}
 type AddFromMasterListError {
-	error: AddFromMasterListErrorInterface!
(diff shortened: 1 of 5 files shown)


graphql/graphql_schema/schema.graphql
```diff
@@ -1,158 +1,160 @@
-type AccessDenied implements LogoutErrorInterface {
-	description: String!
-	fullError: String!
-}
 type AddFromMasterListError {
-	error: AddFromMasterListErrorInterface!
+  error: AddFromMasterListErrorInterface!
```
clemens-msupply, line 2: could you please check why the spacing is different? e.g. if your editor did it or somebody else messed it up before. (trying to avoid that the whole file changes and makes it hard to see what the change was)

roxy-dao: Opps.. that might have been prettier

clemens-msupply (approved):
Thanks looks good

PR #1074, opened 2022-05-06
Add CORS to readme, include http://localhost:8000 in defaults

### LINKED ISSUE #1073: Improve documentation for CORS Settings
More restrictive CORS settings can cause issues with setting up a development environment, documenting these in the README will assist with this.

jmbrunskill:
Closes #1073

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ server:
 
 ## CORS settings
 
-By default remote-server limits Cross-Origin Resource Sharing to the origins configured in the server section of the configuration yaml
+By default remote-server limits Cross-Origin Resource Sharing to the origins configured in the server section of the configuration yaml.
 This is a security mechanism to reduce the risk of a malicious site accessing an authenticated connection with mSupply.
 Rust enforces the allowed origins in requests, even if the browser doesn't, by returning a 400 error when the Origin isn't specified in a request, or if doesn't match one of origins configured in cors_origins.
 
@@ -165,6 +165,6 @@ For development purposes you can set the `debug_cors_permissive` parameter to 't
 ```
 server:
   port: 8000
-  ddebug_cors_permissive: true
+  debug_cors_permissive: true
   cors_origins: [http://localhost:3003, https://youwebserver:yourport]
 ```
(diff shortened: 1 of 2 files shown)


andreievg (approved):
Thanks

README.md
```diff
@@ -146,3 +146,25 @@ To enable development mode the `develop` flag must be set to `true` in the setti
...
+server:
+  port: 8000
+  cors_origins: [http://localhost:3003, https://youwebserver:yourport]
+````
+
+For development purposes you can set the `debug_cors_permissive` parameter to 'true' to allow any origin.
+```
+server:
+  port: 8000
+  ddebug_cors_permissive: true
```
clemens-msupply, line 168: ddebug

README.md
```diff
@@ -146,3 +146,25 @@ To enable development mode the `develop` flag must be set to `true` in the setti
 server:
   develop: true
 ```
+
+## CORS settings
+
+By default remote-server limits Cross-Origin Resource Sharing to the origins configured in the server section of the configuration yaml
+This is a security mechanism to reduce the risk of a malicious site accessing an authenticated connection with mSupply.
+Rust enforces the allowed origins in requests, even if the browser doesn't, by returning a 400 error when the Origin isn't specified in a request, or if doesn't match one of origins configured in cors_origins.
```
clemens-msupply, line 154: This is weird what if you want use your api outside of the browser as a general purpose api?

Missing "." after "configuration yaml"

clemens-msupply (approved):

README.md
andreievg: @jmbrunskill, what happens if origin is not specified ?

jmbrunskill: @andreievg 400 error if origin is not specified.

andreievg: thanks

jmbrunskill: > This is weird what if you want use your api outside of the browser as a general purpose api?
I assume you need to use custom code to handle this scenario? I've just raised ticket #1075 to handle same origin requests using the allowed_origin_fn I think this should allow us to implement the behaviour you're expecting.

PR #1078, opened 2022-05-09
Android update

clemens-msupply:
- Update config for Android (add danger_allow_http flag for now)
- Fix for sqlite problem on Android
- Some refactoring
- Improved debug output for DB transactions


repository/src/db_diesel/storage_connection.rs
```diff
@@ -205,6 +210,14 @@ impl StorageConnectionManager {
             transaction_level: Cell::new(0),
         })
     }
+
+    // Note, this method is only needed for an Android workaround to avoid adding a diesel
+    // dependency to the server crate.
+    pub fn execute(&self, sql: &str) -> Result<(), RepositoryError> {
```
andreievg, line 216: was thinking about this for a little, and though maybe it's a good idea to relocate DB specific methods (like `get_storage_connection_manager` and run_db_migrations) to StorageConnectionManager and could even have initsql here, just food for though

android/src/android.rs
```diff
@@ -75,39 +74,38 @@ pub mod android {
         _: JClass,
         port: jchar,
         db_path: JString,
+        cache_dir: JString,
     ) -> jlong {
         android_logger::init_once(Config::default().with_min_level(Level::Trace));
 
         let (off_switch, off_switch_receiver) = oneshot::channel();
         let db_path: String = env.get_string(db_path).unwrap().into();
+        let cache_dir: String = env.get_string(cache_dir).unwrap().into();
```
andreievg, line 83: can cache_dir be deduced from relative path or does it have to be passed on from Java ?

andreievg (approved):

repository/src/db_diesel/storage_connection.rs
clemens-msupply: yes or in another StorageUtils helper

Btw I hope I can remove this execute method at some point when the underlying build issue is fixed...

android/src/android.rs
clemens-msupply: not sure if it is a standard path or depends on the device. It is definitely app id dependent, i.e. every app has its own dir. In Android there are getCacheDir and getFilesDir method for it so it safer to use these.

Issue #980, opened 2022-03-27
Report: default query definition


clemens-msupply:
- Decide what goes into the default report queries.
- Are default queries the same for all customers? (probably yes because otherwise the docs might get confusing)
- Are default queries editable? (see above)
- Where are default queries stored? In a DB?
- How to extend a default query? just add new fields or introduce a new version, e.g. STOCKTAKE_QUERY_V2

PR #1081, opened 2022-05-10
1079 remove automatic filter of name belonging to current store

### LINKED ISSUE #1079: Remove automatic filter of name belonging to current store
This was added to filter out current store name from names list (to prevent user selecting the same name as current store). It's been pointed out isVisible filter would already filter out current store's name since name store join shouldn't exist between a store and it's associated name

andreievg:
closes #1079 

Also added store query, original requirement for this was for reports to be able to get current store name, thus additions of store query.

(I think we can quite easily use `StoreNode` in `me` query now, I didn't want to go too far out of scope of the issue)

### DIFF
--- a/graphql/general/src/lib.rs
+++ b/graphql/general/src/lib.rs
@@ -61,6 +61,15 @@ impl GeneralQueries {
         get_names(ctx, store_id, page, filter, sort)
     }
 
+    pub async fn store(
+        &self,
+        ctx: &Context<'_>,
+        store_id: String,
+        id: String,
+    ) -> Result<StoreResponse> {
+        get_store(ctx, &store_id, &id)
+    }
+
     pub async fn stores(
         &self,
         ctx: &Context<'_>,
--- a/graphql/general/src/lib.rs
+++ b/graphql/general/src/lib.rs
@@ -61,13 +61,8 @@ impl GeneralQueries {
         get_names(ctx, store_id, page, filter, sort)
     }
 
-    pub async fn store(
-        &self,
-        ctx: &Context<'_>,
-        store_id: String,
-        id: String,
-    ) -> Result<StoreResponse> {
-        get_store(ctx, &store_id, &id)
+    pub async fn store(&self, ctx: &Context<'_>, id: String) -> Result<StoreResponse> {
+        get_store(ctx, &id)
     }
 
     pub async fn stores(
--- a/graphql/general/src/queries/store.rs
+++ b/graphql/general/src/queries/store.rs
@@ -1,5 +1,6 @@
 use async_graphql::*;
 use graphql_core::generic_filters::{EqualFilterInput, EqualFilterStringInput};
+use graphql_core::simple_generic_errors::{NodeError, NodeErrorInterface};
 use graphql_core::standard_graphql_error::validate_auth;
 use graphql_core::{
     generic_filters::SimpleStringFilterInput, pagination::PaginationInput,
@@ -43,8 +44,39 @@ pub struct StoreConnector {
 }
 
 #[derive(Union)]
+pub enum StoreResponse {
+    Error(NodeError),
+    Response(StoreNode),
+}
+
+#[derive(Union)]
 pub enum StoresResponse {
     Response(StoreConnector),
+}
+
+pub fn get_store(ctx: &Context<'_>, store_id: &str, id: &str) -> Result<StoreResponse> {
+    validate_auth(
+        ctx,
+        &ResourceAccessRequest {
+            resource: Resource::QueryInvoice,
+            store_id: Some(store_id.to_string()),
+        },
+    )?;
+
+    let service_provider = ctx.service_provider();
+    let service_context = service_provider.context()?;
+    let service = &service_provider.general_service;
+
+    let store_option = service.get_store(&service_context, id)?;
+
+    let response = match store_option {
+        Some(store) => StoreResponse::Response(StoreNode::from_domain(store)),
+        None => StoreResponse::Error(NodeError {
+            error: NodeErrorInterface::record_not_found(),
+        }),
+    };
+
+    Ok(response)
 }
 
 pub fn stores(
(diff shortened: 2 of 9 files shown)


graphql/general/src/queries/store.rs
```diff
@@ -42,11 +43,42 @@ pub struct StoreConnector {
...
 #[derive(Union)]
 pub enum StoresResponse {
     Response(StoreConnector),
 }
 
+pub fn get_store(ctx: &Context<'_>, store_id: &str, id: &str) -> Result<StoreResponse> {
+    validate_auth(
+        ctx,
+        &ResourceAccessRequest {
+            resource: Resource::QueryInvoice,
```
clemens-msupply, line 61: Wrong resource...

graphql/general/src/lib.rs
```diff
@@ -61,6 +61,15 @@ impl GeneralQueries {
         get_names(ctx, store_id, page, filter, sort)
     }
 
+    pub async fn store(
+        &self,
+        ctx: &Context<'_>,
+        store_id: String,
+        id: String,
+    ) -> Result<StoreResponse> {
```
clemens-msupply, line 69: in which use-case would `store_id` and `id` be different? and if they are different can't we just check if user has permissions to lookup other stores (even not having access to the particular store)? If we could avoid it I think just having one store_id is more straight forward?

graphql/general/src/queries/store.rs
```diff
@@ -42,11 +43,42 @@ pub struct StoreConnector {
...
+            store_id: Some(store_id.to_string()),
+        },
+    )?;
+
+    let service_provider = ctx.service_provider();
+    let service_context = service_provider.context()?;
+    let service = &service_provider.general_service;
+
+    let store_option = service.get_store(&service_context, id)?;
```
clemens-msupply, line 70: think just `store` would work as well here (up to you)

graphql/general/src/queries/store.rs
```diff
@@ -42,11 +43,42 @@ pub struct StoreConnector {
     nodes: Vec<StoreNode>,
 }
 
+#[derive(Union)]
+pub enum StoreResponse {
+    Error(NodeError),
```
clemens-msupply, line 48: Out of scope of this PR but we could probably get rid of this NodeError and return and optional like response?

service/src/store.rs
```diff
@@ -19,3 +19,10 @@ pub fn get_stores(
         count: i64_to_u32(repository.count(filter)?),
     })
 }
+
+pub fn get_store(ctx: &ServiceContext, id: &str) -> Result<Option<Store>, RepositoryError> {
```
clemens-msupply, line 23: Could take a StoreFilter as a parameter (instead of the id) to make it more powerful...

clemens-msupply (approved):
Do we need store_id and id in the query? (see comment)

graphql/general/src/queries/store.rs
andreievg: thanks: a0fd529cd5dad3ce6a8a456b1b5402839605cf5d

graphql/general/src/queries/store.rs
andreievg: Hmm we have the following for single responses
```
#[derive(Union)]
pub enum InvoiceResponse {
    Error(NodeError),
    Response(InvoiceNode),
}

#[derive(Union)]
pub enum RequisitionResponse {
    Error(RecordNotFound),
    Response(RequisitionNode),
}

#[derive(Union)]
pub enum LocationResponse {
    Error(NodeError),
    Response(LocationNode),
}
```

Was trying to be consistent on that front. (although can see that RequisitionResponse varies). We should ask front end what the preferred response should be. Also one benefit of using NodeError, would be to return something like RecordBelongsToAnotherStore (perhaps not for `store` query but maybe for others.

graphql/general/src/queries/store.rs
andreievg: i like store_option for some reason, although I don't think it make much difference in the long run

service/src/store.rs
andreievg: good idea, ee801ddc097787a66ee494cd544a2463190dcaef, i think we discussed this sometime.

i.e. in invoice and requisition we have a special service to get_invoice/requisition and then get_invoice/requisition_by_number (could just be one), and similar in stocktake, we use get_stocktakes (for get_stocktake_by_number) service with a bit extra parameters and extra conversions (just get_stocktake service that accepts), I think this pattern might have been better for others

graphql/general/src/lib.rs
andreievg: Not a current requirement use case, was just trying to align with existing use case.

>  and if they are different can't we just check if user has permissions to lookup other stores (even not having access to the particular store)?

I am not sure how store permissions differentiate between global permissions, in current mSupply ALL permissions are linked to store, I had a look and it seems that some permissions have been made as 'global' (where store_id can be None). I've remove store_id for now, but would like to get some clarity about this. 1c566f1d0abc380a414cf71cb7f032b8cee3992e

andreievg:
@clemens-msupply i've also added to default reports 196f7a46969e663a3d67a3cc3e1f8ff3e14a33b5, do we have a test that check the bit that connects queried data to report generation btw ? Wanted to add this to the test but couldn't find (did a manual test on the queries though)

graphql/general/src/queries/store.rs
clemens-msupply: for these "get_node" method I can't really think of errors that are not std errors but this means they aren't any... And if we go with something like `get_node(id) -> Result<Option<Node>>` we can't introduce any new none std errors. The get_node would be very convenient to use though.

graphql/general/src/lib.rs
clemens-msupply: ok I see thanks

clemens-msupply:
> @clemens-msupply i've also added to default reports [196f7a4](https://github.com/openmsupply/remote-server/commit/196f7a46969e663a3d67a3cc3e1f8ff3e14a33b5), do we have a test that check the bit that connects queried data to report generation btw ? Wanted to add this to the test but couldn't find (did a manual test on the queries though)

Thanks, no no tests right now. This was kind of blocked by some related open questions: https://github.com/openmsupply/remote-server/issues/980. But yes I we can start writing basic tests for default queries regardless.

graphql/general/src/queries/store.rs
andreievg: I think original `SingleRecordError` was there for that reason

PR #1083, opened 2022-05-10
Docker build tools

andreievg:
Changes as per commit descriptions. Was easier to move changes then updating https://github.com/openmsupply/remote-server/pull/1020 from develop. There is one difference, refresh date was relocated to `demo` crate


demo/src/refresh_dates.rs
```diff
@@ -0,0 +1,592 @@
...
+
+            sql_query(&query).execute(&self.connection.connection)?;
+        }
+
+        Ok(())
+    }
+}
+
+// Couldn't find less hacky way to serialise manually
+table! {
```
clemens-msupply, line 247: had a quick look at the [ToSql](https://docs.diesel.rs/diesel/serialize/trait.ToSql.html#) crate but gave up pretty quickly :)

server/src/cli.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[derive(clap::Subcommand)]
+enum Action {
+    /// Export graphql schema
+    ExportGraphqlSchema,
+    /// Initialise empty database (existing database will be dropped, and new one created and migrated)
+    InitialiseDatabase,
+    /// Initilise from running mSupply server
+    InitialiseFromCentral {
```
clemens-msupply, line 31: Think some more details what exactly this InitialiseFromCentral command is doing would be helpful for the user

service/src/login.rs
```diff
@@ -71,9 +72,12 @@ impl LoginService {
...
             Ok(result) => Ok(result),
             Err(err) => {
-                let delay = Duration::from_secs(min_err_response_time_sec)
-                    - now.elapsed().unwrap_or(Duration::from_secs(0));
-                tokio::time::sleep(delay).await;
+                let elapsed = now.elapsed().unwrap_or(Duration::from_secs(0));
+                let minimum = Duration::from_secs(min_err_response_time_sec);
+                if elapsed < minimum {
+                    tokio::time::sleep(minimum - elapsed).await;
+                }
```
clemens-msupply, line 79: Was it just a refactor or was it a problem? I remember fixing a test issue where subtracting a Duration panicked because the result becomes negative but couldn't find it right now...

clemens-msupply (approved):
Looks good thanks you!

service/src/login.rs
andreievg: Hmm, I can't replicate this in API, even if i put a breakpoint in 4d and delay a bit longer. But in cli, when `min_err_response_time_sec = 0` it throws an overflow error

demo/src/refresh_dates.rs
andreievg: yeah i tried it too but to no resolve (it serialises to bytes, meant to be sent as part of parameters, i.e. $1, $2, so doesn't work as string substitution), but `sql_query` doesn't support parameters

server/src/cli.rs
andreievg: 83f11cc0eb66feb67643a05bd1a8e27a77ffd572, will need to change description once we specify sync credentials and url via cli

PR #1086, opened 2022-05-12
1070 android tls

### LINKED ISSUE #1070: Android requires permissive cors settings -> security issue?
remote-server on Android runs on localhost and thus requires permissive cors settings, check if this is a security problem.

clemens-msupply:
- Add option to configure the certs directory
- Use rustls instead of openssl to avoid openssl issues on Android

Closes #1070

### DIFF
--- a/android/src/android.rs
+++ b/android/src/android.rs
@@ -3,14 +3,17 @@
 pub mod android {
     extern crate jni;
 
+    use std::io::Write;
+    use std::path::{Path, PathBuf};
     use std::thread::{self, JoinHandle};
     use std::{collections::HashMap, sync::Mutex};
 
     use jni::sys::{jchar, jshort};
+    use rcgen::generate_simple_self_signed;
     use repository::database_settings::DatabaseSettings;
 
     use server::settings::{ServerSettings, Settings};
-    use server::start_server;
+    use server::{start_server, PRIVATE_CERT_FILE, PUBLIC_CERT_FILE};
     use tokio::sync::oneshot;
 
     use self::jni::objects::{JClass, JString};
@@ -56,6 +59,20 @@ pub mod android {
         })
     });
 
+    fn generate_certs(cert_dir: &PathBuf) {
+        let subject_alt_names = vec!["localhost".to_string()];
+        let cert = generate_simple_self_signed(subject_alt_names).unwrap();
+
+        std::fs::create_dir_all(cert_dir).unwrap();
+        let mut file = std::fs::File::create(cert_dir.join(PRIVATE_CERT_FILE)).unwrap();
+        file.write_all(cert.serialize_private_key_pem().as_bytes())
+            .unwrap();
+
+        let mut file = std::fs::File::create(cert_dir.join(PUBLIC_CERT_FILE)).unwrap();
+        file.write_all(cert.serialize_pem().unwrap().as_bytes())
+            .unwrap();
+    }
+
     /// Bindings test method. TODO: remove when not needed anymore
     #[no_mangle]
     pub unsafe extern "C" fn Java_org_openmsupply_client_RemoteServer_rustHelloWorld(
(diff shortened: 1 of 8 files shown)


server/src/lib.rs
```diff
@@ -341,32 +355,67 @@ pub async fn start_server(
...
 }
 
-const PRIVATE_CERT_FILE: &str = "./certs/key.pem";
-const PUBLIC_CERT_FILE: &str = "./certs/cert.pem";
-fn find_certs() -> ServerCertType {
-    if !Path::new(PRIVATE_CERT_FILE).exists() || !Path::new(PUBLIC_CERT_FILE).exists() {
+pub const PRIVATE_CERT_FILE: &str = "key.pem";
+pub const PUBLIC_CERT_FILE: &str = "cert.pem";
+
+fn find_certs(cert_dir: &str) -> ServerCertType {
```
andreievg, line 374: Would prefer a seperate file for cert specific methods.

clemens-msupply: [d468c5f](https://github.com/openmsupply/remote-server/pull/1086/commits/d468c5ffca6e8727516611e5680dc251f69267ca)

I moved all self signed cert related code. Think its much cleaner now, thanks!

android/src/android.rs
```diff
@@ -56,6 +60,20 @@ pub mod android {
         })
     });
 
+    fn generate_certs(cert_dir: &PathBuf) {
+        let subject_alt_names = vec!["localhost".to_string()];
+        let cert = generate_simple_self_signed(subject_alt_names).unwrap();
```
clemens-msupply, line 65: btw. this is the easiest way to generate a self signed cert I ever seen :-) By default it uses elliptic curve key pairs which are also very fast to generate.

jmbrunskill: Yes, that's very nice!

server/Cargo.toml
```diff
@@ -22,16 +22,18 @@ service = { path = "../service" }
 util = { path = "../util" }
 
 actix-cors = "0.6.1"
-actix-web = { version= "4.0.1", features = ["openssl"] } 
+actix-web = { version= "4.0.1", features = ["rustls"] } 
 actix-files = "0.6.0"
 openssl = { version = "0.10", features = ["v110"] }
```
andreievg, line 27: can we remove this now ?

andreievg (approved):
Looks good, thanks, just removal of openssl from deps would be great

server/src/self_signed_certs.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Load rustls server config
+pub fn load_self_signed_certs_rustls(
+    cert_files: SelfSignedCertFiles,
+) -> Result<ServerConfig, anyhow::Error> {
+    let certfile = std::fs::File::open(&cert_files.public_cert_file)?;
+    let mut reader = BufReader::new(certfile);
+    let certs = rustls_pemfile::certs(&mut reader)
+        .unwrap()
+        .iter()
```
andreievg, line 43: this is probably stylistic thing, but I personally like into_iter/cloneless, becomes:

 .into_iter()
.map(rustls::Certificate)
.collect();

clemens-msupply: thanks, think I actually copied it from the rustls examples :)

[7c5e2a7](https://github.com/openmsupply/remote-server/pull/1086/commits/7c5e2a7443a39e836e5758f31da92e533a8ebee9)

clemens-msupply:
> Looks good, thanks, just removal of openssl from deps would be great

Sorry forgot to mention. I tried this but unfortunately it didn't make the Android build happy. Some dependency still uses openssl and if I don't bundle it explicitly it can't find the shared libs when doing the cross compilation... Bit annoying.

andreievg:
> Sorry forgot to mention. I tried this but unfortunately it didn't make the Android build happy. Some dependency still uses openssl and if I don't bundle it explicitly it can't find the shared libs when doing the cross compilation... Bit annoying.

Is this only when compiling for android targets ? (i tried removing openssl and associated reference in android features), and managed to build from android folder (for default target)

clemens-msupply:
> > Sorry forgot to mention. I tried this but unfortunately it didn't make the Android build happy. Some dependency still uses openssl and if I don't bundle it explicitly it can't find the shared libs when doing the cross compilation... Bit annoying.
> 
> Is this only when compiling for android targets ? (i tried removing openssl and associated reference in android features), and managed to build from android folder (for default target)

It only if you compile for a different target architecture...

Issue #867, opened 2022-03-02
Make sure all basic fields are exposed in query filters and sorts for all types


andreievg:

andreievg:
@roxy-dao this will be good to do also, thank you

roxy-dao:
- Unexposed fields
  - Invoices (Filter by Pricing Rows too?)
    - Filter
      - [ ] transportReference
      - [ ] colour
      - [ ] onHold
      - [ ] nameId
      - [ ] userId
    - Sort
      - [ ] theirReference
      - [ ] transportReference
      - [ ] colour
  - Items
    - Sort
      - [ ] type
   - Locations
    - Filter
      - [ ] onHold
  - Requisitions
    - Filter
      - [ ] nameId
      - [ ] userId
      - [ ] linkedRequisitionId
      - [ ] expectedDeliveryDate
      - [ ] maxMonthsOfStock
      - [ ] minMonthOfStock
      - [ ] name (NameRow)
      - [ ] code (NameRow)
      - [ ] isCustomer (NameRow)
      - [ ] isSupplier (NameRow)
    - Sort
      - [ ] expectedDeliveryDate
      - [ ] colour
      - [ ] theirReference
      - [ ] maxMonthsOfStock
      - [ ] minMonthsOfStock
      - [ ] code (name_row)
  - Stocktakes (can add to #981 )
    - Filter
      - [ ] userId
      - [ ] isLocked
      - [ ] inventoryAdjustmentId

andreievg:
@roxy-dao, I've updated above, thanks for compiling the list

roxy-dao:
> @roxy-dao, I've updated above, thanks for compiling the list

Thanks 😄

PR #1090, opened 2022-05-12
#867 make sure all basic fields are exposed in query filters and sorts for all types

roxy-dao:
#867 

name_id and name for requisition queries seem to be under otherPartyName & otherPartyID

### DIFF
--- a/graphql/general/src/queries/item.rs
+++ b/graphql/general/src/queries/item.rs
@@ -20,6 +20,7 @@ use service::{
 pub enum ItemSortFieldInput {
     Name,
     Code,
+    Type,
 }
 
 #[derive(InputObject)]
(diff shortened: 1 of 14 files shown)


andreievg (approved):
Looks good thanks, would be good to do one more thing (but not essential), in `create_filtered_query` sometimes we don't destructure `filter`, which means it's possible to add filter without actually enforcing it, would be great if all `create_filtered_query` destructured filter

roxy-dao:
> `[ffec29a](/openmsupply/remote-server/pull/1090/commits/ffec29a6e11ea13ae03f6f456626a357179c1248) `

[ffec29a](https://github.com/openmsupply/remote-server/pull/1090/commits/ffec29a6e11ea13ae03f6f456626a357179c1248)

clemens-msupply (approved):
LGTM thanks!

PR #1093, opened 2022-05-15
Add comment why we still have openssl as a dependency

clemens-msupply:
@andreievg though I better put a comment into the Cargo.toml file regarding your question why we still need openssl.


andreievg (approved):
Thanks

PR #1098, opened 2022-05-17
Update permissions validation to use explict 'NoPermissionRequired'

jmbrunskill:
Note: This might need a bit of testing, any routes that were using a resource without permissions mapping may start failing.
I've included the ones that are in the comments but there might be others.

### DIFF
--- a/service/src/authorisation.rs
+++ b/service/src/authorisation.rs
@@ -363,7 +363,7 @@ fn validate_resource_permissions(
     })
 }
 
-pub trait ValidationServiceTrait: Send + Sync {
+pub trait AuthorisationServiceTrait: Send + Sync {
     fn validate(
         &self,
         ctx: &ServiceContext,
@@ -373,19 +373,19 @@ pub trait ValidationServiceTrait: Send + Sync {
     ) -> Result<ValidatedUser, ValidationError>;
 }
 
-pub struct ValidationService {
+pub struct AuthorisationService {
     pub resource_permissions: HashMap<Resource, PermissionDSL>,
 }
 
-impl ValidationService {
+impl AuthorisationService {
     pub fn new() -> Self {
-        ValidationService {
+        AuthorisationService {
             resource_permissions: all_permissions(),
         }
     }
 }
 
-impl ValidationServiceTrait for ValidationService {
+impl AuthorisationServiceTrait for AuthorisationService {
     fn validate(
         &self,
         context: &ServiceContext,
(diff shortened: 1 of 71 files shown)


service/src/permission_validation.rs
```diff
@@ -425,6 +429,14 @@ impl ValidationServiceTrait for ValidationService {
                     ));
                 }
             };
+        } else {
+            //The requested resource doesn't have a permission mapping assigned (server error)
+            return Err(ValidationError::Denied(
+                ValidationDeniedKind::InsufficientPermission {
```
jmbrunskill, line 435: This should possilby be a new ValidationDeniedError kind?

jmbrunskill:
@clemens-msupply I have discussed this briefly with you before, but should I add on re-factoring ValidationService into AuthService as part of the this PR? I think this helps explain it's purpose, validation could be use for other non-authorisation related purposes. Have already done this for Health Supply Hub but didn't for remote-server yet.

service/src/permission_validation.rs
clemens-msupply: Or just InternalError because it shouldn't happen (and basic testing would catch it anyway)

clemens-msupply: You could return early again here by doing:
```rust
let required_permissions = match required_permissions {
  Some(required_permissions) => required_permissions,
  None => return Err(ValidationError::InternalError(...)
};
```

andreievg:
> This might need a bit of testing

Hmm, would these changes not be covered by existing tests ?

> Yes think this is a good idea! @andreievg what do you think?

Sounds good. `Validation` as a term is used a lot in other services, to validate the inputs etc.. So yeah `Authorisation` sounds like a better name for user validation

jmbrunskill:
> Hmm, would these changes not be covered by existing tests ?

I'm not sure how comprehensive the testing would be in this area.

If code is relying on the default 'allow' behaviour without a comment in the Resource to PermissionDSL Map, I predict there is probably is no test case for it, but happy to be proven wrong. 

Tests are passing.

service/src/authorisation.rs
```diff
@@ -363,7 +363,7 @@ fn validate_resource_permissions(
     })
 }
 
-pub trait ValidationServiceTrait: Send + Sync {
+pub trait AuthorisationServiceTrait: Send + Sync {
```
clemens-msupply, line 366: I think just AuthService is better because it does authentication and authorisation...

jmbrunskill: I'm happy either way @andreievg ?

andreievg: Oki

jmbrunskill: Done :)

clemens-msupply (approved):
thanks!

Issue #1085, opened 2022-05-11
Can login with wrong credential


bijaySussol:
Been relayed with:
https://github.com/openmsupply/openmsupply-client/issues/1197
https://github.com/openmsupply/remote-server/issues/1061

Access thingy, so raising a new one

Think this time is nailed.
What is happening is: Even if the user tries to login to a wrong password i.e with previous password, It allows to login.

### Steps to reproduce:

- Setup Central, remote and the Front end
- After the setup is completed, It should work normally
- Once it is on workable state
- From Central server, change that User password, but make sure you know your last password
- Once the Password is changed
- From the Frontend, Type that user last password
- This user will be authenticated fine from Remote server and takes to the Front end Dashboard page
- Doing restart of the remote server still does not make any difference

But....

- Once from the frontend, If logged in with the new password which is just changed
- And tries to re-login with the Previous Password, It does not validate which is correct But redoing `Steps to reproduce` will again be a problem

clemens-msupply:
I suspect that in `service/src/login.rs` in `do_login` the central server login failure doesn't get interpreted as a LoginFailure error...

andreievg:
what should happen to user credentials once user tries to login to central but has entered invalid password (but it's valid password in remote data) ?

andreievg:
do we have a flag 'outdated password' or something like that ?

clemens-msupply:
Good question. So we would have to remove the local user/permissions if:
1) Central server available
2) AND user login failed with "wrong password"
2) AND entered password is the correct current remote server password

andreievg:
> Good question. So we would have to remove the local user/permissions if:
> 1) Central server available
> 2) AND user login failed with "wrong password"
> 2) AND entered password is the correct current remote server password

I think the condition is very good, `but` we shouldn't remove user or permissions (removing user would mean a placeholder user will be now shown in UI). Can we just add a flag on a user ? Full name for the flag would be `last_login_to_central_with_local_password_failed`, but of-course it can be `allow_offline_login` = false/true (with comment in db why it's there)

clemens-msupply:
you mean if a user is still logged into another client and does a mutation and then the placeholder is used because the user has been deleted during a parallel login? good point!

I can think of two options:
1) Invalidate existing sessions for this user (by removing all tokens from the TokenBucket) so that the user is forced to log in using the new correct password.
2) Or as you said flag the user so that the user need to login with the new password.

[...]

PR #1099, opened 2022-05-17
Can login with wrong credential #1085

roxy-dao:
#1085 

User shouldn't be able to log in with old password in api now if password has been changed in central.

Note: Still can use old password if not logged into the central server and new password has not been saved in remote

Test: 
- [ ] Change password in central
- [ ] Try to login with old pass in api while still connected to central
- [ ] Disconnect from central and try to log in with old password
- [ ] Connect back to central and save new password by logging in

### DIFF
--- a/service/src/apis/login_v4.rs
+++ b/service/src/apis/login_v4.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug)]
 pub enum LoginV4Error {
-    Unauthorized,
+    Unauthorised,
     ConnectionError(reqwest::Error),
 }
 
@@ -139,10 +139,12 @@ impl LoginApiV4 {
             .map_err(|e| LoginV4Error::ConnectionError(e))?;
 
         if reqwest::StatusCode::UNAUTHORIZED == response.status() {
-            return Err(LoginV4Error::Unauthorized);
+            return Err(LoginV4Error::Unauthorised);
         }
 
         let response = response
+            .error_for_status()
+            .map_err(|e| LoginV4Error::ConnectionError(e))?
             .json()
             .await
             .map_err(|e| LoginV4Error::ConnectionError(e))?;
--- a/service/src/apis/login_v4.rs
+++ b/service/src/apis/login_v4.rs
@@ -1,7 +1,11 @@
 use reqwest::{Client, Url};
 use serde::{Deserialize, Serialize};
 
-pub type LoginV4Error = anyhow::Error;
+#[derive(Debug)]
+pub enum LoginV4Error {
+    Unauthorized,
+    ConnectionError(anyhow::Error),
+}
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub enum LoginUserTypeV4 {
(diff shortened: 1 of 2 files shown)


service/src/apis/login_v4.rs
```diff
@@ -128,13 +132,21 @@ impl LoginApiV4 {
...
-            .await?
-            .error_for_status()?;
+            .await
+            .map_err(|e| LoginV4Error::ConnectionError(e.into()))?;
 
-        let response = response.json().await?;
-        Ok(response)
+        match response.status() {
+            reqwest::StatusCode::UNAUTHORIZED => Err(LoginV4Error::Unauthorized),
+            _ => {
```
andreievg, line 143: hmm, we did error_for_status previously, maybe it's a good idea to keep that check ?

Also don't really need to use anyhow error now can use reqwest::Error ?

Lastly kind of like reducing nesting (this is probably stylistic preference, see `if let` vs match in the example below):

https://github.com/openmsupply/remote-server/compare/%231085-Can-login-with-wrong-credential-...1085-pr-comment-1

You'll note that from_reqwest_error was created, although it looks like a perfect time to use from trait, but we've been deliberately avoiding it as it adds indirection that's hard to trace (we do use from impl but in cases where it is repeatedly used).

service/src/login.rs
```diff
@@ -141,19 +141,28 @@ impl LoginService {
...
-                FetchUserError::ConnectionError(format!(
-                    "Failed to reach the central server to fetch data for {}: {:?}",
-                    username, err
-                ))
-            })?;
+        {
+            Ok(user_data) => user_data,
+            Err(err) => match err {
+                LoginV4Error::Unauthorized => {
+                    return Err(FetchUserError::Unauthenticated);
```
andreievg, line 155: If Unauthorised, but can login with DB credentials, we would need to make sure to 'disable' user being able to login, i've left another comment on the issue.

> Note: Still can use old password if not logged into the central server and new password has not been saved in remote

I think we need to try and disable access to user in offline mode with their old password if they tried to login with their old password while there is a connection to central

> Test: 
> - [ ] Change password in central
> - [ ] Try to login with old pass in api while still connected to central
> - [ ] Disconnect from central and try to log in with old password
> - [ ] Connect back to central and save new password by logging in

We should try to automate testing as much as possible, see below (last test fails btw as per above)

https://github.com/openmsupply/remote-server/compare/1085-pr-comment-1...1085-pr-comment-2

service/src/apis/login_v4.rs
roxy-dao: error_for_status causes the old password login while on central 🤔

service/src/login.rs
```diff
@@ -334,53 +346,169 @@ mod test {
...
+            )
+            .await;
+
+            assert!(
+                matches!(result, Err(LoginError::LoginFailure)),
+                "expected login failure, got {:#?}",
+                result
+            );
+        }
+        // If central server is not accessible after trying to login with old password, make sure old password does not work
```
andreievg, line 485: One last thing please, can you add this issue in the comment please https://github.com/openmsupply/remote-server/issues/1101

service/src/apis/login_v4.rs
```diff
@@ -128,13 +132,23 @@ impl LoginApiV4 {
...
-            .await?
-            .error_for_status()?;
+            .await
+            .map_err(|e| LoginV4Error::ConnectionError(e))?;
+
+        if reqwest::StatusCode::UNAUTHORIZED == response.status() {
+            return Err(LoginV4Error::Unauthorised);
+        }
+
+        let response = response
```
andreievg, line 145: Btw don't need to do Ok(response) in case, that last await returns a result

andreievg (approved):
thanks for all of the changes.

only thing really before merging is adding the issue in the comment

bijaySussol (approved):
@roxy-dao : 
Nice, I tested it with those changes, This works great !
Thanks

PR #1100, opened 2022-05-18
1096 improved name sync

clemens-msupply:


repository/src/db_diesel/name_row.rs
```diff
@@ -1,33 +1,113 @@
...
+    Male,
+    TransgenderMale,
+    TransgenderMaleHormone,
+    TransgenderMaleSurgical,
+    TransgenderFemale,
+    TransgenderFemaleHormone,
+    TransgenderFemaleSurgical,
+    Unknown,
+    NonBinary,
+}
```
clemens-msupply, line 53: This classification is from SNOMED but is it too detailed?

server/src/sync/translation_remote/mod.rs
```diff
@@ -60,56 +58,3 @@ pub fn table_name_to_central(table: &ChangelogTableName) -> &'static str {
...
-{
-    x.map(|date| date.and_hms(0, 0, 0)).serialize(s)
-}
-
-/// Currently v5 returns times in sec and v3 expects a time string when posting. To make it more
-/// consistent v5 behaviour might change in the future. This helper will make it easy to do the
-/// change on our side.
-pub fn naive_time<'de, D: Deserializer<'de>>(d: D) -> Result<NaiveTime, D::Error> {
-    let secs = u32::deserialize(d)?;
-    Ok(NaiveTime::from_num_seconds_from_midnight(secs, 0))
```
clemens-msupply, line 114: I moved these method from translation_remote so that they can be used in translation_central

server/src/sync/translation_central/name.rs
```diff
@@ -1,20 +1,106 @@
...
+    comment: Option<String>,
+    #[serde(deserialize_with = "empty_str_as_option")]
+    country: Option<String>,
+
+    #[serde(deserialize_with = "empty_str_as_option")]
+    #[serde(rename = "bill_address1")]
+    address1: Option<String>,
+    #[serde(deserialize_with = "empty_str_as_option")]
+    #[serde(rename = "bill_address2")]
+    address2: Option<String>,
```
clemens-msupply, line 81: I mapped bill_address to address since bill_address is also used for the patient address. Was not sure about this...

server/src/sync/translation_central/name.rs
```diff
@@ -1,20 +1,106 @@
...
+    #[serde(deserialize_with = "empty_str_as_option")]
+    #[serde(rename = "url")]
+    website: Option<String>,
+
+    #[serde(rename = "manufacturer")]
+    is_manufacturer: bool,
+    #[serde(rename = "donor")]
+    is_donor: bool,
+    #[serde(rename = "hold")]
+    on_hold: bool,
```
clemens-msupply, line 95: In the api we use is_on_hold but this should be just on_hold, right?

andreievg: sure, i think in api too just `on_hold`, not sure why I've made it `is_on_hold`

andreievg (approved):
I had a quick look, thanks for changes, lgtm, I'll probably check it out in a bit more depth later

PR #1102, opened 2022-05-18
Fix postgres tests on systems with nanosec time precision

clemens-msupply:
Postgres tests where failing on my machine and I think I finally found the reason: my system time has nano second precision but postgres has micro second precision. This always uses micro second precision for the test.


service/src/sync_processor/test.rs
```diff
@@ -21,6 +23,14 @@ mod test_update {
         sync_processor::{invoice::common::get_lines_for_invoice, process_records, Record},
     };
 
+    // Postgres stores timestamps with microsecond precision. This method produces a now
+    // NaiveDateTime timestamp with microsecond precision so that it can be compared to timestamps
+    // from the DB.
+    fn now() -> NaiveDateTime {
+        let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
```
andreievg, line 30: Hmm, how about just using with_nanosecod ?

Utc::now().naive_utc().with_nanosecond(0).unwrap()

Also was wondering if this should be generic helper method that we use everywhere `datetime_now()` or something like this

clemens-msupply: thanks, I tried it but unfortunately it doesn't work "Makes a new `NaiveDateTime` with nanoseconds since the whole non-leap second changed." so it changes the milli and micro seconds as well :(

Yes will keep it in mind to refactor the time method when I next need it!

andreievg (approved):

